PR #2, opened 2021-07-01
Scaffold monorepo

wildmolasses:
Monorepo with /app and /contracts folders!

Make sure to install `volta` and check the README's for other dev setup instructions and flow.


README.md
```diff
@@ -1,56 +1,48 @@
...
-npm install
+```sh
+cp app/.env.template app/.env
+nano app/.env
 
-To install the drizzle/vue-plugin you may have to do the following
-`$ npm install --legacy-peer-deps @drizzle/vue-plugin`
-`$ npm install --legacy-peer-deps vuex`
+cp contracts/.env.template contracts/.env
+nano contracts/.env
```
mds1, line 21: A bit of a nit, but `cp` and `nano` don't exist on Windows so I'd recommend just explaining the steps in regular words so it's better documented for cross-platform/accessible to new Windows devs

app/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "scripts": {
+    "clean": "rimraf dist",
+    "dev": "vite",
+    "build": "vue-tsc --noEmit && vite build",
+    "serve": "vite preview",
+    "test:unit": "vue-cli-service test:unit",
+    "test:e2e": "vue-cli-service test:e2e",
+    "lint": "eslint --ext .ts,.js,.vue .",
+    "precommit": "lint-staged",
+    "prettier": "prettier --write"
```
mds1, line 14: This command doesn't work! Options to fix are:
```suggestion
    "prettier": "prettier --write ."
```

OR 

```suggestion
    "prettier": "prettier --write \"**/*.{vue,json,md,sol,js,ts}\""
```

contracts/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@dcgrants/contracts",
+  "version": "0.0.1",
+  "dependencies": {},
+  "devDependencies": {
+    "@ethersproject/abstract-signer": "^5.0.6",
+    "@ethersproject/bignumber": "^5.0.8",
```
mds1, line 7: Recommend removing these since they are included with `ethers` so we don't need them here, and should also bump the ethers dependency below to 5.4.0 which adds EIP-1559 support

contracts/types/augmentations.d.ts
```diff
@@ -0,0 +1,13 @@
+// eslint-disable @typescript-eslint/no-explicit-any
```
mds1, line 1: Do we need this file? In past projects I've removed it and it seems to not change anything, so I don't quite understand what it does?

tsconfig.settings.json
```diff
@@ -0,0 +1,13 @@
+{
+    "compilerOptions": {
+      "module": "CommonJS",
+      "sourceMap": true,
+      "target": "ES2018",
```
mds1, line 5: Why ES2018 here? More generically, we should decide how wide of browser range we need to support. I'd love to use ES2020 to use `bigint`, but for Gitcoin it feels more appropriate to target something older for wider device/browser support

README.md
```diff
@@ -1,56 +1,48 @@
...
+yarn test
+```
 
-"0xCdF0b25D17356ccb34ACe7e27418DA1Bd81298fA","0x7d39759ec589ca32d75f86a4d25697132d9fea82f68d5dc9c254836340182b88","http://blah.com/hi.json"
+### Serve
 
-"0x7d39759ec589ca32d75f86a4d25697132d9fea82f68d5dc9c254836340182b88","0xCdF0b25D17356ccb34ACe7e27418DA1Bd81298fA","0xCdF0b25D17356ccb34ACe7e27418DA1Bd81298fA","https://blah.com/blah.json","0","0x7de9759ece89ca92c7538624d25697132d9fea82f68d5dc9c254836340182b87"
+```sh
+yarn build
+yarn start
```
mds1, line 47: Seems there's no `start` script

mds1 (changes requested):
Looks great overall! Was able to build both the contracts and app folders

contracts/types/augmentations.d.ts
wildmolasses: looks like it augments the mocha context with contract types. see https://gist.github.com/PaulRBerg/0a8b4806898d81685719d7f7a26afff0

That being said, i don't think we usually use the mocha context, so i'll delete

tsconfig.settings.json
wildmolasses: whatever you suggest we can do, just let us know. otherwise i'll leave at es2018.

contracts/types/augmentations.d.ts
wildmolasses: actually using the mocha context here does look a bit cleaner than

```js
let token;

//

token = await erc20.deploy(...)
```

tsconfig.settings.json
apbendi: Created an issue to track this here: https://github.com/dcgtc/dcgrants/issues/3

Let's leave at ES2018 for this PR though.

contracts/hardhat.config.ts
```diff
@@ -0,0 +1,102 @@
...
+  mainnet: 1,
+  rinkeby: 4,
+  ropsten: 3,
+};
+
+// Ensure that we have all the environment variables we need.
+let mnemonic = process.env.MNEMONIC as string;
+if (!mnemonic) {
+  console.warn('Please set your MNEMONIC in a .env file');
+  mnemonic = 'taxi taxi taxi taxi taxi taxi taxi taxi taxi taxi taxi taxi';
```
mds1, line 31: Recommend using `test test test test test test test test test test test junk`, which is the default Hardhat mnemonic. It will be a bit more compatible with things like e.g. the Hardhat optimism plugins if we end up using those

app/README.md
```diff
@@ -0,0 +1,48 @@
+# Ethereum App Template
+
+Ethereum frontend app built with the following stack:
+
+- [Vue 3](https://v3.vuejs.org/) as the foundation
+- [Tailwind CSS](https://tailwindcss.com) for styling
```
phutchins, line 6: We may opt to replace Nightwind and Tailwind but we can leave them for now.

phutchins (approved):
Left a few comments but other than that all looks good! It builds and runs for me and everything feels clean.

app/README.md
mds1: Could you speak to this a bit more? Replacing the full CSS from Tailwind to some other system becomes a bigger task the longer we wait, so I think we should decide this sooner rather than later (Nightwind is just plug and play for now, so is much easier to remove later)

phutchins: Our designer @scco is big on simple design without frameworks. If we decided not to use it, it may be more of a completely remove type of thing than a replace with something else so the effort might not be as bad as you think. Our direction right now is to build just enough UI to test/try the platform that we're building so we shouldn't spend time just yet worrying about making it pretty so if we were to wipe the ui in a few weeks and start over it shouldn't mater. I'm currently in between as I have not used Tailwind before and would like to give it a shot. I also would like for it to be relatively easy for people that are not our designer to make additions/changes so a framework might be useful. On the other hand they can make things more complex. I think we can decide this over the next week or two as @scco will be devoting more time to this project a few days from now.

mds1: Thanks for the context! Created https://github.com/dcgtc/dcgrants/issues/4 to track this

README.md
```diff
@@ -1,56 +1,70 @@
...
-### Vue
-```
-npm install
+Copy `app/.env.template` to `app/.env` and edit, providing your own env vars
 
-To install the drizzle/vue-plugin you may have to do the following
-`$ npm install --legacy-peer-deps @drizzle/vue-plugin`
-`$ npm install --legacy-peer-deps vuex`
+```bash
+cp app/.env.template app/.env
```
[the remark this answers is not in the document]
apbendi: Done in 09abc65724d22ac51fcc65204c06e49f4815c513

README.md
```diff
@@ -1,56 +1,70 @@
...
-or
+To ensure that everyone is using the same version of nodejs on this project, [volta](https://volta.sh) is recommended!
 
-`$ truffle migrate`--reset --compile-all
+### Set your env files
 
-### Vue
-```
-npm install
+Copy `app/.env.template` to `app/.env` and edit, providing your own env vars
```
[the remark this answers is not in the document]
apbendi: Done in 09abc65724d22ac51fcc65204c06e49f4815c513

app/src/components/LayoutFooter.vue
```diff
@@ -0,0 +1,82 @@
...
+          class="text-gray-400 hover:text-gray-500"
+        >
+          <span class="sr-only">{{ item.name }}</span>
+          <component :is="item.icon" class="h-6 w-6" />
+        </a>
+      </div>
+      <div class="mt-8 md:mt-0 md:order-1">
+        <SunIcon v-if="isDark" @click="toggleDarkMode" class="h-5 w-5 text-gray-400 cursor-pointer" />
+        <MoonIcon v-else @click="toggleDarkMode" class="h-5 w-5 text-gray-400 cursor-pointer" />
+        <p class="text-sm text-center text-gray-400 mt-2">&copy; 2020 ScopeLift. All rights reserved.</p>
```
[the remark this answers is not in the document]
apbendi: Ahh good call— sorry forgot to remove the ScopeLift stuff form our template! I replaced with Gitcoin for now, though perhaps a Creative Commons License should be considered for site content in the future if it aligns better with the ethos of the project!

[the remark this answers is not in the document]
apbendi: resolved in 6087cbc0adf2eb4ae9737bf26f41c2978fb25766

app/src/components/LayoutFooter.vue
```diff
@@ -0,0 +1,82 @@
...
+        h('svg', { fill: 'currentColor', viewBox: '0 0 24 24' }, [
+          h('path', {
+            d: 'M8.29 20.251c7.547 0 11.675-6.253 11.675-11.675 0-.178 0-.355-.012-.53A8.348 8.348 0 0022 5.92a8.19 8.19 0 01-2.357.646 4.118 4.118 0 001.804-2.27 8.224 8.224 0 01-2.605.996 4.107 4.107 0 00-6.993 3.743 11.65 11.65 0 01-8.457-4.287 4.106 4.106 0 001.27 5.477A4.072 4.072 0 012.8 9.713v.052a4.105 4.105 0 003.292 4.022 4.095 4.095 0 01-1.853.07 4.108 4.108 0 003.834 2.85A8.233 8.233 0 012 18.407a11.616 11.616 0 006.29 1.84',
+          }),
+        ]),
+    }),
+  },
+  {
+    name: 'GitHub',
+    href: 'https://github.com/ScopeLift',
```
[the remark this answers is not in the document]
apbendi: Done in 6087cbc0adf2eb4ae9737bf26f41c2978fb25766

app/src/components/LayoutFooter.vue
```diff
@@ -0,0 +1,82 @@
...
+<script lang="ts">
+import { defineComponent, h } from 'vue';
+import { MoonIcon, SunIcon } from '@heroicons/vue/solid';
+import useSettingsStore from 'src/store/settings';
+
+// See here to find more social media icon SVG paths: https://www.flaticon.com/free-icons/social-media
+const navigation = [
+  {
+    name: 'Twitter',
+    href: 'https://twitter.com/ScopeLift',
```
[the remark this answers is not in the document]
apbendi: done in 6087cbc0adf2eb4ae9737bf26f41c2978fb25766

Issue #3, opened 2021-07-07
Decide which version of JavaScript to support


apbendi:
Currently configured to ES2018, but would be nice to move to ES2020 for bigint support. Need to research browser support for ES2020 and decide if it's sufficient before going this route.

apbendi:
Closed for discussion in #77

PR #5, opened 2021-07-08
Re-import Original Prototype Contracts

apbendi:
* Adds @phutchins original contracts prototyping back to the repo
* Sets up a test harness for those contracts
* Remove template Greeter contract & tests


contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+
+  function isActive(bytes32 _id) public view returns (bool) {
+    return grants[_id].state == State.ACTIVE;
+  }
+
+  function isClosed(bytes32 _id) public view returns (bool) {
+    return grants[_id].state == State.CLOSED;
+  }
+
+  function getGrantCount() public view returns (uint256) {
```
apbendi, line 75: Not really needed since grantCount is public.

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+  // Mapping from GrantID to set of Owner Address, Payout Address, Metadata Pointer, Replaces Grant (optional)
+  // TODO: Implement safemath for avoiding overflows with ints
+  uint256 public grantCount = 0;
+  mapping(bytes32 => Grant) public grants;
+  bytes32[] public grantIds;
+
+  struct Grant {
+    bytes32 id;
+    address owner;
+    address payout;
```
apbendi, line 49: nit: recommend changing name to `payee` or `payoutAddr` for clarity

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+  //State public state;
+
+  // Grant Pointer Object
+  // Mapping from GrantID to set of Owner Address, Payout Address, Metadata Pointer, Replaces Grant (optional)
+  // TODO: Implement safemath for avoiding overflows with ints
+  uint256 public grantCount = 0;
+  mapping(bytes32 => Grant) public grants;
+  bytes32[] public grantIds;
+
+  struct Grant {
```
apbendi, line 46: Should the data about a Grant live in a Registry, or should it live in some external structure, where a Registry simply collects a list of Grant Unique identifiers? This question relates to the general question above about what the purpose of a GrantRegistry is.

An alternative to this approach would be for each Grant to be a contract instance that conforms to some `GrantInterface` requirement. This solves the unique identifier problem: the contract instance becomes the identifier. It would also allow for more flexibility and extensibility of the functionality around a Grant itself (imagine for example a Grant instance that is itself a mini DAO whose members have control of the parameters which define that grant and where the funds go). Furthermore, if we use something like UpgradeableProxies, it might provide a migration path for Grant owners to migrate their Grant contract over the course of our iteration with this product.

Another alternative (or perhaps orthogonal) approach would be to make each Grant instance an ERC721 NFT that could be transferred. In this instance, the ERC721 contract could be a sort of "global registry" of all Grants, where Rounds/Registry could collect the ERC721 identifiers.

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+    bytes32 id;
+    address owner;
+    address payout;
+    string metaPtr;
+    State state;
+    bytes32 replaces;
+  }
+
+  function activate(bytes32 _id) external onlyOwner {
```
apbendi, line 55: nit: recommend name change to `activateGrant` for clarity.

Also recommend emitting an event.

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+    grants[_id].state = State.ACTIVE;
+  }
+
+  function closed(bytes32 _id) public onlyOwner {
```
apbendi, line 59: nit: recommend name change to `closeGrant` for clarity.

Also recommend emitting an event.

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+    grants[_id].state = State.CLOSED;
+  }
+
+  function isConfig(bytes32 _id) public view returns (bool) {
```
apbendi, line 63: nit: recommend changing to or `isGrantConfigurable` for clarity

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+    return grants[_id].state == State.CONFIG;
+  }
+
+  function isActive(bytes32 _id) public view returns (bool) {
```
apbendi, line 67: nit: recommend changing to `isGrantActive` for clarity

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+    return grants[_id].state == State.ACTIVE;
+  }
+
+  function isClosed(bytes32 _id) public view returns (bool) {
```
apbendi, line 71: nit: recommend changing to `isGrantClosed` for clarity

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+    return grants[_id].state == State.CLOSED;
+  }
+
+  function getGrantCount() public view returns (uint256) {
+    return grantCount;
+  }
+
+  function getAllGrantIds() public view returns (bytes32[] memory) {
```
apbendi, line 79: These would only be useful as offchain convenience methods, as they'd eventually run out of gas if executed onchain once enough grants were added. Since anyone can add a Grant, we'd expect that to happen eventually.

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+  // When a grant changes statuses
+
+  // _mint function
+  // Allow creation of a grant and setting the initial parameters
+  function _mint(
+    bytes32 _id,
+    address _owner,
+    address _payout,
+    string memory _metaPtr,
+    State _state,
```
apbendi, line 122: Why would we want to make `state` an input parameter here?

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+    bytes32 _replaces
+  ) public {
+    incrementGrantCount();
+    grants[_id] = Grant(_id, _owner, _payout, _metaPtr, _state, _replaces);
```
apbendi, line 126: If `_id` already exists, this will overwrite the existing `Grant`

contracts/contracts/GrantRounds.sol
```diff
@@ -0,0 +1,48 @@
...
+    bytes32 _grantRoundsId,
+    string memory _metaPtr
+  ) {
+    owner = _owner;
+    grantRoundsId = _grantRoundsId;
+    metaPtr = _metaPtr;
+  }
+
+  // State
+  enum State {
```
apbendi, line 28: I'm not convinced we need these explicitly stored states. We might be able to get away with two parameters: `startTimestamp` and `endTimestamp`, with the state of the round derived from these values and the current blocktime.

contracts/contracts/GrantRounds.sol
```diff
@@ -0,0 +1,48 @@
+pragma solidity ^0.8.5;
+
+contract GrantRounds {
```
apbendi, line 3: nit: I'd `GrantRounds` sounds awkward. I would go with `GrantsRound`.

contracts/contracts/GrantRounds.sol
```diff
@@ -0,0 +1,48 @@
...
+  // When a Grant Round is created
+  // When a donation is made to a Grant
+
+  uint256 public grantRoundCount = 0;
+  mapping(bytes32 => GrantRound) public grantRounds;
+
+  struct GrantRound {
+    bytes32 id;
+    address owner;
+    bytes[] proof;
```
apbendi, line 45: What is the `proof`?

contracts/contracts/GrantRounds.sol
```diff
@@ -0,0 +1,48 @@
...
+    bytes32 root;
```
apbendi, line 46: What is the `root`?

contracts/contracts/GrantRounds.sol
```diff
@@ -0,0 +1,48 @@
...
+    ACTIVE,
+    CLAIM,
+    CLOSED
+  }
+
+  // Events
+  // When a Grant Round is created
+  // When a donation is made to a Grant
+
+  uint256 public grantRoundCount = 0;
```
apbendi, line 39: Hmm, is the idea that a `GrantRounds` contract would enable more than one round? This might explain the (seemingly) awkward naming I mentioned above.

I would question why we'd want a singleton contract that hosts multiple rounds. I can see this adding a lot of complexity. Couldn't/wouldn't each `GrantsRound` be deployed as it's own contract (possibly with a factory and minimal proxies if desired)?

apbendi (commented):
Lots of questions/comments, as promised! In addition to the specific questions below, some generic recommendations:

* We should add natspec comments to all contracts/params/functions and require those as a part of our development flow from the jump
* We should organize contracts in a conventional way, such as declaring Events, then defining Structs, Enums, then public vars, then functions, then modifiers. The specifics don't matter, just that we agree on and follow a convention.
* We should have a convention about naming params/memory/storage vars. We tend to prefix any non-storage var with `_` to make it totally clear when an action writes/reads to storage. Again, specific convention is less important that having one.
* Moving forward, test coverage should be required alongside additional contract functionality before a PR is merged.

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
+pragma solidity ^0.8.5;
+
+contract GrantRegistry {
+  // Grant ID - The hash of the grants meetadata
+  bytes32 public grantRegistryId;
+  // Metadata Pointer
+  string public metaPtr;
+
+  // Owner
+  address owner;
```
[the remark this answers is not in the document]
mds1: I agree we probably do want that. For something as simple as ownership, we might want to just implement it ourselves, even if we copy/paste some of the `Ownable` code. Just to cut down on dependencies and contract junk/bloat (like `Context` and unnecessary getters)

------------------------

Also we probably want to make `owner` public. When not specified, it defaults to private

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+
+  modifier onlyOwner() {
+    require(msg.sender == owner);
+    _;
+  }
+
+  // TODO: Only allow the owner (which is the curator) address to configure the contract
+  // while its in config mode, then switch to active. Once active, the owner can only
+  // close the registry (have a conversation around this approach)
```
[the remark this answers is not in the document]
mds1: I had similar questions. Not quite sure what the purpose of `State` is

contracts/contracts/GrantRegistry.sol
mds1: Agreed, we should remove this

contracts/contracts/GrantRegistry.sol
mds1: I think we can remove this method. `getAllGrants` serves a similar purpose, but returns the ID + additional data. So I'm not sure what the use case is for `getAllGrantIds`

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+
+  function getAllGrants() public view returns (Grant[] memory) {
+    Grant[] memory ret = new Grant[](grantCount);
+    bytes32 grantId;
+    for (uint256 i = 0; i < grantCount; i++) {
+      grantId = grantIds[i];
+      ret[i] = grants[grantId];
+    }
+    return ret;
+  }
```
mds1, line 95: Should be able to save some gas by removing an MSTORE—even though it's an off-chain helper, you can still run out of gas because nodes have a gas limit on `eth_call`s. For example (using a `returnData` to be more explicit than `ret`):

```solidity
  Grant[] memory returnData = new Grant[](grantCount);
  for (uint256 i = 0; i < grantCount; i++) {
    returnData[i] = grants[grantIds[i]];
  }
  return returnData;
```

Also, given that this can run out of gas, there should also be a version of this method that lets you enter a start and end index, because eventually this call will fall when `grantCount` is large enough. For example

```solidity
function getAllGrants() external view returns (Grant[] memory) {
    return getSomeGrants(0, grantsCount);
}

function getSomeGrants(uint256 startIndex, uint256 endIndex) public view returns (Grant[] memory) {
    Grant[] memory returnData = new Grant[](grantCount);
    bytes32 grantId;
    for (uint256 i = startIndex; i < endIndex; i++) {
      grantId = grantIds[i];
      returnData[i] = grants[grantId];
    }
    return returnData;
}
```

contracts/contracts/GrantRegistry.sol
mds1: Another thought if grants are deployed from a factory, the unique ID can just be counter equal to `grantCount`. The first time a new grant is deployed it has an ID of 0 since `grantCount = 0`, next one has ID of 1

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
...
+    incrementGrantCount();
+    grants[_id] = Grant(_id, _owner, _payout, _metaPtr, _state, _replaces);
+    grantIds.push(_id);
+    // Emit an event here for grant minted
+    emit NewGrant(_id, _owner, _payout, _replaces);
+  }
+
+  function incrementGrantCount() internal {
+    grantCount += 1;
+  }
```
mds1, line 134: Feels like this should just be inlined for now since it's only used once. Saves some bytecode + gas

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
+pragma solidity ^0.8.5;
+
+contract GrantRegistry {
+  // Grant ID - The hash of the grants meetadata
```
mds1, line 4: I'm going to pushback on using the hash of the metadata as the grant ID, for the reasons @apbendi has previously mentioned—an onchain transaction to deploy a new grant just to fix a typo is terrible (and expensive) UX

contracts/contracts/GrantRounds.sol
```diff
@@ -0,0 +1,48 @@
+pragma solidity ^0.8.5;
+
+contract GrantRounds {
+  // Grant Round ID - The hash of the grant round metadata
```
mds1, line 4: Some comment as above about not using the hash of a round as metadata

contracts/contracts/GrantRounds.sol
```diff
@@ -0,0 +1,48 @@
...
+  bytes32 public grantRoundsId;
+  // Metadata Pointer
+  string public metaPtr;
+
+  // Owner
+  address owner;
```
mds1, line 10: `owner` should be public and update-able as mentioned earlier

mds1 (commented):
Agree with all of @apbendi's comments, and left some more. 

Also, lots of functions are `public` when they really should be `external`

contracts/contracts/GrantRegistry.sol
thelostone-mc: makes sense but since this lives beyond anyone's control we don't want weird scenario like me getting my grant verified by the DAO and then updating content to pose as metamask / umbra / etc. Just a thought

contracts/contracts/GrantRegistry.sol
```diff
@@ -0,0 +1,135 @@
+pragma solidity ^0.8.5;
+
+contract GrantRegistry {
```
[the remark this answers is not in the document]
phutchins: The goal of the GrantRegistry is to allow anyone to create a Grant. 

Our options as I see them are as follows:
1) Have a dApp deploy a new contract for each grant that is created
2) Have a factory pattern contract that deploys a new grant contract for each new grant created
3) Have a contract that follows a pattern similar to the multi token standard (ERC1155) where grants are created and tracked within a single GrantRegistry contract

Option #3 seems the most cost effective to me but it would be great to see a more data informed comparison between our options. In this first round of prototype, it may be worth simply implementing #3 unless we have a strong reason not to and reassess if needed.

After our discussion this morning, I think we can move forward assuming that there will be one GrantRegistry contract that we will support. Anyone could of course deploy their own GrantRegistry contract but that would result in a completely separate ecosystem.

This GrantRegistry would hold ALL grants and not be filtered or curated. A grant would and could only ever exist in one GrantRegistry.

A GrantRound would have a subset of grants from a GrantRegistry. The list could live on or off chain. 

If on-chain, it could be a list contract which contains an array of grants. The GrantRound would link to the List contract to define which grants are in the round.

If off-chain, the GrantRound could link to off chain metadata which might be a JSON document containing an array of grants.

The lists could be created manually but I imagine would be eventually be the result of first curation (removing the 'bad' grants through some process, or including the 'good' grants using properties like sybil score and the entities participation policy), then applying a filter (selecting grants with the desired properties such as theme, capital raised, etc...)

contracts/contracts/GrantRegistry.sol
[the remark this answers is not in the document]
phutchins: I added this when I was considering deploying a contract for each grant. I have since changed the approach to having a single contract which manages many grants.

We can remove this.

contracts/contracts/GrantRegistry.sol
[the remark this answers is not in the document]
phutchins: I believe we can remove this. I was playing around with the idea of taking Grants and GrantRounds through different states but I think it is only necessary in instances aGrantRound within a GrantRounds contract.

We can remove the state bits from this contract.

contracts/contracts/GrantRegistry.sol
phutchins: Sounds good.

contracts/contracts/GrantRegistry.sol
phutchins: Makes sense to me. Will change.

contracts/contracts/GrantRegistry.sol
phutchins: I think my comment above covers most of this.

Regarding the ID, the count makes sense to me.

contracts/contracts/GrantRegistry.sol
phutchins: Sounds good.

contracts/contracts/GrantRegistry.sol
phutchins: Sounds good.

contracts/contracts/GrantRegistry.sol
phutchins: Sounds good.

Issue #6, opened 2021-07-09
Grant Registry Requirements Implmentation


apbendi:
- [ ] Global
- [ ] Store the data for each grant
- [ ] Anyone can create a grant
- [ ] Only the grant owner can update their grant
- [ ] Grant Registry needs to have an upgrade path itself

mds1:
Regarding upgradeability, I'd like to discuss that a bit more. I have a two main thoughts here:

1. If the contract can be upgraded, someone (such as the Gitcoin DAO) needs special privileges, e.g. an owner. The current version of this contract is credibly neutral, but a version containing privileged roles is not. IMO this is a heavy downside, and makes it much more likely that additional `GrantRegistry` contracts may be deployed by other entities, which fragments the grants ecosystem. For example, in the current implementation of the Registry, no grant can be censored / blocked from creation, but a future version could add this. Therefore who doesn't trust the owner might therefore want to create their own registry instead

2. One reason for upgradeability might be to add more data to the `Grant` struct. However, we can work around that without needing to update the `GrantRegistry` implementation. A simple version of this is to add one more field to the grants struct of type `bytes extraData`. This `extraData` field can be set to anything, so allows for future extensibility specified by some (off-chain) encoding scheme. For (a totally random) example, say after the first round we say "It'd be useful if the `Grant` struct had a `lastUpdateTimestamp` and `lastMatchAmount` fields. Let's specify the first 16 bytes of `extraData` as the timestamp, and the next 24 bytes as the last match amount".  We'd of course need one more `update` method to let owners set the bytes field

Therefore, I think we should think about *what specifically* we may want to upgrade in the future, and try to think of ways to account for it now that don't require new contracts.

In the very worst case, if we really need new functionality later, you can deploy a `GrantRegistryV2` and initialize it with the `GrantRegistryV1` state, OR communicate to grant owners that they must migrate themselves and recreate their grant. Requiring a migration to a `GrantRegistryV2` (instead of an in-place update) is arguably a good thing as it clears state grants from state that frontends and curators need to handle

apbendi:
I'm very sympathetic to this argument @mds1 and would be eager to hear what @phutchins and others think.

One contribution I'd make is that there are also gradations in between the DAO having total upgradeability powers, and no privileges whatsoever. One quick example: the DAO could have "pause" functionality, that let's them freeze the addition/removal of existing Grants, which would be useful in the migration to a would-be V2.

apbendi:
Standup discussion:

[...]

PR #7, opened 2021-07-09
Initial implementation of GrantRegistry

mds1:
This contract behaves as follows:
- It is designed as a singleton, i.e. we assume there is only one `GrantRegistry`
- Anyone can create a grant with `createGrant`
- Only the grant owner can edit their grant (details below)
- `getAllGrants` and `getGrants` view methods are implemented to bulk-fetch grant data
- We do not deploy a new contract for each grant as I don't see a reason for it at this time, but we can add this later

Implementation details:
- A grants `id` is simply a counter, starting a 0 and incrementing after each grant is created
- A `GrantCreated` is emitted on grant creation
- A `Grant` struct contains the grants ID, owner, payee and metadata pointer
    - While the ID is unnecessary in the grant struct (because you must know it to fetch the grant's data), it is convenient so a frontend querying a grant by ID receives an object with all on-chain data for that grant
    - To avoid increasing gas costs, we use a `uint96` for the grant ID. Addresses are 20 bytes (160 bits), meaning we can pack a 160 bit address and 96 bit ID into a single 256 bit storage slot.
    - This packing gives us a limit of `2^96-1 = 7.9e28` grants that can be stored by the registry. This represents creating 1 grant per second for 2.5e21 years, so that should be sufficient! 😛
- To make dgrants as accessible (i.e. cheap and gas-efficient) as possible, I avoided modifiers and instead inlines `require` statements. There is probably some more gas golfing we can do at a later date
- There are four ways to edit a grant:
    - Update individual fields with `updateGrantOwner`,  `updateGrantPayee`, `updateGrantMetaPtr`
    - Update the full grant at once with `updateGrant`
    - All update operations emit the same event to simplify off-chain analysis (i.e. fewer topics to fetch, sort, and parse)

Upgradeability is not considered here and will first be fleshed out in #6


contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,124 @@
...
+    Grant storage grant = grants[_id];
+    require(msg.sender == grant.owner, "Not authorized");
+    grant.payee = _payee;
+    emit GrantUpdated(grant.id, grant.owner, grant.payee, grant.metaPtr);
   }
 
-  function getGrantCount() public view returns (uint256) {
-    return grantCount;
+  /**
+   * @notice Update the payee of a grant
```
apbendi, line 69: Notice is incorrect due to copy/paste

contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,124 @@
...
-  function isActive(bytes32 _id) public view returns (bool) {
-    return grants[_id].state == State.ACTIVE;
+  /**
+   * @notice Update the owner of a grant
+   * @param _id ID of grant to update
+   * @param _owner New owner address
+   */
+  function updateGrantOwner(uint96 _id, address _owner) external {
+    Grant storage grant = grants[_id];
+    require(msg.sender == grant.owner, "Not authorized");
```
apbendi, line 51: What's the actual gas savings from not using a modifier? Does the optimizer not inline modifiers?? The repeated code makes me a little twitchy but I do think keeping gas down is worthwhile.

contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,124 @@
...
+    emit GrantUpdated(grant.id, grant.owner, grant.payee, grant.metaPtr);
   }
 
+  // --- View functions ---
+  /**
+   * @notice Returns an array of all grants and their on-chain data
+   * @dev May run out of gas for large values `grantCount`, depending on the node's RpcGasLimit. In these cases,
+   * `getGrants` can be used to fetch a subset of grants and aggregate the results of various calls off-chain
+   */
   function getAllGrants() public view returns (Grant[] memory) {
```
apbendi, line 108: Can be `external` instead of `public`

contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,124 @@
...
-    grantCount = getGrantCount();
-    myGrant[] memory ret = new myGrant[](grantCount);
-    for (uint i = 0; i < grantCount; i++) {
-      ret[i] = grants[i];
+  /**
+   * @notice Returns a range of grants and their on-chain data
+   * @param _startId Grant ID of first grant to return, inclusive, i.e. this grant ID is included in return data
+   * @param _endId Grant ID of last grant to return, exclusive, i.e. this grant ID is NOT included in return data
+   */
+  function getGrants(uint96 _startId, uint96 _endId) public view returns (Grant[] memory) {
```
apbendi, line 117: Can be `external` instead of `public`

contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,124 @@
...
+   * @param _payee New payee address
+   * @param _metaPtr New URL that points to grant metadata
+   */
+  function updateGrant(
+    uint96 _id,
+    address _owner,
+    address _payee,
+    string calldata _metaPtr
+  ) external {
+    Grant storage grant = grants[_id];
```
apbendi, line 94: Asking, because I'm not sure off the top of my head: is there any gas difference if we just update the whole grant in place? Thinking it might prevent this storage read.

contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,124 @@
...
 
-  function closed(bytes32 _id) public onlyOwner {
-    grants[_id].state = State.CLOSED;
-  }
+  // --- Events ---
+  /// @notice Emitted when a new grant is created
+  event GrantCreated(uint96 indexed id, address indexed owner, address indexed payee, string metaPtr);
+
+  /// @notice Emitted when a grant's owner is changed
+  event GrantUpdated(uint96 indexed id, address indexed owner, address indexed payee, string metaPtr);
```
apbendi, line 24: Wondering if we should include the old value in the update event?

contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,124 @@
 pragma solidity ^0.8.5;
 
 contract GrantRegistry {
```
apbendi, line 3: We should add a natspec comment for the contract.

contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,124 @@
...
   }
 
-  function isActive(bytes32 _id) public view returns (bool) {
-    return grants[_id].state == State.ACTIVE;
+  /**
+   * @notice Update the owner of a grant
+   * @param _id ID of grant to update
+   * @param _owner New owner address
+   */
+  function updateGrantOwner(uint96 _id, address _owner) external {
```
apbendi, line 49: Can you expand on your reasoning for 4 separate update methods, instead of just 1? Is it just gas savings?

contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,124 @@
...
-    myGrant[] memory ret = new myGrant[](grantCount);
-    for (uint i = 0; i < grantCount; i++) {
-      ret[i] = grants[i];
+  /**
+   * @notice Returns a range of grants and their on-chain data
+   * @param _startId Grant ID of first grant to return, inclusive, i.e. this grant ID is included in return data
+   * @param _endId Grant ID of last grant to return, exclusive, i.e. this grant ID is NOT included in return data
+   */
+  function getGrants(uint96 _startId, uint96 _endId) public view returns (Grant[] memory) {
+    Grant[] memory returnData = new Grant[](_endId - _startId);
```
apbendi, line 118: nit: what do you think of `grantList` instead of `returnData`?

contracts/test/GrantRegistry.test.ts
```diff
@@ -0,0 +1,222 @@
...
+const { deployContract } = waffle;
+const { isAddress } = ethers.utils;
+const randomAddress = () => ethers.Wallet.createRandom().address;
+const randomPtr = () => String(Math.random()); // doesn't need to be a valid URL for testing
+const BN = (x: BigNumberish) => BigNumber.from(x);
+
+// --- Type definitions ---
+// The output from ethers/typechain allows array or object access to grant data, so we must define types for
+// handling the Grant struct as done below
+type GrantObject = {
```
apbendi, line 21: We'll almost certainly need these type definitions in other tests, so might be worth extracting them to a file and importing them now establish the pattern. Eventually, we may even want to share these test types with the frontend. Worth considering.

contracts/test/GrantRegistry.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    return { owner, payee, metaPtr };
+  }
+
+  // Helper method to verify that two Grant objects are equal
+  function expectEqualGrants(grant1: Grant, grant2: Grant) {
+    expect(grant1.id).to.equal(grant2.id);
+    expect(grant1.owner).to.equal(grant2.owner);
+    expect(grant1.payee).to.equal(grant2.payee);
+    expect(grant1.metaPtr).to.equal(grant2.metaPtr);
+  }
```
apbendi, line 49: might be nice to move these into a test util file as well

apbendi (commented):
Looks like a great start! Nice work. Left comments but mostly small stuff and some questions.

contracts/test/GrantRegistry.test.ts
wildmolasses: i'm happy to create a shared types directory at the root, either here in this PR or as a followup. Let me know how I Can help.

contracts/test/GrantRegistry.test.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  beforeEach(async () => {
+    [deployer, user1, user2] = await ethers.getSigners();
+
+    const grantRegistryArtifact: Artifact = await artifacts.readArtifact('GrantRegistry');
+    registry = <GrantRegistry>await deployContract(deployer, grantRegistryArtifact);
```
wildmolasses, line 55: this syntax looks a little strange to me. <...> is usually used to supply a type for a generic. so what is the generic here? Or is this some other language feature entirely?

contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,124 @@
...
-  /*
-  function getAllGrants() public view returns (myGrant[] memory) {
-    grantCount = getGrantCount();
-    myGrant[] memory ret = new myGrant[](grantCount);
-    for (uint i = 0; i < grantCount; i++) {
-      ret[i] = grants[i];
+  /**
+   * @notice Returns a range of grants and their on-chain data
+   * @param _startId Grant ID of first grant to return, inclusive, i.e. this grant ID is included in return data
+   * @param _endId Grant ID of last grant to return, exclusive, i.e. this grant ID is NOT included in return data
```
wildmolasses, line 115: why exclude the grant with id `_endId` in the returned grant array? i can see that it saves some typing in that we can use `grantCount` instead of `grantCount-1`, but as a con it adds complexity to consumers of this method i.e. now i have to remember the upper boundary is exclusive.

contracts/test/GrantRegistry.test.ts
```diff
@@ -0,0 +1,222 @@
+// --- External imports ---
```
wildmolasses, line 1: nice import organization!

wildmolasses (approved):
lgtm, just 1 suggestion on `getGrants` and then a question!

contracts/contracts/GrantRegistry.sol
mds1: Here's my current understanding (have not tested/confirmed this recently): 

When using modifiers, the function body and modifier code are inlined together, so, compared to using a function to replace the modifier, a modifier will have bigger bytecode but will be cheaper since there's no JUMPs between methods. Functions (that replace a modifier's job) will have smaller bytecode but be a little more expensive due to JUMPs.

But in this case, our modifier would need to read from storage, AND our function needs to read from the same storage slot. I don't think there's a way to have the modifier and function use the same SLOAD (maybe the optimizer would do this? Haven't tested it), so inlining it this way ensures we only use one SLOAD

SLOADs [cost](https://github.com/wolflo/evm-opcodes/blob/hardfork-berlin/gas.md#a6-sload) 2100 gas for the first read of a slot and 100 gas for subsequent reads

contracts/contracts/GrantRegistry.sol
mds1: This one does need to be `public` since it's called by `getAllGrants`

contracts/contracts/GrantRegistry.sol
mds1: We have to do the storage read to check the owner at the beginning of the function, unless I'm misunderstanding something?

contracts/contracts/GrantRegistry.sol
mds1: I was thinking about that, but decided against it because of gas savings, and if you need the old values they are still available through other logs. So I couldn't think of a strong reason for including 3 additional parameters

contracts/contracts/GrantRegistry.sol
mds1: Actually we'd only read from the same storage slot twice in the case of updating `owner`, since `payee` and `metaPtr` have their own storage slots. So you could argue for only using this pattern when updating `owner`.

I think this pattern should still be the cheapest by a few gas though, so I think we should leave it for now and open a separate issue to gas-golf the whole contract more generally

contracts/contracts/GrantRegistry.sol
mds1: Yea the two reasons I thought of for splitting this up are:

1. Gas savings: The separate methods ensures you only write to a minimum number of storage slots
2. UX/devex: Requiring all of `owner`, `payee`, and `metaPtr` to be passed just to update one of those 3 is inconvenient and error prone

contracts/test/GrantRegistry.test.ts
mds1: Shared types directory at the root is a good idea. Created this issue and assigned it to you @wildmolasses: https://github.com/dcgtc/dgrants/issues/8

For this PR, we don't currently have other tests that need these types, and don't know whether #8 or new contract tests will be done first, so seems we should leave this as-is for now to avoid work that may be unnecessary

contracts/test/GrantRegistry.test.ts
mds1: Hm, where else would we need to check that two Grant objects are equal?

contracts/test/GrantRegistry.test.ts
mds1: It's a way of type casting, so `<GrantRegistry>await deployContract()` is identical to `(await deployContract()) as GrantRegistry`

https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions

contracts/contracts/GrantRegistry.sol
mds1: This format of "inclusive start index, exclusive end index" seems to be the standard way to slice arrays, which I why I did it. For example, JavaScript's `slice`, Python's list/array indexing, and Solidity's array slicing all work this way

contracts/test/GrantRegistry.test.ts
mds1: Thanks!

contracts/contracts/GrantRegistry.sol
mds1: Fixed in 756f836a84c1f1b47bf58f12734df5e871fd6b50

mds1: Fixed in 756f836a84c1f1b47bf58f12734df5e871fd6b50

contracts/contracts/GrantRegistry.sol
mds1: Fixed in 756f836a84c1f1b47bf58f12734df5e871fd6b50

contracts/contracts/GrantRegistry.sol
mds1: Done in 756f836a84c1f1b47bf58f12734df5e871fd6b50

contracts/contracts/GrantRegistry.sol
mds1: Changed in 756f836a84c1f1b47bf58f12734df5e871fd6b50

contracts/contracts/GrantRegistry.sol
apbendi: Yep, you're right. Reading the `owner` property loads the storage slot for the grant `_id` anyway, and there's no way to avoid that check. I don't think there's any room for improvement here.

contracts/contracts/GrantRegistry.sol
apbendi: Sounds good. I'm fine with keeping all four, though it's worth considering that if you want to update 2 of 3 params, you still end up in the same situation gas/DevEx wise. One thing we might want to discuss is whether a multicall method on the contract might be helpful in this regard.

contracts/test/GrantRegistry.test.ts
apbendi: Sounds like a good plan. Feel free to tackle this at your convenience @wildmolasses. I've updated the PR description to note that existing tests should also be updated to use the shared type dir when the change is made.

contracts/test/GrantRegistry.test.ts
apbendi: Off the top of my head, one example might be a test of `GrantRounds` that shows the same grant can exist in two rounds or something? I think we'd want utilities like this in a generic place so that they're available and discoverable for future test authors who might be working on a new test, and wouldn't know to look for them in an existing test they didn't author.

contracts/contracts/GrantRegistry.sol
mds1: Yea I thought about adding a multicall but figured this probably wasn't the right PR for it. Created https://github.com/dcgtc/dgrants/issues/9 to track it

contracts/contracts/GrantRegistry.sol
apbendi: Yeah, I think this convention stems from `0` indexing— since the `length` of an array is always 1 greater than the largest index position, the second index is conventionally non-inclusive, so if you want the whole array you ask for 0 to length. I also find it confusing @wildmolasses but let's keep it as-is since this is the convention used by the languages most devs interacting with this contract will be familiar with.

contracts/test/GrantRegistry.test.ts
apbendi: I was confused by this at first too, but I'm fine with using this syntax. Any objections @wildmolasses.

contracts/contracts/GrantRegistry.sol
apbendi: Gotcha. I think in general we should avoid over-optimization initially, favoring more readable code, and look to do a gas-golfing pass before deployment. This feels like it's bordering on that "premature optimization" line, since modifiers are a widely used language feature that improve readability and reduce the likelihood of a dev error. Let's leave it as-is for now, but be careful about that over optimization line moving forward!

contracts/contracts/GrantRegistry.sol
apbendi: Sounds good. Let's leave it out for now, but as we're working on the frontend, we can think about if we hit any places where we'd wish we had the old values in the event.

contracts/test/GrantRegistry.test.ts
mds1: That makes sense, done in d263ee4b655acd56d3cc4634253fdf38f761febf

apbendi (approved):
Updates look great! 🎉

contracts/contracts/GrantRegistry.sol
wildmolasses: Go is like this as well! But R is not! 🤷‍♂️  OK, sounds good!

contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,132 @@
...
-
-  // Grant Pointer Object
-  // Mapping from GrantID to set of Owner Address, Payout Address, Metadata Pointer, Replaces Grant (optional)
-  // TODO: Implement safemath for avoiding overflows with ints
-  uint256 public grantCount = 0;
-  mapping(bytes32 => Grant) public grants;
-  bytes32[] public grantIds;
+  // --- Data ---
+  /// @notice Number of grants stored in this registry
+  uint96 public grantCount;
```
thelostone-mc, line 14: why do we store this value btw? 
Can this not be something which derived via a subgraph or so ?

contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,132 @@
...
-    address owner;
-    address payout;
-    string metaPtr;
-    State state;
-    bytes32 replaces;
+    uint96 id; // grant ID, as uint96 to pack into same slot as owner (this implies a max of 2^96-1 = 7.9e28 grants)
+    address owner; // grant owner (has permissions to modify grant information)
+    address payee; // address that receives funds donated to this grant
+    string metaPtr; // URL pointing to grant metadata (for off-chain use)
   }
```
thelostone-mc, line 22: Is it overkill to have a status for Grant ?
Status to indicate -> open for funding / closed for funding

contracts/contracts/GrantRegistry.sol
```diff
@@ -1,135 +1,132 @@
...
+  /**
+   * @notice Create a new grant in the registry
+   * @param _owner Grant owner (has permissions to modify grant information)
+   * @param _payee Address that receives funds donated to this grant
+   * @param _metaPtr URL pointing to grant metadata (for off-chain use)
+   */
+  function createGrant(
+    address _owner,
+    address _payee,
+    string memory _metaPtr
```
thelostone-mc, line 44: noob question !
Do we need to worry about grants sharing same _metaPtr or nah it doesn't matter cause the entity would take care of it while creation ?

contracts/test/GrantRegistry.test.ts
wildmolasses: is this an area where we value consistency? if so, for the following 3 reasons i recommend we use the "as" assertions over angle bracket assertions:

- we use the `as` syntax in all other repos we work on
- angle bracket syntax is not compatible with jsx
- angle brackets are already used for generics

contracts/contracts/GrantRegistry.sol
mds1: The `grantCount` is used to determine the grants ID in the `createGrant` method, which is why we store it

contracts/contracts/GrantRegistry.sol
mds1: Hmm, I don't think we need it, since that can be part of the JSON schema that will live off-chain at the `metaPtr` URL. But curious to hear what @apbendi or @phutchins think

contracts/contracts/GrantRegistry.sol
mds1: Ideally we don't want two grants sharing the same `metaPtr`, because why would two grants have the same metadata, but I don't think this is worth enforcing on-chain. Probably fine to leave this to the matching eligibility layer, but again tagging @apbendi and @phutchins for input

contracts/contracts/GrantRegistry.sol
phutchins: If metadata is unavailable and the status lives in the metadata, how do we handle displaying a grant? This type of situation makes me think that it should live in the grant however in this early stage it prob doesn't mater as much. I'd. say that we leave it in the grant for now to make it easier for filter and display and then move it to the metadata if we determine that the cost is too high.

contracts/contracts/GrantRegistry.sol
phutchins: Agree on your approach here @mds1.

phutchins (approved):

contracts/contracts/GrantRegistry.sol
mds1: One idea @apbendi mentioned is we could also just set the payee address to the zero address to indicate a grant is not available for funding, that way we don't need another parameter here. OTOH, we can pack the status into the `payee` slot so it won't take up any additional storage space/gas

thelostone-mc: Ah yes that would be perfect @mds1

Issue #14, opened 2021-07-15
Improve wallet connection experience


mds1:
Currently it can be a bit jumpy of a UX when automatically connecting to your last used wallet

jfarid27:
I did a bit of research on this and I ended up with this line as the culprit. The problem is the ENS lookup is an async request that takes some time. If it's removed, the view update is quite snappy. I'm not exactly sure what a good fix is here.

https://github.com/dcgtc/dgrants/blob/main/app/src/store/wallet.ts#L150

PR #15, opened 2021-07-15
Grant registry UI prototyping

### LINKED ISSUE #11: Grant Registry UI Page Prototyping
This issue defines the routes and baseline functionality for the prototype UI's interaction with the `GrantRegistry` contract implmented in #7

- [x] Registry List Page
   - [x] route: `/dgrants`
   - [x] Shows a list of every grant in the registry
   - [x] For each grant, displays: grant id, owner address, payee address, metaPtr bytes
   - [x] Each Grant listed can be clicked to go to the Grant Detail age
   - [x] Has a "Create Grant" button that goes to the New Grant page
- [x] Grant Detail Page
   - [x] route:`/dgrants/:id` where `:id` is the integer ID of the Grant in the GrantRegistry
   - [x] Shows each field for the grant: id, owner address, payee address, metaPtr bytes
   - [x] If the user's connected wallet address is the owner address for this grant, it should have an "Edit Grant" button. Clicking the button should make the various fields for the grant editable. Clicking "Save" should submit the appropriate transaction iff the fields have changed.
- [x] New Grant Page
   - [x] route: `/dgrants/new`
   - [x] Should have fields for each (configurable) Grant field: owner address, payee address, metaPtr bytes
   - [x] Should do basic validation on the fields (all 3 are required; addresses should be valid addresses)
   - [x] Should have a "Create" button that is clickable when the fields are populated properly and submits the appropriate on-chain transaction
   - [x] Should redirect the user to the Grant Detail Page of the newly created Grant when the tx is mined

Note: in the long term we will (obviously) not show the metaPtr bytes, but rather load, parse, and display the metadata from the storage location. This is a temporary requirement to demonstrate interaction with the Registry is working properly.

### LINKED ISSUE #18: Rename Infura env variable for clarity
In our .env files, the VITA_INFURA_API_KEY variable could be renamed to VITA_INFURA_PROJECT_ID (or VITA_INFURA_ID) to match the way that the value is named on Infura. For consistency, the INFURA_ID variable could be renamed to INFURA_PROJECT_ID (or left alone if we use VITA_INFURA_ID).

mds1:
Closes #11, closes #18

- Implements everything described in #11
- Setup script forks mainnet, deploys `GrantRegistry` and creates dummy grants with arbitrary data
- Read `app/README.md` for setup/troubleshooting instructions when testing against a forked Hardhat node
- Didn't focus on designs too much, just pulled and modified some Tailwind UI components (for example, edit page is basically a copy/paste of the New Grant page to save time for now)
- Wallet connection can be a bit flaky, so created #14 to track this

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -39,6 +39,14 @@ yarn test:unit
 yarn test:e2e
 ```
 
+### Troubleshooting
+
+If you send a transaction in MetaMask against your local node, then restart your local node, your next transaction will use a nonce that is too large and your transaction will fail. To fix this, go to MetaMask Settings > Advanced > Reset Account. This will reset the MetaMask nonce counter so it matches what Hardhat is expecting.
+
+If you see `project ID does not have access to archive state` in the terminal, simply quit the process and re-run `yarn dev`. This happens because Infura only gives access to the last 128 blocks of data (~30 minutes), unless you pay for an archive node plan. If you leave your forked mainnet node running for longer than this, it tries to query for data older than 128 blocks and fails. Restarting your node re-forks from the latest mainnet block.
+
+If the app loads and the block number is zero after connecting your wallet, there's likely a `CALL_EXCEPTION` in the console. Simply refresh the page and this should be fixed.
+
 ## Notes / Customization
 
 Notes on customizing this app:
--- a/app/README.md
+++ b/app/README.md
@@ -16,13 +16,10 @@ Ethereum frontend app built with the following stack:
 # Install packages
 yarn install
 
-# Start local node and fill contract with dummy data (must be run before `yarn dev`)
+# Run in development mode (start local node and fills contract with dummy data)
 cd ../contracts
 yarn app:node
-yarn app:setup
 cd ../app
-
-# Run in development mode
 yarn dev
 
 # Compiles and minifies for production
--- a/app/README.md
+++ b/app/README.md
@@ -15,7 +15,7 @@ Ethereum frontend app built with the following stack:
 For local testing, add a network to MetaMask with the following information:
 
 - Name: Hardhat
-- New RPC URL: https://127.0.0.1:8545
+- New RPC URL: http://127.0.0.1:8545
 - Chain ID: 31337
 
 ```sh
--- a/app/README.md
+++ b/app/README.md
@@ -16,6 +16,12 @@ Ethereum frontend app built with the following stack:
 # Install packages
 yarn install
 
+# Start local node and fill contract with dummy data (must be run before `yarn dev`)
+cd ../contracts
+yarn app:node
+yarn app:setup
+cd ../app
+
 # Run in development mode
 yarn dev
 
--- a/app/README.md
+++ b/app/README.md
@@ -12,14 +12,17 @@ Ethereum frontend app built with the following stack:
 
 ## Setup
 
+For local testing, add a network to MetaMask with the following information:
+
+- Name: Hardhat
+- New RPC URL: https://127.0.0.1:8545
+- Chain ID: 31337
+
 ```sh
 # Install packages
 yarn install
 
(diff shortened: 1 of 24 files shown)


app/src/views/GrantRegistryNewGrant.vue
```diff
@@ -0,0 +1,117 @@
...
+    </div>
+
+    <div class="mt-8 sm:mx-auto sm:w-full sm:max-w-md text-left">
+      <div class="py-8 px-4 border shadow sm:rounded-lg sm:px-6">
+        <form class="space-y-6" @submit.prevent="createGrant">
+          <!-- Owner address -->
+          <BaseInput
+            v-model="form.owner"
+            description="The owner has permission to edit the grant"
+            id="owner-address"
```
phutchins, line 19: When in dark mode its difficult to see the data in the fields.

phutchins (changes requested):

app/README.md
```diff
@@ -12,11 +12,17 @@ Ethereum frontend app built with the following stack:
 
 ## Setup
 
+For local testing, add a network to MetaMask with the following information:
+
+- Name: Hardhat
+- New RPC URL: https://127.0.0.1:8545
```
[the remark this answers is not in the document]
mds1: Fixed in a854369

app/src/views/GrantRegistryNewGrant.vue
mds1: Good catch, I forgot to check how dark mode looked. Made some improvements in 6f68b3f0ba8ce6c631d811373389117b1ab8d485 so it should look better now

apbendi (approved):
All functionality seems to be working as expected. DevEx is great as well 👍 Nice work!

One question I did have was around your decision to fork mainnet for the test network. Wasn't clear to me why we needed that, so was just wondering what your motivation was?

Otherwise good to merge from my perspective.

mds1:
> One question I did have was around your decision to fork mainnet for the test network. Wasn't clear to me why we needed that, so was just wondering what your motivation was?

It's not needed, but I thought it made things easier for two reasons:
1. If we didn't fork mainnet, anyone who wants to test the app would need to either (1) import a Hardhat private key into MetaMask, or (2) setup a new MetaMask instance using the Hardhat mnemonic. Forking lets you just use your standard MetaMask account that you normally use, so it seemed easier (you have to change RPC in both cases anyway)
2. The Multicall contract used to read data is deployed at the same address on all networks, so forking lets us ensure that's also true on localhost (you should be able to do this without forking with `hardhat_setCode` though), that way we can hardcode that address as a constant in the app code, instead of varying by chain ID

thelostone-mc (approved):
🔥

apbendi:
> It's not needed, but I thought it made things easier for two reasons....

Cool, sounds good. Happy to leave it forking, though I did want to mention that I think it's somewhat risky to use funded mainnet accounts while doing dev/test work. It really ups the risk of forgetting/getting confused about which network you're connected to, and signing/sending an errant transaction that results in lost funds. I don't think we should encourage this workflow in README's or docs. It's relatively easy to import a private key(s) to use account[N] of the Hardhat network into MetaMask, and I think this is what we should encourage devs to do.

mds1:
I think every development account I use has some mainnet funds, because sometimes you gotta test in prod! 😅

Though for now this shouldn't matter since the app/transactions will fail since the contracts don't exist on mainnet. But I do see your point. Should we just add one step to the README, such as:

```
For local testing, add a network to MetaMask with the following information:

- Name: Hardhat
- New RPC URL: http://127.0.0.1:8545
- Chain ID: 31337

You should also add one of the Hardhat default private keys as a MetaMask account and use that account when testing the app. This ensures you do not accidentally send/lose mainnet funds <--- NEW TO README
```

Issue #12, opened 2021-07-13
Single GrantsRound Contract Spec


apbendi:
This issue is a minimal spec for a smart contract implementing a single matching grant round.

- [ ] Has an `owner` with special internal privileges (see below)
- [ ] Has a `startTime` timestamp that must in the future at the time of deployment
- [ ] Has an `endTime` timestamp that must be after the startTime
- [ ] Has a `donationToken` address that is assumed to be an ERC20 implementation
- [ ] Has a `registry` address that refers to a GrantsRegistry instance
- [ ] Has a string `metaPtr` variable that is set at deployment
- [ ] Has a `minContribution` variable that is set at deployment, representing the minimum matching contribution that can be made to a grant during the round
- [ ] Has a `hasPaidOut` boolean variable that defaults to false, and toggles only when the owner calls the payout method (see below)
- [ ] At any point before `endTime`, (including before the `startTime`), the Round can accept matching pool donations in `donationToken` via a special method. Internally, the method should use `transferFrom` to take custody of the funds and emit an event.
- [ ] During the active period, the Round can accept matching donations for a specific Grant via a special method.
    - [ ] The method should take the `grantId` as a parameter, and verify that the Grant exists in the GrantRegistry
    - [ ] The method should ensure the `amount` is greater than or equal to the `minContribution` amount
    - [ ] The method should use `transferFrom` to send the funds the Grant's `payee` address
    - [ ] The method should emit an event
- [ ] After the active period, the Round should disallow matching pool donations and grant contributions
- [ ] After the active period, the `owner` should be able to transfer all matching pool funds to a `payoutAddress` via a special method call. Only the owner should be able to call this method, and only after the `endTime` has passed.

Open Questions:

* Should matching donations be allowed while the round is active, or only before?
* Should a minimum round duration be enforced in any way at deployment? 

Notes:

 * For simplicity and further consideration, I've left any concept of whitelisting or curating grants from the round out of this spec. The only requirement is that the grant exist in the Registry.

thelostone-mc:
> Should matching donations be allowed while the round is active, or only before?

I'd say it should !
The reason being this only affects how the distribution is calculated and it allows projects to get in more funds 
 
> Should a minimum round duration be enforced in any way at deployment?

Feels more like a nice to have than a necessity. Any reason you think this might be beneficial ?

===============

**Open Questions**

- Would we allow the start-date / end-date to be extended ? 
[...]

PR #17, opened 2021-07-15
GrantsRound contract with tests

corydickson:
Addresses #12

### DIFF
--- a/contracts/contracts/GrantRegistry.sol
+++ b/contracts/contracts/GrantRegistry.sol
@@ -129,4 +129,14 @@ contract GrantRegistry {
     }
     return grantList;
   }
+
+  function isGrantInRegistry(uint96 _id) public view returns (bool) {
+    Grant storage grant = grants[_id];
+    return grant.id == _id;
+  }
+
+  function getGrantPayee(uint96 _id) public view returns (address) {
+    Grant storage grant = grants[_id];
+    return grant.payee;
+  }
 }
--- a/contracts/contracts/GrantRegistry.sol
+++ b/contracts/contracts/GrantRegistry.sol
@@ -131,15 +131,6 @@ contract GrantRegistry {
   }
 
   /**
-   * @notice Returns a boolean that is true when a given grant ID is in the registry
-   * @param _id Grant ID of a possible grant in the registry
-   */
-  function isGrantInRegistry(uint96 _id) public view returns (bool) {
-    Grant storage grant = grants[_id];
-    return grant.id == _id;
-  }
-
-  /**
    * @notice Returns the address that will be used to pay out donations for a given grant
    * @dev The payee may be set to null address
    * @param _id Grant ID used to retrieve the payee address in the registry
--- a/contracts/contracts/GrantRegistry.sol
+++ b/contracts/contracts/GrantRegistry.sol
@@ -130,11 +130,20 @@ contract GrantRegistry {
     return grantList;
   }
 
+  /**
+   * @notice Returns a boolean that is true when a given grant ID is in the registry
+   * @param _id Grant ID of a possible grant in the registry
+   */
   function isGrantInRegistry(uint96 _id) public view returns (bool) {
     Grant storage grant = grants[_id];
     return grant.id == _id;
   }
 
+  /**
+   * @notice Returns the address that will be used to pay out donations for a given grant
+   * @dev The payee may be set to null address
+   * @param _id Grant ID used to retrieve the payee address in the registry
+   */
   function getGrantPayee(uint96 _id) public view returns (address) {
     Grant storage grant = grants[_id];
     return grant.payee;
--- a/contracts/contracts/GrantRound.sol
+++ b/contracts/contracts/GrantRound.sol
@@ -6,25 +6,25 @@ import "./GrantRegistry.sol";
 
 contract GrantRound {
   using SafeERC20 for IERC20;
-  // Unix timestamp of the start of the round
+  // @notice Unix timestamp of the start of the round
   uint256 startTime;
-  // Unix timestamp of the end of the round
+  // @notice Unix timestamp of the end of the round
   uint256 endTime;
-  // Contract owner
+  // @notice Contract owner
   address owner;
-  // GrantsRegistry
+  // @notice GrantsRegistry
   address registry;
-  // ERC20 token that accepts pool donations
+  // @notice ERC20 token that accepts pool donations
   address donationToken;
-  // URL pointing to grant metadata (for off-chain use)
+  // @notice URL pointing to grant round metadata (for off-chain use)
   string metaPtr;
-  // minimum contribution amt that can be made
+  // @notice minimum contribution amt that can be made
   uint256 minContribution;
-  // Set to true if grant round has ended and payouts have been released
+  // @notice Set to true if grant round has ended and payouts have been released
   bool hasPaidOut;
 
   /// @notice Emitted when a grant receives a donation
-  event GrantDonation(uint96 indexed id, address indexed payee, uint256 donationAmount);
+  event DonationSent(uint96 indexed id, address indexed token, uint256 amount, address dest, address indexed donor);
 
   modifier onlyOwner() {
     require(msg.sender == owner);
@@ -46,6 +46,16 @@ contract GrantRound {
     _;
   }
 
+  /**
+   * @notice Instantiates a new grant round
+   * @param _owner Grant round owner that has permission to payout the matching pool
+   * @param _registry Address that contains the grant metadata
+   * @param _donationToken Address of the ERC20 token in which donations are made
+   * @param _startTime Unix timestamp of the start of the round
+   * @param _endTime Unix timestamp of the end of the round
+   * @param _metaPtr URL pointing to the grant round metadata
+   * @param _minContribution Miniumum donation amount that can be made using the given donation token
+   */
   constructor(
     address _owner,
     address _registry,
@@ -65,37 +75,56 @@ contract GrantRound {
     minContribution = _minContribution;
   }
 
-  function acceptMatchingPool(uint256 donationAmount) external beforeRoundEnd {
-    _safeTransferDonationToken(donationAmount);
+  /**
+   * @notice Before the round ends this method accepts matching pool funds
+   * @param _amount The amount of donation token that can be sent to the contract for the matching pool
+   */
+  function acceptMatchingPool(uint256 _amount) external beforeRoundEnd {
+    _safeTransferDonationToken(msg.sender, address(this), _amount);
   }
 
-  function donateToGrant(uint256 donationAmount, uint96 grantId) external activeRound {
-    require(donationAmount >= minContribution, "Donation amount must be greater than minimum contribution");
+  /**
+   * @notice During an active round users can use this method to send donation tokens to a specified grant
+   * @param _donationAmount The number of tokens to be sent to grant receipient
+   * @param _grantId The id of the grant in the registry
+   */
+  function donateToGrant(uint256 _donationAmount, uint96 _grantId) external activeRound {
+    require(_donationAmount >= minContribution, "Donation amount must be greater than minimum contribution");
 
     GrantRegistry gRegistry = GrantRegistry(registry);
+    require(gRegistry.isGrantInRegistry(_grantId), "Grant with given id does not exist in registry provided");
 
-    require(gRegistry.isGrantInRegistry(grantId), "Grant with given id does not exist in registry provided");
+    address payee = gRegistry.getGrantPayee(_grantId);
+    require(payee != address(0), "Payee not set in the grant metadata");
+    _safeTransferDonationToken(msg.sender, payee, _donationAmount);
 
-    _safeTransferDonationToken(donationAmount);
-
-    address payee = gRegistry.getGrantPayee(grantId);
-
-    IERC20 token = IERC20(donationToken);
-    token.safeTransfer(payee, donationAmount);
-
-    emit GrantDonation(grantId, payee, donationAmount);
+    emit DonationSent(_grantId, donationToken, _donationAmount, payee, msg.sender);
   }
 
-  function payoutGrant(address payoutAddress) external roundEnd onlyOwner {
+  /**
+   * @notice When the round ends the owner can send the remaining matching pool funds to a given address
+   * @param _payoutAddress An address to receive the remaining matching pool funds in the contract
+   */
+  function payoutGrant(address _payoutAddress) external roundEnd onlyOwner {
     IERC20 token = IERC20(donationToken);
     uint256 balance = token.balanceOf(address(this));
-    token.safeTransfer(payoutAddress, balance);
+    token.safeTransfer(_payoutAddress, balance);
 
     hasPaidOut = true;
   }
 
-  function _safeTransferDonationToken(uint256 amount) internal {
+  /**
+   * @notice Helper function that wraps safeTransferFrom for the ERC20 donation token
+   * @param _from Address that sends to tokens to a particular address
+   * @param _to Address that receives the donation token
+   * @param _amount The amount of donation token to be sent to a given address
+   */
+  function _safeTransferDonationToken(
+    address _from,
+    address _to,
+    uint256 _amount
+  ) internal {
     IERC20 token = IERC20(donationToken);
-    token.safeTransferFrom(msg.sender, address(this), amount);
+    token.safeTransferFrom(_from, _to, _amount);
   }
 }
--- a/contracts/contracts/GrantRound.sol
+++ b/contracts/contracts/GrantRound.sol
@@ -7,42 +7,42 @@ import "./GrantRegistry.sol";
 contract GrantRound {
   using SafeERC20 for IERC20;
   // @notice Unix timestamp of the start of the round
-  uint256 startTime;
+  uint256 public startTime;
   // @notice Unix timestamp of the end of the round
-  uint256 endTime;
+  uint256 public endTime;
   // @notice Contract owner
-  address owner;
+  address public owner;
   // @notice GrantsRegistry
-  address registry;
+  address public registry;
   // @notice ERC20 token that accepts pool donations
-  address donationToken;
+  address public donationToken;
   // @notice URL pointing to grant round metadata (for off-chain use)
-  string metaPtr;
+  string public metaPtr;
   // @notice minimum contribution amt that can be made
-  uint256 minContribution;
+  uint256 public minContribution;
   // @notice Set to true if grant round has ended and payouts have been released
-  bool hasPaidOut;
+  bool public hasPaidOut;
 
   /// @notice Emitted when a grant receives a donation
   event DonationSent(uint96 indexed id, address indexed token, uint256 amount, address dest, address indexed donor);
 
   modifier onlyOwner() {
-    require(msg.sender == owner);
+    require(msg.sender == owner, "Only owner can call this method");
     _;
   }
 
   modifier beforeRoundEnd() {
-    require(block.timestamp < endTime);
+    require(block.timestamp < endTime, "Time has passed to complete this tx");
     _;
   }
 
   modifier activeRound() {
-    require(block.timestamp >= startTime && block.timestamp < endTime);
+    require(block.timestamp >= startTime && block.timestamp < endTime, "Donations must be sent during an active round");
     _;
   }
 
   modifier roundEnd() {
-    require(block.timestamp >= endTime);
+    require(block.timestamp >= endTime, "Method must be called after the active round has ended");
     _;
   }
 
--- a/contracts/contracts/GrantRound.sol
+++ b/contracts/contracts/GrantRound.sol
@@ -15,7 +15,7 @@ contract GrantRound {
   // @notice GrantsRegistry
   address public registry;
   // @notice ERC20 token that accepts pool donations
-  address public donationToken;
+  IERC20 public donationToken;
   // @notice URL pointing to grant round metadata (for off-chain use)
   string public metaPtr;
   // @notice minimum contribution amt that can be made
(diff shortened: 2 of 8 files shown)


contracts/test/GrantRound.test.ts
```diff
@@ -0,0 +1,153 @@
...
+    });
+
+    it('sends remaining matching pool funds to payout address', async function () {
+      await mockERC20.mock.balanceOf.withArgs(grantPayee.address).returns(ethers.utils.parseEther(balances[0]));
+      await mockERC20.mock.balanceOf.withArgs(roundContract.address).returns(ethers.utils.parseEther(balances[0]));
+      await mockERC20.mock.transferFrom.withArgs(roundContract.address, grantPayee.address, balances[0]).returns(true);
+
+      // End the round for all subsequent tests
+      await timeTravel(deployer.provider, endTime + 1);
+      // await roundContract.connect(grantRoundOwner).payoutGrant(grantPayee.address);
```
corydickson, line 126: There seems to be an obscure bug with the mock contract that I'm still trying to work through. Doubtful it's the contract implementation but obviously could be wrong on that :grimacing:

contracts/test/GrantRound.test.ts
```diff
@@ -0,0 +1,153 @@
...
+    await mockERC20.mock.transfer.withArgs(donor.address, ethers.utils.parseEther(balances[0]));
+
+    const grantRegistryArtifact: Artifact = await artifacts.readArtifact('GrantRegistry');
+    registry = <GrantRegistry>await deployContract(deployer, grantRegistryArtifact);
+
+    // Create an example grant
+    await registry.connect(deployer).createGrant(deployer.address, grantPayee.address, mockUrl);
+
+    const grantRoundArtifact: Artifact = await artifacts.readArtifact('GrantRound');
+    const roundContractFactory = new ethers.ContractFactory(
```
corydickson, line 46: Mocks didn't seem to play nicely otherwise

contracts/test/GrantRound.test.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  beforeEach(async () => {
+    const mockUrl: string = 'https://test.url';
+    const minContribution: number = 50;
+    [deployer, grantRoundOwner, donor, grantPayee, mpUser] = await ethers.getSigners();
+    mockERC20 = await deployMockContract(deployer, IERC20Artifact.abi);
+    // Seed funds for matching pool user
+    await mockERC20.mock.transfer.withArgs(mpUser.address, ethers.utils.parseEther(balances[0]));
+    // Seed funds for initial donor
+    await mockERC20.mock.transfer.withArgs(donor.address, ethers.utils.parseEther(balances[0]));
```
mds1, line 37: Are these mock calls doing anything? If I comment them out tests still pass, and based on [the docs](https://ethereum-waffle.readthedocs.io/en/latest/mock-contract.html) it seems these need one more method chained on

contracts/test/GrantRound.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      );
+    });
+
+    it('donations revert if not above minimum contribution', async function () {
+      const grantId: number = 0;
+      await mockERC20.mock.balanceOf.withArgs(donor.address).returns(ethers.utils.parseEther(balances[0]));
+      await mockERC20.mock.balanceOf.withArgs(grantPayee.address).returns(ethers.utils.parseEther(balances[0]));
+      await mockERC20.mock.transfer.withArgs(grantPayee.address, balances[0]).returns(true);
+
+      void expect(roundContract.connect(donor).donateToGrant(10, grantId)).to.be.revertedWith(
```
mds1, line 102: These should all be `await` instead of `void`. With `void`, if the expect matcher would fail, you'll get an error logged to the console but since it's not `await`ed the test will still pass (even though it shouldn't)

contracts/test/GrantRound.test.ts
```diff
@@ -0,0 +1,153 @@
...
+      await roundContract.connect(mpUser).acceptMatchingPool(balances[0]);
+      expect(await roundContract.connect(deployer).checkBalance(roundContract.address)).to.be.equal(
+        ethers.utils.parseEther(balances[0])
+      );
+    });
+  });
+
+  describe('donateToGrant - Add funds during active round', () => {
+    it('sends donation token to the grant payee with a given grant id', async function () {
+      const grantId: number = 0;
```
mds1, line 85: Nit, but you don't need to specify the type here (and elsewhere where values are set). Typescript can infer it since you defined an initial number value
```suggestion
      const grantId = 0;
```

contracts/test/utils.ts
```diff
@@ -27,3 +27,8 @@ export function expectEqualGrants(grant1: Grant, grant2: Grant) {
   expect(grant1.payee).to.equal(grant2.payee);
   expect(grant1.metaPtr).to.equal(grant2.metaPtr);
 }
+
+export async function timeTravel(provider: any, seconds: number) {
+  await provider.send('evm_increaseTime', [Number(seconds)]);
```
mds1, line 32: Don't need to cast here since we specify the input as a number
```suggestion
  await provider.send('evm_increaseTime', [seconds]);
```

contracts/test/GrantRound.test.ts
mds1: Just generically, when pushing I'd recommend leaving this uncommented out so the test fails, that way it's more clear that there's still work to do on this PR

Overall, my impression here is that Waffle mocks aren't the best fit for these tests / testing actual token transfers, and we should probably just replace it with a `MockERC20` contract that behaves like a normal ERC20

contracts/contracts/GrantRegistry.sol
```diff
@@ -129,4 +129,23 @@ contract GrantRegistry {
...
   }
+
+  /**
+   * @notice Returns a boolean that is true when a given grant ID is in the registry
+   * @param _id Grant ID of a possible grant in the registry
+   */
+  function isGrantInRegistry(uint96 _id) public view returns (bool) {
+    Grant storage grant = grants[_id];
+    return grant.id == _id;
+  }
```
mds1, line 140: Can we instead just `return _id  < grantCount`? Feels simpler, plus then you can move that logic into the GrantRound contract instead of needing a special method here

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+    emit DonationSent(_grantId, address(donationToken), _donationAmount, payee, msg.sender);
+  }
+
+  /**
+   * @notice When the round ends the owner can send the remaining matching pool funds to a given address
+   * @param _payoutAddress An address to receive the remaining matching pool funds in the contract
+   */
+  function payoutGrant(address _payoutAddress) external roundEnd onlyOwner {
+    uint256 balance = donationToken.balanceOf(address(this));
+    donationToken.safeTransferFrom(address(this), _payoutAddress, balance);
```
mds1, line 110: This could just be a regular `transfer` since we're transferring from this contract. No need for `transferFrom` here!

mds1 (changes requested):
Looks great overall! 

Left a lot of comments, but mostly best practices type of things. There's also been a lot of discussion in #12 so we should review that and make sure any changes/new requirements are accounted for here

The biggest change I mentioned is that I think we should switch from a Waffle mock to an actual mock token contract that behaves like an ERC20, since it seems there's a few issues/bugs that arise/aren't caught with the Waffle mock approach.

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+  // @notice Contract owner
+  address public owner;
+  // @notice GrantsRegistry
+  address public registry;
+  // @notice ERC20 token that accepts pool donations
+  IERC20 public donationToken;
+  // @notice URL pointing to grant round metadata (for off-chain use)
+  string public metaPtr;
+  // @notice minimum contribution amt that can be made
+  uint256 public minContribution;
```
thelostone-mc, line 22: thoughts on maxing this immutable? 
Highly unlikely this would need to be changed

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+  // @notice Set to true if grant round has ended and payouts have been released
+  bool public hasPaidOut;
+
+  /// @notice Emitted when a grant receives a donation
+  event DonationSent(uint96 indexed id, address indexed token, uint256 amount, address dest, address indexed donor);
```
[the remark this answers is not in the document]
thelostone-mc: Agreed ^
So if we were doing batch transactions (N) -> we'd fire N DonationSent event right ?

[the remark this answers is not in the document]
mds1: Yep exactly!

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+
+  modifier onlyOwner() {
+    require(msg.sender == owner, "Only owner can call this method");
+    _;
+  }
+
+  modifier beforeRoundEnd() {
+    require(block.timestamp < endTime, "Time has passed to complete this tx");
```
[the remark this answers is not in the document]
thelostone-mc: The comment could be -> "Action cannot be performed as round has ended"

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+    metaPtr = _metaPtr;
+    minContribution = _minContribution;
+  }
+
+  /**
+   * @notice Before the round ends this method accepts matching pool funds
+   * @param _amount The amount of donation token that can be sent to the contract for the matching pool
+   */
+  function acceptMatchingPool(uint256 _amount) external beforeRoundEnd {
+    _safeTransferDonationToken(msg.sender, address(this), _amount);
```
thelostone-mc, line 83: Could we just do `donationToken.safeTransferFrom(msg.sender, address(this), _amount);` instead ?

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  /**
+   * @notice During an active round users can use this method to send donation tokens to a specified grant
```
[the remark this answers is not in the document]
thelostone-mc: Yup ! 
The `donationToken ` is taken into account only when you are contributing to the matching Pool.
As a contributor, one should be allowed to contribute with any ERC20 token

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+   * @param _donationAmount The number of tokens to be sent to grant receipient
+   * @param _grantId The id of the grant in the registry
+   */
+  function donateToGrant(uint256 _donationAmount, uint96 _grantId) external activeRound {
```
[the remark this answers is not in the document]
thelostone-mc: Would it make sense to have just a `donateToGrants` function which handles 1 to n contributions as opposed to having 2 separate ones ?

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+    require(_donationAmount >= minContribution, "Donation amount must be greater than minimum contribution");
```
thelostone-mc, line 92: Would it make sense to have this a modifier just to keep the function cleaner ?

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+
+    GrantRegistry gRegistry = GrantRegistry(registry);
+    require(gRegistry.isGrantInRegistry(_grantId), "Grant with given id does not exist in registry provided");
+
+    address payee = gRegistry.getGrantPayee(_grantId);
+    require(payee != address(0), "Payee not set in the grant metadata");
```
thelostone-mc, line 98: Would it make sense to group L94-L98 into it's own modifier and pass that ?
Essentially all if checks happen via modifiers

contracts/contracts/GrantRound.sol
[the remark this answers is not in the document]
mds1: Ah, yea good call I think that's the way to go. That's how we did it with BulkCheckout

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+    donationToken.transfer(payee, _donationAmount);
```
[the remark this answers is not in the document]
thelostone-mc: @mds1 would it be worth using `safeTransferFrom` here

contracts/package.json
```diff
@@ -1,7 +1,9 @@
 {
   "name": "@dcgrants/contracts",
   "version": "0.0.1",
-  "dependencies": {},
+  "dependencies": {
+    "@openzeppelin/contracts": "^4.2.0"
```
thelostone-mc, line 5: Do we want to fix the version as opposed to to having the `^` ? aka
`"@openzeppelin/contracts": "4.2.0"`

mds1: I think this is ok. It's pinned in the lock file and OZ is pretty stable, especially around ERC20 behavior which is what we're using it for

contracts/contracts/GrantRound.sol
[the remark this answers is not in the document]
mds1: Yea this should be `donationToken.safeTransferFrom`. There's an argument to be made that we don't need these `safe*` methods if we only plan to use trusted, safe, ERC-20 compliant donation tokens (e.g. not USDT!) to save gas. Bt we can save that discussion for a later PR when we gas golf the contracts

contracts/test/GrantRound.test.ts
```diff
@@ -0,0 +1,153 @@
...
+// --- Parse and define helpers ---
+const { deployContract, deployMockContract } = waffle;
+const { isAddress } = ethers.utils;
+
+describe('GrantRound', function () {
+  let deployer: SignerWithAddress,
+    grantRoundOwner: SignerWithAddress,
+    donor: SignerWithAddress,
+    grantPayee: SignerWithAddress,
+    mpUser: SignerWithAddress;
```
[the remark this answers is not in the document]
wildmolasses: looks like it's "matching pool user"

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+
+    emit DonationSent(_grantId, address(donationToken), _donationAmount, payee, msg.sender);
+  }
+
+  /**
+   * @notice When the round ends the owner can send the remaining matching pool funds to a given address
+   * @param _payoutAddress An address to receive the remaining matching pool funds in the contract
+   */
+  function payoutGrant(address _payoutAddress) external roundEnd onlyOwner {
```
thelostone-mc, line 108: I'm not super clear on this function.
Once the round has been completed -> for this version we assume the distribution curve is available.

- Is this not saved on the contract ?
- If not -> is the understanding right that if the round had 5 grants -> the owner would have to invoke this 5 times ?
- Would it make more sense to have this a function which takes in all grantId and how much amount needs to be paid to them -> invoke the bulkTransfer method in a BulkTransfer contract and pay it all in one shot ?
- Also shouldn't hasPaidOut=true be set only after all the payouts are done (cause only then the round is actually complete )

====

Another way of going about this is, where we let the grant owners invoke this function and funds would be claimed by them (but this complicates thing as the contract would have to be aware of how much each grants is supposed to get from the matching pool )

cc @mds1 / @corydickson  / @apbendi  / @phutchins

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
+pragma solidity ^0.8.5;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "./GrantRegistry.sol";
+
+contract GrantRound {
+  using SafeERC20 for IERC20;
```
apbendi, line 8: nit: Let's put a line of whitespace between each variable declaration, unless anyone is highly opposed to this. I think it's much more readable with some breathing room.

contracts/contracts/GrantRound.sol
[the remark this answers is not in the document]
apbendi: Also, should we agree to follow the convention of "ContractName: Message" for all revert strings? For example, this would be "GrantRound: Action cannot be performed as round has ended."

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+  string public metaPtr;
+  // @notice minimum contribution amt that can be made
+  uint256 public minContribution;
+  // @notice Set to true if grant round has ended and payouts have been released
+  bool public hasPaidOut;
+
+  /// @notice Emitted when a grant receives a donation
+  event DonationSent(uint96 indexed id, address indexed token, uint256 amount, address dest, address indexed donor);
+
+  modifier onlyOwner() {
```
apbendi, line 29: nit: can we move modifiers to the end of the contract? Really I don't care where they are, as long as we're consistent, but generally I've seen them placed at the end by convention. Anyone have strong feelings here?

contracts/contracts/GrantRound.sol
apbendi: Agree with @thelostone-mc— not sure the overhead of an internal function is worthwhile here. Doesn't seem any more clear than dot syntax.

contracts/contracts/GrantRound.sol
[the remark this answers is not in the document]
apbendi: I think a `GrantRound` should only accept donations in a single token, for both matching pool & grant contributions. Enabling donations with arbitrary tokens can/should be handled a layer up with a swapping contract that leverages Uniswap or another on chain Dex.

contracts/contracts/GrantRound.sol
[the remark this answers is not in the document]
apbendi: Let's hold off on batch donations in this PR, as it will probably be impacted by #20 anyway, and should be straightforward to implement regardless.

contracts/contracts/GrantRound.sol
apbendi: It does add some gas overhead. I'm OK either way.

contracts/contracts/GrantRound.sol
apbendi: I think I actually prefer the discrete checks as implemented, but don't feel strongly.

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+
+  /**
+   * @notice When the round ends the owner can send the remaining matching pool funds to a given address
+   * @param _payoutAddress An address to receive the remaining matching pool funds in the contract
+   */
+  function payoutGrant(address _payoutAddress) external roundEnd onlyOwner {
+    uint256 balance = donationToken.balanceOf(address(this));
+    donationToken.safeTransferFrom(address(this), _payoutAddress, balance);
+
+    hasPaidOut = true;
```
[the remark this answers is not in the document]
apbendi: Agreed on the checks-effects-interactions pattern

> And for some reason the compiler thinks this is unreachable code?

Hmm where are you seeing that? No warning for me.

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  // For testing purposes only
+  function checkBalance(address addr) public view returns (uint256) {
+    uint256 balance = donationToken.balanceOf(addr);
+    return balance;
+  }
```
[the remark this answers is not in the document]
apbendi: Agreed, should move this up to the test implementation.

apbendi (commented):
Great start! Left some additional feedback and responses to others comments. LMK if you have any follow up. Thanks @corydickson.

contracts/contracts/GrantRound.sol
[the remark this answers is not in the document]
mds1: 🤷 
image

contracts/contracts/GrantRound.sol
phutchins: This might should be called payoutGrants or approvePayout or something along those lines as it is transferring funds to somewhere else so that it can pay out multiple grants. I see this more as the approval to move funds to the destination responsible for paying out all of the grants.

contracts/contracts/GrantRound.sol
thelostone-mc: Ah alright ! If it saves gas costs , best to stick to what we have.

contracts/contracts/GrantRound.sol
thelostone-mc: Ah alright ! If it saves gas costs , best to stick to what we have.

contracts/contracts/GrantRound.sol
thelostone-mc: Ignore ^ this I wasn't aware we were letting the actual payout happen via another contract. All chill

contracts/test/GrantRound.test.ts
corydickson: I can change this to the other way we deploy contracts once I get rid of the mocks. Essentially it wasn't recognizing certain function calls

contracts/contracts/GrantRound.sol
```diff
@@ -0,0 +1,134 @@
...
+   * @param _metaPtr URL pointing to the grant round metadata
+   * @param _minContribution Miniumum donation amount that can be made using the given donation token
+   */
+  constructor(
+    address _owner,
+    address _registry,
+    IERC20 _donationToken,
+    uint256 _startTime,
+    uint256 _endTime,
+    string memory _metaPtr,
```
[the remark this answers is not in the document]
corydickson: Solidity compiler doesn't accept calldata for parameters defined in a constructor

contracts/contracts/GrantRegistry.sol
corydickson: Thinking about this a bit more, I think you're right and should just remove this function entirely, having the condition above cause a revert in the GrantRound contract

apbendi (approved):
Thanks for addressing the feedback @corydickson. Added a couple commits to address very minor issues, including minor formatting changes, and making the failing test skipped instead  so it doesn't result in failing CI once #25 is merged.

phutchins (approved):
Looks good to merge.

Issue #34, opened 2021-07-22
Add validation logic to `GrantRound` constructor arguments


mds1:
The following should be verified in the constructor. Recommend waiting for #21 to be completed first since there will likely be some merge conflicts, as the tests should be probably refactored here to use Waffle fixtures to simplify snapshots/reverts for going forwards/backwards in time

```solidity
require(_donationToken.totalSupply() > 0, "GrantRound: Invalid token");
require(_startTime >= block.timestamp, "GrantRound: Start time has already passed");
require(_endTime > _startTime, "GrantRound: End time must be after start time");
require(_registry.grantCount() >= 0, "GrantRound: Invalid registry"); // verify this call doesn't revert
```

Issue #21, opened 2021-07-16
Add MockERC20 contract to improve testing


corydickson:
Currently #17, uses Waffle Mock ERC20 framework to set return values for `IERC20` functions called in the `GrantRound` contract. Instead we'd like to introspect into the state of token transfers. This is to ensure that the methods implemented successfully send the tokens and balances are updated correctly; along with handling common revert edge cases.

apbendi:
closed by #48

PR #35, opened 2021-07-22
Add GrantRoundFactory and tests

### LINKED ISSUE #19: Implement Grants Round Factory
This issue is dependent on #17 

We will use the factory pattern to allow users to create new `GrantRound` instance. The factory should:

- [ ] Take all the parameters needed to initialize a new `GrantRound` and deploy one via a public method call
- [ ] Emit an event with the parameters when it does so
- [ ] Remove existing `GrantRounds` contract and tests

Questions:

* Should the factory method or the `GrantRound` constructor do any sanity checking of the address provided as the ERC20 token? For example, calling the `symbol` method, which would cause revert if the address passed was not an ERC20 implementation.

mds1:
Closes #19

In #19 we discussed adding verification to constructor args, but I pulled that out into #34 instead to avoid conflicts with #21 which is currently in progress


mds1:
@wildmolasses Looks like the CI is failing even though tests pass locally, would love if you could help get that fixed!

mds1:
CI fixed after rebasing against #36.

For reference, the problem solved by #36 was: "`yarn` normally installs deps and then runs the typechain build. but since deps didnt change, the cache step is skipping the yarn step. so the typechain stuff isn't built". Thanks @wildmolasses!

phutchins (approved):
Looks good to me and builds.

apbendi (approved):
lgtm

PR #37, opened 2021-07-23
Add grant payout admin role to contract

corydickson:
Addresses #39

### DIFF
--- a/contracts/contracts/GrantRound.sol
+++ b/contracts/contracts/GrantRound.sol
@@ -47,6 +47,7 @@ contract GrantRound {
 
   /**
    * @notice Instantiates a new grant round
+   * @param _owner Grant round owner that has permission to update the metadata pointer
    * @param _payoutAdmin Grant round administrator that has permission to payout the matching pool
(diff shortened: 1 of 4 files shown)


contracts/contracts/GrantRound.sol
```diff
@@ -50,16 +56,17 @@ contract GrantRound {
...
-    address _owner,
+    address _payoutAdmin,
     GrantRegistry _registry,
     IERC20 _donationToken,
     uint256 _startTime,
     uint256 _endTime,
     string memory _metaPtr,
     uint256 _minContribution
   ) {
+    owner = msg.sender;
```
mds1, line 67: PR looks good! Only comment is that `owner` should also be a constructor arg. The `GrantRound` contracts will be deployed from `GrantRoundFactory` which means the owner would be the factory contract

mds1:
Ah one more thing I just realized—`owner` and `payoutAdmin` should both either be `immutable` or have methods to update them. I think it'd be nice if these were immutable, but curious to hear thoughts. cc @phutchins and @apbendi for input

PR #42, opened 2021-07-24
Provide checks in Grant Round Constructor arguments 

corydickson:
Addresses #34


contracts/test/GrantRound.test.ts
```diff
@@ -158,7 +162,7 @@ describe('GrantRound', function () {
       // TODO: add token.balanceOf check in mock ERC20
     });
 
-    it('reverts if not the grant round payout administrator', async function () {
+    it.skip('reverts if not the grant round payout administrator', async function () {
```
corydickson, line 165: I'll be refactoring these tests anyway with the ERC20 mocks but one thing that's trouble some is because the contract now checks the block time stamp when it is deployed (an instance is created before each test), the time travel to a respective grant round instance endTime is now in the past which is problematic for testing. 

Any ideas on how to get around this?

mds1: The way we should do this is with Waffle fixtures: https://ethereum-waffle.readthedocs.io/en/latest/fixtures.html

Basically:
1. Move the content of the `beforeEach` hook into a method called `setup` or something similar
2. The content of the before each hook should now assign all variables e.g. `({registry, roundContract} = await loadFixture(setup);`
3. Adjust time _foward_ in tests as necessary

You can have other fixture methods to load at the beginning of tests too if that's helpful.

The Waffle fixtures use `evm_snapshot` and `evm_revert` methods to revert to chain's state (include timestamp) before each test. That way at the beginning of each test, time resets -> contract deploys -> time can be fast forwarded again if needed

If you need help with this let me know!

contracts/test/utils.ts
```diff
@@ -20,3 +20,9 @@ export async function timeTravel(provider: any, seconds: number) {
   await provider.send('evm_increaseTime', [seconds]);
   await provider.send('evm_mine', []);
 }
+
+export async function setNextBlockTimestamp(provider: any, timestamp: number, delay: number) {
```
mds1, line 24: Instead of passing in a provider to these test helpers (`timeTravel` also, you can just use the globally available `ethers.provider`, where `ethers` is available from `import { ethers } from 'hardhat';`

apbendi (changes requested):
Looks good overall @corydickson! I think, though, that we'd want test cases capturing the deployment failure in each invalid case. Would you mind implementing those?

PR #46, opened 2021-07-27
chore: create dummy grantRound

thelostone-mc:
#### Description 

This creates 
- a dummy GrantRoundFactory
- creates a GrantRound using createGrantRound method on the above deployed contract

 

image


thelostone-mc:
Closing this out and cherry-picking this onto https://github.com/dcgtc/dgrants/tree/add-grant-round-manager

PR #48, opened 2021-07-28
Mock ERC20 tests for GrantRound 

corydickson:
Addresses #21

### DIFF
--- a/contracts/contracts/test/MockToken.sol
+++ b/contracts/contracts/test/MockToken.sol
@@ -1,0 +1,9 @@
+pragma solidity ^0.8.6;
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+contract MockToken is ERC20 {
+  constructor(uint256 initialSupply) public ERC20("Mock", "MCK") {
+    _mint(msg.sender, initialSupply);
+  }
+}
--- a/contracts/contracts/test/MockToken.sol
+++ b/contracts/contracts/test/MockToken.sol
@@ -3,7 +3,11 @@ pragma solidity ^0.8.6;
 import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 
 contract MockToken is ERC20 {
-  constructor(uint256 initialSupply) public ERC20("Mock", "MCK") {
+  constructor(uint256 initialSupply) ERC20("Mock", "MCK") {
+    _mint(msg.sender, initialSupply);
+  }
+
+  function mint(uint256 initialSupply) external {
     _mint(msg.sender, initialSupply);
   }
 }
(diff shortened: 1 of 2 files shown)


contracts/test/GrantRound.test.ts
```diff
@@ -23,21 +25,21 @@ describe('GrantRound', function () {
     mpUser: SignerWithAddress; // matching pool user
   let registry: GrantRegistry;
   let roundContract: Contract;
-  let mockERC20: MockContract;
-  const balances: string[] = ['100', '200'];
+  let mockERC20: MockToken;
+  const balances: string[] = ['100', '2000'];
```
apbendi, line 29: If I'm understanding this right, this is just an array to capture constants for expected balances throughout the test suite. Do I have that right @corydickson? If so, we ought to pull them into two separate named constants. It's definitely a bit annoying, but I think it will make reading the tests easier at the callsite. Right now, seeing `balances[0]` or `balances[1]` seems cryptic, and would read better if it was something named liked `DONOR_AMOUNT` or `DonorAmount` or whatever (I don't feel strongly about the casing convention).

apbendi (commented):
Looks good and tests pass! One comment about naming constants in the test file.

thelostone-mc (approved):

contracts/contracts/test/MockToken.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity ^0.8.6;
+
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+
+contract MockToken is ERC20 {
+  constructor(uint256 initialSupply) public ERC20("Mock", "MCK") {
+    _mint(msg.sender, initialSupply);
+  }
```
mds1, line 8: Could we add an `external` mint function that anyone can call? That way we can mint during a test if needed, and it's a bit easier to fund multiple accounts that way

contracts/test/GrantRound.test.ts
```diff
@@ -149,46 +137,27 @@ describe('GrantRound', function () {
...
-      await mockERC20.mock.balanceOf.withArgs(roundContract.address).returns(ethers.utils.parseEther(balances[0]));
-      await mockERC20.mock.transfer.withArgs(grantPayee.address, balances[0]).returns(true);
-
       await roundContract.connect(payoutAdmin).payoutGrants(grantPayee.address);
-
-      // TODO: add token.balanceOf check in mock ERC20
+      expect(await mockERC20.balanceOf(grantPayee.address)).to.be.equal(ethers.utils.parseEther(balances[0]));
     });
 
     it.skip('reverts if not the grant round payout administrator', async function () {
```
mds1, line 147: Why are these tests skipped? I can't remember what the blocker is for getting these implemented

mds1 (changes requested):
Two small comments, but otherwise LGTM!

contracts/test/GrantRound.test.ts
corydickson: Will be address in #50

Issue #29, opened 2021-07-21
GrantRound UI Spec


apbendi:
- [ ] Round Detail Page
    - [x] route: `/dgrants/rounds/:address` where `:address` is the hex address of a `GrantRound` contract instance
    - [ ] Displays the following information:
        - [x] Round status: Upcoming, Active, Ended (better copy here?)
        - [x] Appropriate time relation to status: "Starts in _____", "Ends in ______", "Ended ______ ago"
        - [x] Hovering over the time description should show a discrete, human readable timestamp for the appropriate time
        - [x] Funding Token (Show name and/or symbol; clicking should take to Etherscan token site?)
        - [ ] Matching Pool Balance OR the amount of the matching pool that was paid out, if it was
        - [x] Owner address (with Etherscan link?)
    - [x] Matching pool contribution form: iff the Round is Upcoming or Active, show a form allowing the user to make a matching pool contribution; specify an amount, then sign/submit the token approval and `addMatchingFunds` method call txs
    - [ ] Payout contract destination: iff the Round `hasPaidOut`, show the contract address (and transaction hash?) where the payout was made to. Link to Etherescan.
- [x] Round List Page: dependent on #19 
     - [x] ~~route: `/dgrants/rounds/factory/:address` where `:address` is the hex address of a `GrantRoundFactory` contract instance~~ — route: `/dgrants/rounds` (updated with assumption that the Factory is a singleton)
     - [x] Show a list of GrantRound instances deployed by this factory, identified by address (in the future, this will be the Round title as fetched from MetaDat)
     - [x] Show three discrete sections: Active, Upcoming, Ended
     - [x] Sort Grants in each section by: Active: End Date, soonest first; Upcoming: Start Date, soonest first; Ended: End Date, recently ended first
     - [x] Clicking each GrantRound should take the user to the appropriate Grant Detail Page

Additional notes/questions:

* A list of Grants included in this Round is not yet included, because as currently conceived, the list of canonical grants is stored in the metadata
* Should we show a list of matching pool contributions? If so, should we have an event emitted from the `addMatchingFunds` method? Should we have a minimum matching pool contribution threshold?

mds1:
> route: /dgrants/rounds/factory/:address where :address is the hex address of a GrantRoundFactory contract instance

[...]

PR #58, opened 2021-07-28
Implements GrantRoundManager UI (grantRound details & list)

gdixon:
This PR:

- implements the GrantRoundFactory UI 
- adds a task to the hardhat config to send `GTC` and `DAI` tokens from a whale account to the hardhat testing accounts
- adds utilities to get `daysAgo` and to convert `unixToLocaleString`
- adds an ERC20 abi to constants
- updates some minor elements to keep styling consistent

---

https://www.loom.com/share/7006b6f312ad4a3c8c5e5e39251ea3b8

---
Spec as defined @ #29 (closes: #29)

- [ ] Round Detail Page
    - [x] route: `/dgrants/rounds/:address` where `:address` is the hex address of a `GrantRound` contract instance
    - [ ] Displays the following information:
        - [x] Round status: Upcoming, Active, Ended (better copy here?)
        - [x] Appropriate time relation to status: "Starts in _____", "Ends in ______", "Ended ______ ago"
        - [x] Hovering over the time description should show a discrete, human readable timestamp for the appropriate time
        - [x] Funding Token (Show name and/or symbol; clicking should take to Etherscan token site?)
        - [ ] Matching Pool Balance OR the amount of the matching pool that was paid out, if it was \*
        - [x] Owner address (with Etherscan link?)
    - [x] Matching pool contribution form: iff the Round is Upcoming or Active, show a form allowing the user to make a matching pool contribution; specify an amount, then sign/submit the token approval and `addMatchingFunds` method call txs
    - [ ] Payout contract destination: iff the Round `hasPaidOut`, show the contract address (and transaction hash?) where the payout was made to. Link to Etherescan. \*
- [x] Round List Page: dependent on #19 
     - [x] ~~route: `/dgrants/rounds/factory/:address` where `:address` is the hex address of a `GrantRoundFactory` contract instance~~ — route: `/dgrants/rounds` (updated with assumption that the Factory is a singleton)
     - [x] Show a list of GrantRound instances deployed by this factory, identified by address (in the future, this will be the Round title as fetched from MetaDat)
     - [x] Show three discrete sections: Active, Upcoming, Ended
     - [x] Sort Grants in each section by: Active: End Date, soonest first; Upcoming: Start Date, soonest first; Ended: End Date, recently ended first
     - [x] Clicking each GrantRound should take the user to the appropriate Grant Detail Page
     
---

\* As discussed in todays standup, I've sidestepped the parts to do with the payout contract and matching pool balance until we have more clarity on the  `GrantRoundManager` changes


app/src/store/data.ts
```diff
@@ -48,6 +59,53 @@ export default function useDataStore() {
...
+
+        return {
+          address: grantRoundAddress,
+          startTime: startTime,
+          endTime: endTime,
+          status: status,
+          donationToken: donationToken,
+          donationTokenName: donationTokenName,
+          donationTokenSymbol: donationTokenSymbol,
+          donationTokenDecimals: donationTokenDecimals,
```
mds1, line 99: My personal preference is always pass around tokens as the [`TokenInfo`](https://github.com/Uniswap/token-lists/blob/fdf9266aefedb9927b182faa8e673a0c29385892/src/types.ts#L1-L12) type defined in Uniswap's [`@uniswap/token-lists`](https://github.com/Uniswap/token-lists) package. It's a nice standard that removes any ambiguity around whether e.g. `donationToken: string;` refers to a symbol, name, or address

In this case we could replace these four fields with:
```suggestion
          donationToken: {
            chainId: <current chainId user is connected to, or 1 otherwise>,
            address: donationToken,
            name: donationTokenName,
            symbol:  donationTokenSymbol,
            decimals: donationTokenDecimals,
            logoURI: null, // TODO we can leave this out for now, and leave a TODO to fetch it from CoinGecko's huge token list (as well as use that to avoid a network request for token info each poll): https://tokenlists.org/token-list?url=https://tokens.coingecko.com/uniswap/all.json
          }
```

app/src/utils/utils.ts
```diff
@@ -25,3 +25,20 @@ export function isValidUrl(val: string) {
 export function isValidAddress(val: string) {
   return isAddress(val);
 }
+
+// Expects a unix timestamp and will return a human readable message of how far in the past/future it is
+export function daysAgo(val = 0) {
+  // use a formatter to establish "in 10 days" vs "10 days ago"
+  const formatter = new Intl.RelativeTimeFormat();
```
mds1, line 32: TIL about this, that's handy. Thanks!

app/src/utils/utils.ts
```diff
@@ -25,3 +25,20 @@ export function isValidUrl(val: string) {
...
+  const deltaDays = (val * 1000 - Date.now()) / (1000 * 3600 * 24);
+
+  // Format days to string (rember to round off deltaDays)
+  return formatter.format(Math.round(deltaDays), 'days');
+}
+
+// convert a unix ts to a toLocaleString
+export const unixToLocaleString = (time: BigNumber) => {
+  return new Date(time.toNumber() * 1000).toLocaleString();
+};
```
mds1, line 44: May as well make this a bit more generic. Also might want to change this from `const` to `function` to be consistent with the rest of the file. (or we can change them all to arrow functions if you prefer, for this utils file I don't think it matters much)

```suggestion
export function unixToLocaleString(time: BigNumberish) {
  return new Date(BigNumer.from(time).toNumber() * 1000).toLocaleString();
};
```

app/src/views/GrantRoundDetails.vue
```diff
@@ -0,0 +1,254 @@
...
+function useGrantRoundDetail() {
+  const { grantRounds, poll } = useDataStore();
+  const { signer, userAddress } = useWalletStore();
+  const route = useRoute();
+
+  // get a single grantRound or an empty/error object (TODO: should the typings be modified to account for an empty object?)
+  const grantRound = computed(() => {
+    if (grantRounds.value) {
+      // filter for a matching GrantRound
+      const round = grantRounds.value.filter((round) => round.address == route.params.address);
```
mds1, line 130: Looks like `==` is used throughout the file, but I typically prefer strict equality `===` if that's good with you. Helps avoid implicit type casting we might not want.

Also with address checks, one thing to be careful of is checksum address. Ethers always uses checksummed addresses so `round.address` will be checksummed, meaning if someone navigates to `/dgrants/rounds/0xabc...def` and the checksummed address is `0xAbC...dEF`, this check will fail. 

ethers has a method called `getAddress` which takes an address and returns the checksummed version, which we should use here

app/src/views/GrantRoundDetails.vue
```diff
@@ -0,0 +1,254 @@
...
+    const { token, amount } = form.value;
+    const areFieldsValid =
+      isValidAddress(<string>token) && token == grantRound.value.donationToken && (amount || 0 > 0);
+    return areFieldsValid;
+  });
+
+  /**
+   * @notice Check against the grantRounds status for a match
+   */
+  const hasStatus = (status: string) => (round: GrantRound) => round.status == status;
```
mds1, line 156: Just curious, why this syntax instead of `const hasStatus = (status: string, round: GrantRound) => round.status == status` ?

app/src/views/GrantRoundDetails.vue
```diff
@@ -0,0 +1,254 @@
...
+import useDataStore from 'src/store/data';
+import useWalletStore from 'src/store/wallet';
+// --- Methods and Data ---
+import { GRANT_ROUND_ABI, ERC20_ABI } from 'src/utils/constants';
+import { BigNumber, Contract, ContractTransaction } from 'src/utils/ethers';
+import { daysAgo, formatAddress, isValidAddress, isValidUrl } from 'src/utils/utils';
+// --- Types ---
+import { GrantRound } from '@dgrants/types';
+import { GrantRound as GrantRoundContract } from '@dgrants/contracts';
+import { BigNumberish, utils } from 'ethers';
```
mds1, line 119: Can we get these from the `src/utils/ethers` file also? That way all ethers imports are in one place (might need to add new exports to that file, and only export the specific utils we need to keep bundle size down)

app/src/views/GrantRoundDetails.vue
```diff
@@ -0,0 +1,254 @@
...
+
+    // poll for the updated state and toggle page state back to display mode
+    await poll();
+    cancelContribution();
+  }
+
+  /**
+   * @notice Check for approved allowance
+   */
+  async function checkAllowance(token: Contract, address: string) {
```
mds1, line 206: This method and `getApproval` feel generic enough that we should probably pull them out into generic helper methods in `utils`

app/src/views/GrantRoundDetails.vue
```diff
@@ -0,0 +1,254 @@
...
+    const token = new Contract(grantRound.value.donationToken, ERC20_ABI, signer.value);
+    const round = <GrantRoundContract>new Contract(grantRound.value.address, GRANT_ROUND_ABI, signer.value);
+
+    // contributionAmount must have the right number of decimals and be hexed
+    const contributionAmount = utils.parseUnits(amount.toString(), grantRound.value.donationTokenDecimals);
+
+    // check if contract is already approved as a spender
+    const allowance = await checkAllowance(token, grantRound.value.address);
+    if (allowance < contributionAmount) {
+      await getApproval(token, grantRound.value.address, contributionAmount);
```
mds1, line 192: My personal preference is for approval amount to be `ethers.constants.MaxUint256` since contracts won't be upgradeable, so this will save user's gas in the long run

contracts/scripts/app.ts
```diff
@@ -30,6 +30,26 @@ const grants = [
...
+    owner: '0x9B4d61BB9b8B6CB33a75B9846f7E2C8d67d137fa',
+    token: '0xDe30da39c46104798bB5aA3fe8B9e0e1F348163F', // GTC address
+    startTime: '1627374646',
+    endTime: '1630079135',
+    metaPtr: 'https://metadata-pointer.com',
+    minContribution: '100',
+  },
+  {
+    owner: '0x9B4d61BB9b8B6CB33a75B9846f7E2C8d67d137fa',
+    token: '0x6b175474e89094c44da98b954eedeac495271d0f', // DAI address
```
mds1, line 45: Let's use the checksummed address here (the rest seem to already be checksummed)
```suggestion
    token: '0x6B175474E89094C44Da98b954EedeAC495271d0F', // DAI address
```

package.json
```diff
@@ -22,7 +22,7 @@
     "build": "lerna run build",
     "clean": "lerna run clean",
     "node": "yarn contracts app:node",
-    "dev": "concurrently --restart-tries 5 --kill-others \"yarn run node\" \"yarn contracts app:setup && yarn app dev\"",
+    "dev": "concurrently --restart-tries 5 --kill-others \"yarn run node\" \"yarn contracts app:setup && yarn contracts app:load && yarn app dev\"",
```
mds1, line 25: Why fund the accounts in a separate script instead of in the `app.ts` script? My only thought is that it might be nice to do this funding first in the same file, as it _might_ be easier to initialize with various states as things get more complex. I think this is fine though and don't feel strongly either way!

types/src/grants.d.ts
```diff
@@ -12,3 +12,20 @@ export type GrantObject = {
...
 export type GrantEthers = GrantArray & GrantObject;
 export type Grant = GrantObject | GrantEthers;
+export type GrantRound = {
+  address: string,
+  owner: string,
+  registry: string,
+  donationToken: string,
+  donationTokenName: string,
+  donationTokenSymbol: string,
+  donationTokenDecimals: BigNumberish,
```
mds1, line 22: Same comment as earlier on using `TokenInfo` type

mds1 (changes requested):
This is awesome! Really work on this.  I left a bunch of comments, but they're mostly small things related to consistency / style choices / best practices

>  Funding Token (Show name and/or symbol; clicking should take to Etherscan token site?)

One other question—where is this? I don't see this functionality

contracts/scripts/app.ts
```diff
@@ -43,6 +63,26 @@ async function main(): Promise<void> {
...
   console.log(`Deployed GrantRegistry to ${registry.address}`);
 
+  const GrantRoundFactory: ContractFactory = await ethers.getContractFactory('GrantRoundFactory', deployer);
+  const roundFactory = await (await GrantRoundFactory.deploy()).deployed();
+  console.log(`Deployed GrantRoundFactory to ${roundFactory.address}`);
+
+  // create the rounds
+  await Promise.all(
+    rounds.map((round) =>
+      roundFactory.createGrantRound(
```
apbendi, line 73: Looks like a rebase may have broken this. When I go to run `yarn dev`, it errors here because we're missing the `payoutAdmin` argument for deployment.

Additionally, after modifying this call to include the missing parameter, the UI starts, but does not show any Rounds at `/dgrants/rounds/`:

Is the app making an assumption about the address of the factory? Do I need to use a specific mnemonic to match that dpeloyment address, for example?

mds1: Related: `owner: round.owner()` on L101 of `data.ts` is also broken. We now have two separate privileged roles so that should be changed to:

```javascript
payoutAdmin: await round.payoutAdmin(),
metadataAdmin: await round.metadataAdmin()
```

contracts/tasks/load-accounts.ts
```diff
@@ -0,0 +1,42 @@
...
+import { parseEther, parseUnits } from 'ethers/lib/utils';
+import { task } from 'hardhat/config';
+
+import { TASK_LOAD_ACCOUNTS } from './task-names';
+
+export const fundAccount = async (hre: any, tokenAddress: string, amount: string, to: string) => {
+  const tokenAbi = ['function transfer(address,uint256) returns (bool)', 'function decimals() view returns (uint8)'];
+
+  // We impersonate the binance exchange account on mainnet as a source of tokens
+  const funderAddress = '0x28C6c06298d514Db089934071355E5743bf21d60';
```
mds1, line 12: I know this is the approach I sent on Discord, but you actually might want to take an approach similar to what I do on the `add-grant-round-manager` branch [here](https://github.com/dcgtc/dgrants/blob/29046d40cfcbbbb6916e6d001bae5913662d455e/contracts/test/utils.ts#L72-L83). It's more robust because since it will work for any token without worrying about Binance's or if Binance changes their wallet address (though you do need to manually add support per-token)

If you rebase on top of that branch you could just import those methods and use them directly

contracts/scripts/app.ts
apbendi: Yes, I think that's correct @mds1. When I added an extra param, it worked, but did not show Rounds in the UI, which is the second part of my comment.

app/src/views/GrantRoundDetails.vue
gdixon: Ahh, I did that to tidy up the filtering over in `GrantRoundsList.vue` and copied the same fn into this file. Would it make sense to keep this in `utils.ts` - I'm not sure its generic enough?

We could just do as you say here, theres no need for currying with the way we're using it.

mds1: Since it's used in two different places, I think it's generic enough to move into the `utils.ts` file!

app/src/store/data.ts
```diff
@@ -48,6 +59,53 @@ export default function useDataStore() {
     lastBlockNumber.value = (blockNumber as BigNumber).toNumber();
     lastBlockTimestamp.value = (timestamp as BigNumber).toNumber();
     grants.value = grantsList as Grant[];
+
+    // Get all rounds from GrantRoundCreated --- TODO: We need to cache these events somewhere (like the graph)
+    const roundList = await roundFactory.value.queryFilter(roundFactory.value.filters.GrantRoundCreated());
+    const roundAddresses = [...roundList.map((e) => e.args?.grantRound)];
+
+    // Pull state from each GrantRound
+    grantRounds.value = await Promise.all(
```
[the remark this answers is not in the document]
gdixon: Resolved in b35f417

app/src/store/data.ts
```diff
@@ -48,6 +59,53 @@ export default function useDataStore() {
...
+      roundAddresses.map(async (grantRoundAddress) => {
```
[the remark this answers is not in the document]
gdixon: Resolved in b35f417

app/src/store/data.ts
```diff
@@ -48,6 +59,53 @@ export default function useDataStore() {
...
+        const donationTokenSymbol = await token.symbol();
+        const donationTokenDecimals = await token.decimals();
+        // get the rounds balance (in whole units)
+        const funds = (await token.balanceOf(grantRoundAddress)) / 10 ** donationTokenDecimals;
+        // discover status from startTime/endTime
+        const now = Date.now();
+        const startTime = await round.startTime();
+        const endTime = await round.endTime();
+        const status =
+          now > startTime.toNumber() * 1000 && now < endTime.toNumber() * 1000
```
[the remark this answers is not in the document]
gdixon: Resolved in b35f417

app/src/store/data.ts
gdixon: Resolved in b35f417

app/src/views/GrantRoundDetails.vue
gdixon: Resolved in d8da394

app/src/utils/utils.ts
gdixon: Resolved in b35f417

app/src/views/GrantRoundDetails.vue
gdixon: Resolved in 85d0d88

app/src/views/GrantRoundDetails.vue
gdixon: Resolved in f095fea

app/src/views/GrantRoundDetails.vue
gdixon: Resolved in 85d0d88

app/src/views/GrantRoundDetails.vue
gdixon: Resolved in 85d0d88

contracts/scripts/app.ts
gdixon: Resolved by the rebase (but I did fix up `GRANT_REGISTRY_ADDRESS` in 9723b29)

package.json
gdixon: Your `setBalance` fn is a lot faster than the other approach so I've replaced it with that! (5871fbb)

gdixon: Your setBalance fn is a lot faster than the other approach so I've replaced it with that! (5871fbb) 🙌

types/src/grants.d.ts
gdixon: Resolved in b35f417

contracts/scripts/app.ts
gdixon: @apbendi - The name change will likely have caused the rpc calls to fail - all fixed up and working as expected in b35f417 🙌

contracts/tasks/load-accounts.ts
gdixon: Awesome, I've gone with this approach 🙌 - resolved in 5871fbb

thelostone-mc (approved):

apbendi:
Hey @gdixon, sorry for the continued churn on this PR, but the contracts are changing out from under you. With #85 matching merged, this PR needs to be rebased, and will need updated in a few ways when it is:

- [ ] When deploying a round, you need to include an extra argument for the [matching token](https://github.com/dcgtc/dgrants/blob/main/contracts/contracts/GrantRoundManager.sol#L78). (It would probably be good to make this something other than GTC in our dev environment data, so we are properly testing the real-world scenario of different matching vs. donation tokens).
- [ ] The matching token should be [read](https://github.com/dcgtc/dgrants/blob/main/contracts/contracts/GrantRound.sol#L31) from the GrantRound so that matching funds are displayed in the right token on the detail page
- [ ] When the user contributes to the matching pool, it should use the correct token as well

Let me know if any of this doesn't make sense, or if I missed anything here.

thelostone-mc:
@apbendi 

> When deploying a round, you need to include an extra argument for the matching token. (It would probably be good to make this something other than GTC in our dev environment data, so we are properly testing the real-world scenario of different matching vs. donation tokens).

I'm curious so as to why we made this change 
Wouldn't the matchingToken be the same as the donationToken always ?

apbendi:
> @apbendi / @gdixon before addressing that could you guys check out #88

@thelostone-mc Matt beat me to the explanation. We discussed this on standup and that scenario he describes was one of the primary reasons we went the split token route.

apbendi (approved):
Looking and working great. Nice work @gdixon 👏 👏 👏

PR #59, opened 2021-07-29
Time dependent fixtures

### LINKED ISSUE #50: Add waffle fixture to test time-dependent GrantRound methods
Currently, tests use `timeTravel` and `setNextBlockTimestamp` to modify blocks. Instead waffle fixtures could be used to snapshot and revert changes made to the chain before each test

corydickson:
Closes #50

### DIFF
--- a/contracts/test/GrantRound.test.ts
+++ b/contracts/test/GrantRound.test.ts
@@ -9,10 +9,10 @@ import { GrantRegistry } from '../typechain/GrantRegistry';
 import { MockToken } from '../typechain/MockToken';
 import { timeTravel, setNextBlockTimestamp } from './utils';
 
-import { Contract } from 'ethers';
+import { ContractFactory } from 'ethers';
 
 // --- Parse and define helpers ---
-const { deployContract } = waffle;
+const { loadFixture, deployContract } = waffle;
 const { MaxUint256 } = ethers.constants;
 const { isAddress } = ethers.utils;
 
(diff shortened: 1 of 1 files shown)


contracts/test/GrantRound.test.ts
```diff
@@ -71,23 +68,28 @@ describe('GrantRound', function () {
...
     await mockERC20.connect(donor).approve(roundContract.address, MaxUint256);
     await mockERC20.connect(payoutAdmin).approve(roundContract.address, MaxUint256);
-  });
+
+    return { mockERC20, roundContract };
+  }
 
   describe('Initialization', () => {
     it('deploys properly', async function () {
+      const { roundContract } = await loadFixture(setup);
```
mds1, line 77: I think best practice for fixtures would be to avoid assigning variables outside of the fixture's scope within the fixture (e.g. like we do with `endTime` and `startTime` on L54-55), and instead load the fixture in a before each hook that assigns all the variables there. 

Plus that way you also don't need to call `loadFixture` in each function. I know the Waffle docs show it that way, but Uniswap does it the other way, e.g. [here](https://github.com/Uniswap/uniswap-v3-core/blob/b2c5555d696428c40c4b236069b3528b2317f3c1/test/UniswapV3Factory.spec.ts#L29-L41). I'm not sure what they need the `createFixtureLoader` method for. Would that be useful for us in any way?

mds1 (changes requested):
One small comment on best practices but otherwise LGTM. You might want to combine this PR with #60 just to avoid any conflicts once either is merged. Up to you, but feel free to rebase one on top of the other if you think it'll make things easier

contracts/test/GrantRound.test.ts
corydickson: I will try this approach but my impression is that we will still need the global variables to access the contract object to make calls as well as signers.

mds1: Yea they will still be global, just would be set in the before each hook instead of in the test itself

```javascript
let roundContract;

beforeEach(async () => { ({ roundContract }) = await loadFixture(setup) });
```

corydickson: Just dug into this and as written couldn't use this snippet. Also `createFixtureLoader` currently takes wallets and not signers unfortunately.

PR #60, opened 2021-07-30
Add revert corner cases for constructor args

corydickson:


mds1 (changes requested):
Looks like we're missing a test for `GrantRound: Invalid registry`, but other this PR looks good!

Easiest way to test that would be to just pass an EOA as the registry address and that call should revert. But it might actually revert with a different message (like `function call to a non-contract account`), which is ok, because we just want to make sure construction fails if the registry is invalid

PR #61, opened 2021-07-30
Add grant round manager

### LINKED ISSUE #40: Grant Round Manager Spec
- [x] Rename the `GrantRoundFactory` to `GrantRoundManager`
- [x] Add an immutable variable `donationToken` to the Manager contract that is defined at deployment
- [x] Remove `_donationToken` from `createGrantRound`, instead using the immutable `donationToken` when creating a Round
- [x] Add an immutable variable `registry`, an instance of `GrantRegistry` to the Manager contract that is defined at deployment
- [x] Add an immutable variable `router`, an instance of the Uniswap v3 Router contract that is defined at deployment
- [ ] Define a new `Donation` struct as follows:

```solidity
// Define inputs required for each contribution in a struct
struct Donation {
  uint96 grantId; // grantId
  uint256 amount; // amount of that token
  address[] swapPath; // Uniswap router swap path
  address[] rounds; // rounds they want the contribution to count for
}
```

- [ ] Create a new method, `swapAndDonate`, which:
    - [ ]  Takes a a `Donation` struct as a parameter
    - [ ] Verifies the `grantId` is valid
    - [ ] Iterates over each `round`, validating it:
        - [ ] Is an active round (between start and end time)
        - [ ] Uses the `donationToken` defined by this Manager
    - [ ] IF `swapPath[0]` defined in `Donation` is the same as the Manager's `donationToken`, uses `transferFrom` to move the token to the Grant's `payee` address
    - [ ] Otherwise:
        - [ ] Use `transferFrom` to move `amount` of the token `swapPath[0]` to the Manager
        - [ ] Use `router` and `swapPath` to swap `swapPath[0]` for `donationToken`
        - [ ] Use `transfer` to send the `donationToken.balanceOf(this)` to the Grant's `payee` address
    - [ ] In both cases: emit an event with the `grantId`, `swapPath[0]` token, `amount`, and `rounds` included
- [ ] Remove `donateToGrant` from the `GrantRound` contract

mds1:
Closes #40 

- [x] Rename the `GrantRoundFactory` to `GrantRoundManager`
- [x] Add an immutable variable `donationToken` to the Manager contract that is defined at deployment
- [x] Remove `_donationToken` from `createGrantRound`, instead using the immutable `donationToken` when creating a Round
- [x] Add an immutable variable `registry`, an instance of `GrantRegistry` to the Manager contract that is defined at deployment
- [x] Add an immutable variable `router`, an instance of the Uniswap v3 Router contract that is defined at deployment
- [x] Define a new `Donation` struct as required for compatibility with Uniswap V3
- [x] Create a new method, `swapAndDonate`, which:
    - [x]  Takes a a `Donation` struct as a parameter
    - [x] Verifies the `grantId` is valid
    - [x] Iterates over each `round`, validating it:
        - [x] Is an active round (between start and end time)
        - [x] Uses the `donationToken` defined by this Manager
    - [x] IF `tokenIn` defined in `Donation` is the same as the Manager's `donationToken`, uses `transferFrom` to move the token to the Grant's `payee` address
    - [x] Otherwise:
        - [x] Use `transferFrom` to move `amount` of the token `tokenIn` to the Manager
        - [x] Use `router` to swap `tokenIn` for `donationToken`
    - [x] In both cases: emit an event with the `grantId`, `swapPath[0]` token, `amountOut`, and `rounds` included
- [x] Remove `donateToGrant` from the `GrantRound` contract

Current limitations:
- Donations must have a direct pair. For example, there is an ETH/GTC pair on Uniswap V3, so this contract will let me donate ETH and will swap it to GTC. However, there is no DAI/GTC pair so donating DAI is currently not possible if GTC is the donation token
- ETH as the donation token is not supported (but you can use WETH)

### DIFF
--- a/contracts/contracts/GrantRoundManager.sol
+++ b/contracts/contracts/GrantRoundManager.sol
@@ -2,17 +2,37 @@
 pragma solidity ^0.8.5;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+import "@uniswap/v3-periphery/contracts/interfaces/ISwapRouter.sol";
+import "./GrantRegistry.sol";
 import "./GrantRound.sol";
 
 contract GrantRoundManager {
+  using Address for address;
+
+  /// @notice Address of the GrantRegistry
+  GrantRegistry public immutable registry;
+
+  /// @notice Address of the Uniswap V3 Router used for token swaps
+  ISwapRouter public immutable router;
+
   /// @notice Address of the ERC20 token in which donations are made
   IERC20 public immutable donationToken;
 
   /// @notice Emitted when a new GrantRound contract is created
   event GrantRoundCreated(address grantRound);
 
-  constructor(IERC20 _donationToken) {
+  constructor(
+    GrantRegistry _registry,
+    ISwapRouter _router,
+    IERC20 _donationToken
+  ) {
+    require(_registry.grantCount() >= 0, "GrantRoundManager: Invalid registry");
+    require(address(_router).isContract(), "GrantRoundManager: Invalid router"); // Router interface doesn't have a state variable to check
     require(_donationToken.totalSupply() > 0, "GrantRoundManager: Invalid token");
+
+    registry = _registry;
+    router = _router;
     donationToken = _donationToken;
   }
 
--- a/contracts/contracts/GrantRoundManager.sol
+++ b/contracts/contracts/GrantRoundManager.sol
@@ -1,18 +1,26 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity ^0.8.5;
 
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "./GrantRound.sol";
 
 contract GrantRoundManager {
+  /// @notice Address of the ERC20 token in which donations are made
+  IERC20 public immutable donationToken;
+
   /// @notice Emitted when a new GrantRound contract is created
   event GrantRoundCreated(address grantRound);
+
+  constructor(IERC20 _donationToken) {
+    require(_donationToken.totalSupply() > 0, "GrantRoundManager: Invalid token");
+    donationToken = _donationToken;
+  }
 
   /**
    * @notice Creates a new GrantRound
    * @param _owner Grant round owner that has permission to update the metadata pointer
    * @param _payoutAdmin Grant round administrator that has permission to payout the matching pool
    * @param _registry Address that contains the grant metadata
-   * @param _donationToken Address of the ERC20 token in which donations are made
    * @param _startTime Unix timestamp of the start of the round
    * @param _endTime Unix timestamp of the end of the round
    * @param _metaPtr URL pointing to the grant round metadata
@@ -22,7 +30,6 @@ contract GrantRoundManager {
     address _owner,
     address _payoutAdmin,
     GrantRegistry _registry,
-    IERC20 _donationToken,
     uint256 _startTime,
     uint256 _endTime,
     string memory _metaPtr,
@@ -32,7 +39,7 @@ contract GrantRoundManager {
       _owner,
       _payoutAdmin,
       _registry,
-      _donationToken,
+      donationToken,
       _startTime,
       _endTime,
       _metaPtr,
--- a/contracts/contracts/GrantRoundManager.sol
+++ b/contracts/contracts/GrantRoundManager.sol
@@ -3,7 +3,7 @@ pragma solidity ^0.8.5;
 
 import "./GrantRound.sol";
 
-contract GrantRoundFactory {
+contract GrantRoundManager {
   /// @notice Emitted when a new GrantRound contract is created
   event GrantRoundCreated(address grantRound);
 
--- a/contracts/contracts/GrantRoundManager.sol
+++ b/contracts/contracts/GrantRoundManager.sol
@@ -10,6 +10,15 @@ import "./GrantRound.sol";
 contract GrantRoundManager {
   using Address for address;
 
+  /// @notice Donation Object
+  struct Donation {
+    uint96 grantId; // grant ID to which donation is being made
+    uint256 amount; // amount of tokens which are being donated
+    address[] swapPath; // Uniswap router swap path
+    // WHY IS THIS AN ARRAY ? CAN THIS BE IERC20
+    GrantRound[] rounds; // rounds against which the donation should be counted
+  }
+
   /// @notice Address of the GrantRegistry
   GrantRegistry public immutable registry;
 
@@ -21,6 +30,9 @@ contract GrantRoundManager {
 
   /// @notice Emitted when a new GrantRound contract is created
   event GrantRoundCreated(address grantRound);
+
+  /// @notice Emit when a donation has been made
+  event GrantDonation(uint96 grantId, address inputToken, uint256 amount, address[] rounds);
 
   constructor(
     GrantRegistry _registry,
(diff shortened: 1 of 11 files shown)


contracts/contracts/GrantRoundManager.sol
```diff
@@ -0,0 +1,171 @@
...
+  /// @notice Donation Object
+  struct Donation {
+    uint96 grantId; // grant ID to which donation is being made
+    GrantRound[] rounds; // rounds against which the donation should be counted
+    IERC20 tokenIn; // token in which the user made the donation
+    uint24 fee; // selected fee tier
+    uint256 deadline; // deadline by when swap has to be happen
+    uint256 amountIn; // amount donated by the user
+    uint256 amountOutMinimum; // minimum amount to be returned after swap
+    uint160 sqrtPriceLimitX96; // determine limits on the pool prices which cannot exceed swap
```
apbendi, line 23: Can we expand on what this is about, maybe link to appropriate docs?

contracts/contracts/GrantRoundManager.sol
```diff
@@ -0,0 +1,171 @@
...
+    require(_rounds.length > 0, "GrantRoundManager: Must specify at least one round");
+
+    // Only allow value to be sent if the input token is WETH
+    IERC20 _tokenIn = _donation.tokenIn;
+    require(
+      (msg.value == 0 && address(_tokenIn) != WETH) || (msg.value > 0 && address(_tokenIn) == WETH),
+      "GrantRoundManager: Invalid token-value pairing"
+    );
+
+    // Wnsure grant recieving donation exists in registry
```
apbendi, line 114: Typo `Wnsure` -> `Ensure`

contracts/contracts/GrantRoundManager.sol
```diff
@@ -0,0 +1,171 @@
...
+    //   - The round has the same donationToken as the GrantRoundManager
+    //   - The round is active
+    for (uint256 i = 0; i < _rounds.length; i++) {
+      require(
+        donationToken == _rounds[i].donationToken(),
+        "GrantRoundManager: GrantRound's donation token does not match GrantRoundManager's donation token"
+      );
+
+      require(
+        block.timestamp >= _rounds[i].startTime() && block.timestamp < _rounds[i].endTime(),
```
apbendi, line 128: Does it make sense to add a helper method to the `GrantRound` like `isActive()` or `status()` or something similar, rather than redefining what it means to be "active" here?

contracts/contracts/GrantRoundManager.sol
```diff
@@ -0,0 +1,171 @@
...
+        "GrantRoundManager: GrantRound is not active"
+      );
+    }
+
+    // --- Swap ---
+    address _payoutAddress = registry.getGrantPayee(_grantId);
+    uint256 _amountIn = _donation.amountIn;
+    uint256 _amountOut = _amountIn; // by default, by may be overwritten in the swap branch below
+
+    if (_tokenIn == donationToken && msg.value == 0) {
```
apbendi, line 138: Unclear what checking for `msg.value == 0` accomplishes. If it's a sanity check, then it would appear we're missing the corresponding sanity check in the `else` branch, which seems to assume without validating— if I'm reading it right— that if there `msg.value` being sent, then `_tokenIn` must be the 0xEeE address for ETH.

apbendi (commented):
Some questions on the contracts. Overall looking good!

contracts/contracts/GrantRoundManager.sol
corydickson: I agree but shouldn't the GrantRound contract be reverting here instead? What other contracts depend on this status state variable?

contracts/contracts/GrantRoundManager.sol
```diff
@@ -0,0 +1,171 @@
...
+  using Address for address;
+  using SafeERC20 for IERC20;
+
+  /// @notice Donation Object
+  struct Donation {
+    uint96 grantId; // grant ID to which donation is being made
+    GrantRound[] rounds; // rounds against which the donation should be counted
+    IERC20 tokenIn; // token in which the user made the donation
+    uint24 fee; // selected fee tier
+    uint256 deadline; // deadline by when swap has to be happen
```
corydickson, line 20: nit: have the comment read: "unix timestamp that represents when the swap must be executed"

contracts/contracts/GrantRoundManager.sol
apbendi: > shouldn't the GrantRound contract be reverting here instead

Mmmm I'm not necessarily sure I agree with this. Why would we want the Round to revert? I was thinking along the lines of a public view method.

> What other contracts depend on this status state variable?

Good question. I'm not sure any right now, but it just seems useful to have the Round define canonically when it is active, as opposed to anything that integrates with the Round having to repeatedly define that logic itself.

mds1: We don't actually call into the `GrantRound` anymore, so agreed with @apbendi that we shouldn't expect it to revert

> Does it make sense to add a helper method to the `GrantRound` like `isActive()` or `status()` or something similar, rather than redefining what it means to be "active" here?

Yea I like that idea, will do!

contracts/contracts/GrantRoundManager.sol
mds1: We don't support a `donationToken` of ETH, so we want to make sure if ETH is sent we don't attempt a transfer.

The missing branch would be added if/when we support ETH, and would do an ETH transfer instead of `_tokenIn.safeTransferFrom()`. That would also require the special ETH_ADDRESS to be defined which we don't have, so I figured we don't need that branch yet.

> we're missing the corresponding sanity check in the `else` branch

Sorry, which check are we missing? In that branch, we just need `_tokenIn != donationToken` and it supports any `msg.value`, so it seems covered

mds1: I guess we could revert if `_tokenIn == donationToken && msg.value > 0` 🤔

mds1: Plus this will all get modified soon anyway, so not too worried about it just yet!

contracts/contracts/GrantRoundManager.sol
mds1: Linked to Uniswap docs in 02163ce

contracts/contracts/GrantRoundManager.sol
mds1: Fixed in 02163ce

contracts/contracts/GrantRoundManager.sol
mds1: Done in 02163ce

contracts/contracts/GrantRoundManager.sol
mds1: Done in 79cbe2a

contracts/contracts/GrantRoundManager.sol
```diff
@@ -0,0 +1,167 @@
...
+      (msg.value == 0 && address(_tokenIn) != WETH) || (msg.value > 0 && address(_tokenIn) == WETH),
+      "GrantRoundManager: Invalid token-value pairing"
+    );
+
+    // Ensure grant recieving donation exists in registry
+    uint96 _grantId = _donation.grantId;
+    require(_grantId < registry.grantCount(), "GrantRoundManager: Grant does not exist in registry");
+
+    // Iterate through each GrantRound to verify:
+    //   - The round has the same donationToken as the GrantRoundManager
```
phutchins, line 119: What is the reason we check to make sure the round has the same donationToken as the GrantRoundManager? If it's not taken in as a parameter it shouldn't be possible to not be the same.

phutchins (approved):
Looks good, just had one question but shouldn't hold up merge.

contracts/contracts/GrantRoundManager.sol
apbendi: > I guess we could revert if _tokenIn == donationToken && msg.value > 0, but Uniswap will fail anyway in that case. Need to check if it fails with a meaningful message

Didn't realize Uniswap would fail. We should at least document that with a comment or the like.

> Plus this will all get modified soon anyway, so not too worried about it just yet!

Agreed.

contracts/contracts/GrantRoundManager.sol
mds1: The round address is an input, so any round can be passed in. If the round was not created from this particular `GrantRoundManager` it may have a different token

Issue #68, opened 2021-08-01
Add a Logout Button in the App UI


jfarid27:
While working on #14 I noticed there are no logout buttons. Minor issue but it seems like a nice to have.

thelostone-mc:
Closing this out as this ended up being removed in the final design we have for the POC! Will reopen if we decide as a team that this needs to be in

PR #69, opened 2021-08-01
Add a Logout button to the UI

jfarid27:
# Description

Adds a switch icon next to the user display name. When clicked, it allows the user to switch their wallet.

# Screenshots

Screen Shot 2021-08-02 at 11 21 21 PM

# Issue

#68

### DIFF
--- a/app/src/components/LayoutHeader.vue
+++ b/app/src/components/LayoutHeader.vue
@@ -22,7 +22,10 @@
         <div class="ml-10 space-x-4">
           <div v-if="userDisplayName" class="text-gray-700">
             {{ userDisplayName }}
-            <XIcon @click="disconnectWallet" class="h-5 w-5 text-yellow-500 mr-2 inline-block cursor-pointer" />
+            <SwitchHorizontalIcon
+              @click="resetWallet"
+              class="h-5 w-5 text-yellow-500 mr-2 inline-block cursor-pointer"
+            />
           </div>
           <div v-else-if="!isSupportedNetwork" class="flex items-center">
             <ExclamationIcon class="h-5 w-5 text-yellow-500 mr-2" />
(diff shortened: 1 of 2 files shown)


apbendi:
Hey @jfarid27, thanks so much for your interest in contributing! At first glance this looks good, but I'd love to hear a bit more about why a "logout" button might be useful. I'm not sure I've seen this pattern on other dapps. Typically I think frontends allow you to change your wallet connection. Can you expand a bit on why you see this as useful? Thanks!

jfarid27:
@apbendi This is just a nice to have. I actually agree that maybe a way to switch accounts would be more appropriate. Let me attempt to work that out since there is a way to do this glancing at the Onboard demo.

jfarid27:
@apbendi I have updated the code to allow the user to switch wallets when the button is clicked. Now it looks like this

Screen Shot 2021-08-02 at 11 21 21 PM

Feel free to close if this isn't wildly necessary. Again this just seemed like a nice to have when testing for the previous issue.

mds1:
Thanks @jfarid27! This is good functionality to have.

Were you able to switch wallets successfully? I just tried connecting with Torus and then switching to MetaMask and get the below issue, where the modal says I'm now connected to MetaMask, but the address in the top right still shows Torus and the below warning/error is in the console. I think you might need to call `onboard.walletReset()` before `onboard.walletSelect()`, thought I'm not certain of this offhand

image

thelostone-mc:
bumping this up once more @jfarid27  ! this does seem worth having in once @mds1 comments are addressed

apbendi:
Hey @jfarid27, we're going to close this for now, as it's gotten a little stale. If to rebase and address the feedback above, we'd be happy to merge! Thanks again for your interest in contributing :)

PR #84, opened 2021-08-04
Support multihop swaps

### LINKED ISSUE #71: Allow Arbitrary Input Tokens for Swap and Donate
Current implementation requires an existing pool for the pair. Determine how to do arbitrary "paths" between tokens for Uniswap v3. Implement this in current contracts, and/or alongside the bulk checkout functionality.

mds1:
Resolves #71

Answering questions from that issue:

> Does the uniswap JS package have built in support finding paths for v3?

Not that I've seen, but we should be able to pull the logic from the Uniswap interface, which I believe is [here](https://github.com/Uniswap/uniswap-interface/blob/392d78b9da71de8e2c08f269369fa40ce8bdfcec/src/hooks/Trades.ts#L95-L127). Optimal pathfinding is more of an app-layer issue, and the dgrants app doesn't support swaps in any way yet, so I have not implemented this as part of this PR

> Do we do the encodePacking off chain or on chain? I think off chain, even though the simple example in the docs does it on chain. If we did it on-chain, we'd have to somehow spread two arrays (tokens/fees) across the call to encodePacked. Not sure how to do that in Solidity.

Agreed we should do it off-chain, since that seems simpler. This PR adds helper methods in `contract/test/utils.ts` that the tests use to do the encoding. The frontend can duplicate/modify these as-needed, as they are currently very basic and always assume we use the medium (0.3%) fee tier for a given pool


contracts/contracts/GrantRoundManager.sol
```diff
@@ -4,23 +4,23 @@ pragma solidity ^0.8.5;
...
   using SafeERC20 for IERC20;
 
-  /// @notice Donation inputs and Uniswap V3 swap inputs: https://docs.uniswap.org/protocol/guides/swaps/single-swaps
+  /// @notice Donation inputs and Uniswap V3 swap inputs: https://docs.uniswap.org/protocol/guides/swaps/multihop-swaps
   struct Donation {
     uint96 grantId; // grant ID to which donation is being made
     GrantRound[] rounds; // rounds against which the donation should be counted
-    IERC20 tokenIn; // token in which the user made the donation
-    uint24 fee; // selected fee tier
+    bytes path; // swap path, or if user is providing donationToken, the address of the donationToken
```
thelostone-mc, line 20: would the address of the donationToken be set as bytes ?

contracts/test/utils.ts
```diff
@@ -50,9 +55,42 @@ export async function setNextBlockTimestamp(timestamp: BigNumberish): Promise<nu
   return timestamp;
 }
 
+// --- Uniswap SDK Helpers ---
+export async function encodeRoute(tokens: SupportedToken[]): Promise<string> {
```
thelostone-mc, line 59: @apbendi / @mds1 would it make sense to move this into it's utils.ts file which is not under `test` cause this would be invoked by dApp as well right ?

contracts/contracts/GrantRoundManager.sol
mds1: Yep, but ethers will handle the conversion for us behind the scenes, so all you need to do is specify the `path` as the token address as shown in L194 here https://github.com/dcgtc/dgrants/blob/45dfae3d110c8935b9549dac94e8883d75c17c91/contracts/test/GrantRoundManager.test.ts#L186-L196

contracts/contracts/libraries/BytesLib.sol
```diff
@@ -0,0 +1,498 @@
+// SPDX-License-Identifier: Unlicense
```
apbendi, line 1: Might want to do a quick sanity check with legal that there are no implications to including this unlicensed code. Defer to @phutchins on that question.

contracts/test/utils.ts
apbendi: It would definitely be nice to share such utilities across app and contract tests. Right now, we do share [types](https://github.com/dcgtc/dgrants/tree/main/types) between them, but I think that's it. I wonder how difficult it would be to pull these helpers into a shared method. Any thoughts here @wildmolasses.

contracts/contracts/libraries/BytesLib.sol
apbendi: Actually @mds1 it looks like there's also an npm package here, so licensing question aside, why check in this dependency? https://www.npmjs.com/package/solidity-bytes-utils

apbendi (approved):
Left a couple small comments but overall looks great. Wasn't too bad! I agree with your assessment that optimal route finding is a frontend concern that can be iterated on there.

Edit: Also looks like we have CI failure, but not seeing that locally, and we also have merge conflicts that need to be resolved.

contracts/contracts/GrantRoundManager.sol
```diff
@@ -136,30 +137,22 @@ contract GrantRoundManager {
...
+        _amountIn,
+        _donation.amountOutMinimum
       );
 
       // If user is sending a token, transfer it to this contract and approve the router to spend it
       if (msg.value == 0) {
-        _tokenIn.safeTransferFrom(msg.sender, address(this), _amountIn);
-        _tokenIn.approve(address(router), type(uint256).max); // TODO optimize so we don't call this every time
+        IERC20(_tokenIn).safeTransferFrom(msg.sender, address(this), _amountIn);
+        IERC20(_tokenIn).approve(address(router), type(uint256).max); // TODO optimize so we don't call this every time
```
thelostone-mc, line 151: Is this case needed ? we do this already on L109

https://github.com/dcgtc/dgrants/blob/08817d4e38fe3c85cbb2b92d193306ec025a0572/contracts/contracts/GrantRoundManager.sol#L109

contracts/test/GrantRoundManager.test.ts
```diff
@@ -210,25 +206,11 @@ describe('GrantRoundManager', () => {
...
-
-      // Give the user 100 DAI
-      const amountIn = parseUnits('100', 18);
-      await setBalance('dai', user.address, amountIn);
-
-      // Set payee address to be a random address
-      const payee = randomAddress();
-      await mockRegistry.mock.getGrantPayee.returns(payee);
-
+    it('input token DAI, output token GTC', async () => {
```
thelostone-mc, line 209:
```suggestion
    it('input token DAI, output token GTC with no direct LP', async () => {
```

contracts/test/utils.ts
```diff
@@ -50,9 +55,42 @@ export async function setNextBlockTimestamp(timestamp: BigNumberish): Promise<nu
...
+}
+
+async function getPoolInstance(token0: SupportedToken, token1: SupportedToken): Promise<Pool> {
+  // Get pool contract
+  token0 = token0 === 'eth' ? 'weth' : token0;
+  token1 = token1 === 'eth' ? 'weth' : token1;
+  const tokenA = new Token(1, tokens[token0].address, tokens[token0].decimals, tokens[token0].symbol, tokens[token0].name); // prettier-ignore
+  const tokenB = new Token(1, tokens[token1].address, tokens[token1].decimals, tokens[token1].symbol, tokens[token1].name); // prettier-ignore
+  const fee = FeeAmount.MEDIUM; // always use Medium fee for now
+  const poolAddress = computePoolAddress({ factoryAddress: UNISWAP_FACTORY, tokenA, tokenB, fee });
```
thelostone-mc, line 80: more of a question! 
looking at https://github.com/Uniswap/uniswap-v3-sdk/blob/cc9acb1d27cf6c7f02be167f8747e4d0e9c99c7c/src/utils/computePoolAddress.ts#L20 
the assumption here is that tokenA, tokenB LP exists ?  It doesn't account for a situation where it does not right ?

contracts/test/utils.ts
```diff
@@ -50,9 +55,42 @@ export async function setNextBlockTimestamp(timestamp: BigNumberish): Promise<nu
...
+// --- Uniswap SDK Helpers ---
+export async function encodeRoute(tokens: SupportedToken[]): Promise<string> {
+  const route = await getRoute(tokens);
+  return encodeRouteToPath(route, false);
+}
+
+async function getRoute(tokens: SupportedToken[]): Promise<Route<Token, Token>> {
+  const poolPromises: Promise<Pool>[] = [];
+  tokens.forEach((_, index) => {
+    if (index > 0) poolPromises.push(getPoolInstance(tokens[index - 1], tokens[index]));
```
thelostone-mc, line 67: would we want to make this cleaner for the sake of readability 

```suggestion
    const from_token = tokens[index];
    const to_token = tokens[index+1];
    poolPromises.push(
     getPoolInstance(from_token, to_token)
    );
```

contracts/test/utils.ts
mds1: I agree we want this eventually, but I don't think we have an issue/spec for integrating swaps into the app so it also may not be worth delaying this PR to share these with the frontend, especially since this version of the functions is too probably simple for the app

contracts/contracts/libraries/BytesLib.sol
mds1: Did not realize there was an npm package, thanks! Updated in beaf5ea

contracts/contracts/GrantRoundManager.sol
mds1: Good catch, at first I made some changes that required this, then removed them and forgot to get rid of this. Fixed in ed764701e4f064c93f2b6036e292895a3cf2f315

contracts/test/GrantRoundManager.test.ts
mds1: Clarified in ed764701e4f064c93f2b6036e292895a3cf2f315

contracts/test/utils.ts
mds1: Right, we don't verify the pool exists, and if the pool doesn't exist the swap will fail. This is ok for now but when integrated into the frontend we'd want to verify the pool exists

contracts/test/utils.ts
mds1: I know @apbendi would agree with you here 😛

So fixed in ed764701e4f064c93f2b6036e292895a3cf2f315

apbendi: yaaasss 🎉  😜

contracts/contracts/libraries/BytesLib.sol
mds1: Oops forgot to address the license issue. This is [The Unlicense](https://spdx.org/licenses/Unlicense.html) (which is very different from "Unlicensed"), which pretty much lets you do anything, so I don't think there should be any issues here

thelostone-mc (approved):

contracts/contracts/libraries/BytesLib.sol
apbendi: Terrible name, but I agree now that I understand what it is!

gdixon (approved):

PR #85, opened 2021-08-04
Add matching pool token

### LINKED ISSUE #63: Separate Donation Tokens Used In Round
Currently, a `GrantRound` has one `donationToken`, which is used for both the matching pool contributions and individual grant contributions. We want to separate these such that the entity who creates the `GrantRound` can specify a different token used for each case:

`donationToken` for individual contributions
`matchingToken` for the matching pool contributions

Both should be provided in the `GrantRound` constructor. The `GrantRoundManager` should continue to hold the `donationToken` fixed for all Grants deployed through it, but should allow the `matchingToken` to be passed as a parameter.

This PR is dependent on #61.

corydickson:
Closes #63

### DIFF
--- a/contracts/contracts/GrantRound.sol
+++ b/contracts/contracts/GrantRound.sol
@@ -72,7 +72,8 @@ contract GrantRound {
     string memory _metaPtr,
     uint256 _minContribution
   ) {
-    require(_donationToken.totalSupply() > 0, "GrantRound: Invalid token");
+    require(_donationToken.totalSupply() > 0, "GrantRound: Invalid donation token");
+    require(_matchingToken.totalSupply() > 0, "GrantRound: Invalid matching token");
     require(_startTime >= block.timestamp, "GrantRound: Start time has already passed");
     require(_endTime > _startTime, "GrantRound: End time must be after start time");
     require(_registry.grantCount() >= 0, "GrantRound: Invalid registry"); // verify this call doesn't revert
@@ -103,9 +104,9 @@ contract GrantRound {
    */
   function payoutGrants(address _payoutAddress) external afterRoundEnd {
     require(msg.sender == payoutAdmin, "GrantRound: Only the payout administrator can call this method");
-    uint256 balance = donationToken.balanceOf(address(this));
+    uint256 balance = matchingToken.balanceOf(address(this));
     hasPaidOut = true;
-    donationToken.safeTransfer(_payoutAddress, balance);
+    matchingToken.safeTransfer(_payoutAddress, balance);
   }
 
   /**
--- a/contracts/contracts/GrantRound.sol
+++ b/contracts/contracts/GrantRound.sol
@@ -55,7 +55,7 @@ contract GrantRound {
    * @param _payoutAdmin Grant round administrator that has permission to payout the matching pool
    * @param _registry Address that contains the grant metadata
    * @param _donationToken Address of the ERC20 token in which donations are made
-   * @param _donationToken Address of the ERC20 token for accepting matching pool contributions
+   * @param _matchingToken Address of the ERC20 token for accepting matching pool contributions
    * @param _startTime Unix timestamp of the start of the round
    * @param _endTime Unix timestamp of the end of the round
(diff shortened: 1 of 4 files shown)


contracts/contracts/GrantRoundManager.sol
```diff
@@ -74,17 +75,20 @@ contract GrantRoundManager {
...
     address _owner,
     address _payoutAdmin,
+    IERC20 _matchingToken,
     GrantRegistry _registry,
     uint256 _startTime,
     uint256 _endTime,
     string memory _metaPtr,
     uint256 _minContribution
   ) external {
+    require(_matchingToken.totalSupply() > 0, "GrantRoundManager: Invalid matching token");
```
corydickson, line 85: Adding similar check to the donation token supply check in this contracts constructor

contracts/contracts/GrantRound.sol
```diff
@@ -24,9 +24,12 @@ contract GrantRound {
   /// @notice GrantsRegistry
   GrantRegistry public immutable registry;
 
-  /// @notice ERC20 token that accepts pool donations
+  /// @notice ERC20 token that accepts donations
```
mds1, line 27: I think this comment for the `donationToken` could be a bit more clear. Maybe

This comment a bit unclear
```suggestion
  /// @notice Token used for all contributions. Contributions in a different token are swapped to this token
```

contracts/contracts/GrantRound.sol
```diff
@@ -24,9 +24,12 @@ contract GrantRound {
...
   IERC20 public immutable donationToken;
 
+  /// @notice ERC20 token that for the matching pool
```
mds1, line 30: Similar, could probably be more explicit here
```suggestion
  /// @notice Token used to payout match amounts at the end of a round
```

contracts/contracts/GrantRound.sol
```diff
@@ -52,6 +55,7 @@ contract GrantRound {
    * @param _payoutAdmin Grant round administrator that has permission to payout the matching pool
    * @param _registry Address that contains the grant metadata
    * @param _donationToken Address of the ERC20 token in which donations are made
+   * @param _donationToken Address of the ERC20 token for accepting matching pool contributions
```
mds1, line 58: Typo. Also would update the `_donationToken` and `_matchingToken` comments to match any changes made to their state variable comments above

```suggestion
   * @param _matchingToken Address of the ERC20 token for accepting matching pool contributions
```

contracts/contracts/GrantRound.sol
```diff
@@ -85,10 +92,10 @@ contract GrantRound {
 
   /**
    * @notice Before the round ends this method accepts matching pool funds
-   * @param _amount The amount of donation token that can be sent to the contract for the matching pool
+   * @param _amount The amount of matching token that can be sent to the contract for the matching pool
```
mds1, line 95: Nit
```suggestion
   * @param _amount The amount of matching token that will be sent to the contract for the matching pool
```

contracts/contracts/GrantRoundManager.sol
```diff
@@ -65,6 +65,7 @@ contract GrantRoundManager {
    * @notice Creates a new GrantRound
    * @param _owner Grant round owner that has permission to update the metadata pointer
    * @param _payoutAdmin Grant round administrator that has permission to payout the matching pool
+   * @param _matchingToken Address for the matching pool tokens
```
mds1, line 68: If we update the comment for this param above, just making a note to remember to do that here also

mds1 (changes requested):
Looks great overall! My only comments are just on clarifying some of the docstrings/comments in the code to be a bit more explicit

contracts/contracts/GrantRoundManager.sol
corydickson: Addressed [17133de ](https://github.com/dcgtc/dgrants/pull/85/commits/17133de9d645a163a0ebc57efa019b2745e10db3)

PR #86, opened 2021-08-04
IPFS metadata storage

### LINKED ISSUE #67: IPFS Grant Metadata Storage Proof-of-Concept
This issue is a spec for a proof of concept/prototype implementation of storing Metadata on IPFS via Fleek. It's meant to demonstrate the end-to-end functionality, without including all details, such as the full Grant Schema, the ability to edit, or viewing past edits of the metadata. All such functionality can be added in future tasks once this proof of concept is built out.

#### Create Grant Flow

- [x] Update the `/dgrants/new` form to remove the Metadata URL field, and replace it with two new fields: Name & Description
- [x] When the user submits the form via "Create Grant":
     - [x] Add the name & description to a JSON object with corresponding keys, encode this JSON as a file, and add it to IPFS via Fleek (more below)
     - [x] Publish the IPFS CID generated above as the metadata ptr property when creating the `GrantRound` via transaction
         * Question: should we publish only the CID, and let the frontend presume this is an IPFS hash and fetch it from whichever gateway it chooses, or should we publish a full URL, specifically the URL to this IPFS data via a particular IPFS gateway that we choose. In other words, if the content hash is `QmXnnyufdzAWL5CqZ2RnSNgPbvCc1ALT73s6epPrRnZ1Xy`, should the metaPtr be simply `"QmXnnyufdzAWL5CqZ2RnSNgPbvCc1ALT73s6epPrRnZ1Xy"` or should it be something like `"https://cloudflare-ipfs.com/ipfs/QmXnnyufdzAWL5CqZ2RnSNgPbvCc1ALT73s6epPrRnZ1Xy"`
         * Answer: see [here](https://github.com/dcgtc/dgrants/issues/67#issuecomment-891366857)

#### Grant Display

Update various places where Grant Metadata is displayed to actually load the data and show, specifically:

- [x] Update `/dgrants` to show the Grant Name & (possibly shortened) Description in each Grant Card; remove the Grant ID and mtaptr URL
- [x] Update `/dgrants/:id` to show Grant Name & Description, remove metadata URL and Grant ID

#### Sample JSON Object (partial schema for this initial implementation)

```json
{
    "name": "My Grant Name",
    "description": "This is my Grant. There are many like it, but this one is mine."
}
```

#### Fleek API

* Use the [ipfs-http-client](https://www.npmjs.com/package/ipfs-http-client) package
* Use the Fleek endpoint with the api key for authentication as such:
 
```javascript
import { create } from 'ipfs-http-client'

const ipfs = create({
  url: "https://ipfs-api.dev.fleek.cool",
  headers: {
    Authorization: "v2 <apiKey>", // or Bearer <JWT> or public <AppKey>
  },
});
```

* Use [ipfs.add](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md#ipfsadddata-options) to add and pin the data, and get back the new CID 
* Use [ipfs.get](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md#ipfsgetipfspath-options) to fetch the content hash

#### Add to grant bootstrap

- make sure that when grants are created locally that their content is properly linked

wildmolasses:
This PR resolves #67 

Important note: This PR uses a trick to accomplish the metaPtr resolution. Instead of resolving the metaPtr directly, it splits the URL, takes the CID, and resolves that with our fleek ipfs connection instance. Even if a metaPtr specified the cloudflare gateway, it would still use fleek to resolve. If a metaPtr did not have a CID in the last position of the URL, the fetch would completely fail.

The metadata fetching isn't robust. If the fetch fails, the metadata status will be 'error' and it will not reattempt a fetch.

There are no UI states yet corresponding to the metadata fetch status states of "pending", "resolved", "error"


apbendi:
Great start @wildmolasses! Before making any changes to the frontend, be sure to rebase to `main`

app/index.html
```diff
@@ -15,6 +15,16 @@
...
     <div id="app"></div>
+    <script>
+      window.global = window;
+    </script>
+    <script type="module">
+      import process from "process";
+      window.process = process;
+    </script>
+    <script>
+      let global = globalThis;
```
thelostone-mc, line 26: where is `globalThis` being populated from ?
would it make sense to combine L19 & L26 under the same script tag ?

app/src/shims.d.ts
```diff
@@ -15,5 +15,6 @@ interface ImportMeta {
   env: {
     VITE_BLOCKNATIVE_API_KEY: string;
     VITE_INFURA_ID: string;
+    VITE_FLEEK_API_KEY: string;
```
thelostone-mc, line 18: Worth renaming it to VITE_FLEEK_STORAGE_API_KEY ? (cause there is hosting key as well and it might confuse folks)

app/src/utils/ipfs.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+  }
+  throw Error('no content');
+};
+
+export const resolveMetaPtr = (url: string) => {
+  return ipfsFetch({
+    cid: url.split('/').reverse()[0],
+  });
+};
```
thelostone-mc, line 38: ^ would it be worth adding comments to the functions on

- what it does
- what are the arguments 

It would be useful if we decide to add in auto doc generation

app/src/config.ts
```diff
@@ -0,0 +1,10 @@
+export default {
+  ipfs: {
+    endpoint: 'https://ipfs-api.dev.fleek.cool',
+    headers: {
+      Authorization: `v2 ${import.meta.env.VITE_FLEEK_API_KEY}`, // or Bearer <JWT> or public <AppKey>
+    },
+    metaPtrEndpoint: 'https://ipfs-api.dev.fleek.cool',
+    // metaPtrEndpoint: 'https://cloudflare-ipfs.com/ipfs',
+  },
+};
```
thelostone-mc, line 10: ^ I'm assuming this would be a diff config for prod / dev. If yes would it make sense doing something like `dev: { ipfs : {...}}`

app/src/store/data.ts
```diff
@@ -159,6 +161,28 @@ export default function useDataStore() {
     lastBlockTimestamp.value = (timestamp as BigNumber).toNumber();
     grants.value = grantsList as Grant[];
     grantRounds.value = grantRoundsList as GrantRound[];
+    const metaPtrs = grants.value.map((grant) => grant.metaPtr);
+    const newMetadata = metaPtrs
+      .filter((metaPtr) => !grantMetadata.value[metaPtr]) // only want new metaPtrs
```
thelostone-mc, line 166: Just double checking -> it looks like a grant can have multiple metadata but we want only the which as status as pending ?
not super clear on what we are doing here

app/src/utils/utils.ts
```diff
@@ -28,6 +28,10 @@ export function isValidAddress(val: string | undefined) {
   return val && isAddress(val);
 }
 
+export function isDefined(val: unknown) {
+  return !!val;
+}
```
thelostone-mc, line 33: I'd recommend moving this to `utils` package

app/src/config.ts
thelostone-mc: or we could just define these in env files

thelostone-mc (commented):
left a couple of comments

app/index.html
mds1: `globalThis` exists on every web page: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis, and since browsers, unlike node.js, don't have a default `global` instance, we need to define it here since some npm packages rely on `global`

Also in favor of combing those script tags for brevity

app/src/utils/utils.ts
mds1: I think we should hold off on this for now per #114.

Though it also feels like this function is simple enough that it can be removed which might even be more readable (e.g. people know what `Boolean()` or `!!` does, but would have to look up what `isDefined` does

app/src/config.ts
mds1: Similarly, I think we can just delete this file and move the contents to `app/src/utils/ipfs.ts`, since that's the only place this is used. That way all IPFS config/abstraction lives in one place which is nice

app/src/store/data.ts
```diff
@@ -159,6 +161,28 @@ export default function useDataStore() {
...
+      Object.keys(newMetadata).map(async (url) => {
+        try {
+          const data = await resolveMetaPtr(url);
+          grantMetadata.value[url] = { status: 'resolved', ...data };
+        } catch (e) {
+          grantMetadata.value[url] = { status: 'error' };
+          console.error(e);
+        }
+      })
+    );
```
mds1, line 185: It feels like lines 174 and 175 should be switched unless I'm misunderstanding what's going on here?

Also one tip that we are using in this file: start by assigning things to intermediate variables such as `_grantMetadata`, then assign to `grantMetadata.value` at the end of the method once all promises have resolved. The reason for that is because the app is reactive to the state, so if you update state as individual promises resolve, the UI will be much jumpier/more frequent updates than if you do it all at once when all promises are resolved

app/src/utils/ipfs.ts
```diff
@@ -0,0 +1,38 @@
...
+});
+
+export const createGrant = async ({ name, description }: GrantMetadata) => {
+  const res = await ipfs.add(JSON.stringify({ name, description }));
+  return res.cid;
+};
+
+export const createMetaPtr = ({ cid }: { cid: string }) => {
+  return `${config.ipfs.metaPtrEndpoint}/${cid}`;
+};
```
mds1, line 17: Nit: maybe rename to `getMetaPtr` as this feels like a getter method that's not actually creating something new

app/src/utils/ipfs.ts
mds1: Agreed, more comments would be useful!

Re docgen, personally I have not been using NatSpec everywhere, so we'll need some comments cleanup before we do docgen, but you're right that is probably best practice going forward

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -102,6 +99,10 @@ function useGrantDetail() {
     if (!grants.value || Array.isArray(id)) return null; // array type unsupported
     return grants.value[Number(id)];
   });
+  const grantMetadata = computed(() => {
+    if (!grant.value) return null;
+    return metadata.value[grant.value.metaPtr];
+  });
```
mds1, line 105: Seems like a good fit for a nice one-liner! 
```suggestion
  const grantMetadata = computed(() => grant.value ? metadata.value[grant.value.metaPtr] : null);
```

app/src/views/GrantRegistryNewGrant.vue
```diff
@@ -75,22 +86,30 @@ function useNewGrant() {
   const { poll } = useDataStore();
 
   // Define form fields and parameters
-  const form = ref<{ owner: string; payee: string; metaPtr: string }>({
+  const form = ref<{ owner: string; payee: string; name: string; description: string }>({
     owner: '',
     payee: '',
-    metaPtr: '',
+    name: '',
+    description: '',
```
mds1, line 93: Should we define a local `Form` type, since I imagine we'll be adding more fields later too?

app/src/views/GrantRegistryNewGrant.vue
```diff
@@ -75,22 +86,30 @@ function useNewGrant() {
...
    * @notice Creates a new grant, parses logs for the Grant ID, and navigates to that grant's page
    */
   async function createGrant() {
     // Send transaction
-    const { owner, payee, metaPtr } = form.value;
+    const { owner, payee, name, description } = form.value;
     if (!signer.value) throw new Error('Please connect a wallet');
+    const metaPtr = await ipfs
+      .createGrant({ name, description })
+      .then((cid) => ipfs.createMetaPtr({ cid: cid.toString() }));
```
mds1, line 112: Just curious why you use a `.then` instead of
```suggestion
    const cid = await ipfs.createGrant({ name, description })
    const metaPtr = await ipfs.createMetaPtr({ cid: cid.toString() }));
```

apbendi (commented):
Overall looking and working great, @wildmolasses. Nice job getting this going.

My only comment is on the way we're processing the metaPtr URI. Instead of hacking off the CID from the end— and assuming it's there in the first place— I think we should just have a whitelisted set of endpoints from which the app will fetch content from. This should be separate from the concern of whether the endpoint is an IPFS gateway or not.

In other words, we should have a hardcoded list of valid endpoints, which can include just `https://ipfs-api.dev.fleek.cool` for now (or also `https://cloudflare-ipfs.com` if we want) and when the metaPtr has a url at one of those endpoints, we should fetch the contents. If it doesn't, it shouldn't fetch, and should probably have an error message like "Unsupported source for grant metadata" or something like that.

Curious to hear what others think of this approach cc @thelostone-mc @mds1 @gdixon

app/src/views/GrantRegistryNewGrant.vue
wildmolasses: my general rule is that if something is used once there's no reason to declare it, i.e. since we don't use `cid` elsewhere there's no reason to declare it. i break this sometimes for readability's sake, but `then` clauses are pretty readable to me. Thoughts?

app/src/utils/utils.ts
wildmolasses: > Overall looking and working great, @wildmolasses. Nice job getting this going.
> 
> My only comment is on the way we're processing the metaPtr URI. Instead of hacking off the CID from the end— and assuming it's there in the first place— I think we should just have a whitelisted set of endpoints from which the app will fetch content from. This should be separate from the concern of whether the endpoint is an IPFS gateway or not.
> 
> In other words, we should have a hardcoded list of valid endpoints, which can include just `https://ipfs-api.dev.fleek.cool` for now (or also `https://cloudflare-ipfs.com` if we want) and when the metaPtr has a url at one of those endpoints, we should fetch the contents. If it doesn't, it shouldn't fetch, and should probably have an error message like "Unsupported source for grant metadata" or something like that.
> 
> Curious to hear what others think of this approach cc @thelostone-mc @mds1 @gdixon

In principle I know that's correct, however different endpoints have different auth models -- well, really, fleek is the one where we need an api key in order to fetch. So if we're going to use the fleek endpoint as our metaPtr, do we want to write conditional logic where *if* the metaPtr is fleek (which is currently the case for any grant created by this app) then use an api key, and otherwise proceed without auth token? it starts to feel a little weird. I thought we could defer on this decision for this POC PR, but maybe now is the time to address this.

wildmolasses:
> Overall looking and working great, @wildmolasses. Nice job getting this going.
> 
> My only comment is on the way we're processing the metaPtr URI. Instead of hacking off the CID from the end— and assuming it's there in the first place— I think we should just have a whitelisted set of endpoints from which the app will fetch content from. This should be separate from the concern of whether the endpoint is an IPFS gateway or not.
> 
> In other words, we should have a hardcoded list of valid endpoints, which can include just `https://ipfs-api.dev.fleek.cool` for now (or also `https://cloudflare-ipfs.com` if we want) and when the metaPtr has a url at one of those endpoints, we should fetch the contents. If it doesn't, it shouldn't fetch, and should probably have an error message like "Unsupported source for grant metadata" or something like that.
> 
> Curious to hear what others think of this approach cc @thelostone-mc @mds1 @gdixon

In principle I know that's correct, however different endpoints have different auth models -- well, really, fleek is the one where we need an api key in order to fetch. So if we're going to use the fleek endpoint as our metaPtr, do we want to write conditional logic where *if* the metaPtr is fleek (which is currently the case for any grant created by this app) then use an api key, and otherwise proceed without auth token? it starts to feel a little weird. I thought we could defer on this decision for this POC PR, but maybe now is the time to address this.

app/src/views/GrantRegistryNewGrant.vue
wildmolasses: If it's only used once I don't think it's necessary, but would defer to you guys

app/src/views/GrantRegistryGrantDetail.vue
wildmolasses: good call!

apbendi:
> In principle I know that's correct, however different endpoints have different auth models -- well, really, fleek is the one where we need an api key in order to fetch. So if we're going to use the fleek endpoint as our metaPtr, do we want to write conditional logic where if the metaPtr is fleek (which is currently the case for any grant created by this app) then use an api key, and otherwise proceed without auth token? it starts to feel a little weird, but maybe this is acceptable.

Per our discussion offline, I think the right approach here is to use keep metaptr fetching dumb, so an http endpoint that requires an API key should not be used. For the immediate term, we can use a whitelisted cloudflare (or another open IPFS gateway) as the metaPtr URL, therefore allowing us to fetch without any authentication.

app/src/store/data.ts
mds1: Marking this comment resolved, since @wildmolasses explained the reasoning on a call to me: TL;DR is some URLs may resolve slowly or not at all, so it's better to design the UI to not be jumpy and handle this well, as opposed to waiting for everything to resolve at once which may be slow

app/src/config.ts
wildmolasses: followed @mds1 suggestion

app/src/store/data.ts
wildmolasses: i'll try to add some comments!

app/src/utils/utils.ts
wildmolasses: i'm using it for form validation; would `Boolean()` work just as well? I like how "isDefined" makes things a little more explicit but could defer to you guys!

thelostone-mc (approved):

PR #89, opened 2021-08-05
Upgrades hardhat version (2.2.0 -> 2.5.0)

gdixon:
We need to upgrade hardhat to the latest version to account for changes in EIP-1559

This is the error I was seeing when starting the dev server:

Screenshot 2021-08-05 at 18 28 50


apbendi (approved):
lgtm

PR #91, opened 2021-08-09
feat: dcurve

thelostone-mc:
# `@dgrants/dcurve`

This package allows you to create a distribution for a specific GrantRound based on the contributions recieved by the Grants.

This package is intended to be used

- by the dApp hosting dGrants to
  - generate the CLR distribution at the end of the round
  - how the prediction would vary for a grant it it were to receive an additional contribution of value X
- as a standalone dApp to allow members of the DAO to verify the distribution/merkle root themselves to ensure that the results are correct

## Structure

```
.
├── src
│   ├── README.md               # getting started guide
│   ├── types.ts                # typescript types
|   ├── index.ts                # root of the project, exporting public interface
│   ├── internal
│       ├── calc                # folder containing implementations of CLR algorithms
│       ├── merkle.ts           # file containing implementations to generate merkle-root and claims
|       ├── fetch.ts            # fetch information from chain
|       ├── clr.ts              # orchestrator
└── ...
```

## Examples/Testing

The distribution calculation has been (inefficiently) added to the datastore (`app/src/store/data.ts`) so that we can test the functionality ahead of getting this merged - please visit any page and check the console.


frankchen07:
@thelostone-mc - in `linear.ts` starting on line 68:

```
    if (grantMatch.match >= matchCap) {
      // normalize match if grant has match greater than cap
      grantMatch.match = matchCap;
    }
  });
  ```
  
What's the use for match cap here? I think that was from downtown stimulus - we're not artificially limiting the cap, but we are normalizing when we hit saturation.

**Comment by frankchen07** [comment]
The rest of the CLR logic LGTM. 👍

**Comment by gdixon** [comment]
All changes are in!

Everything is working correctly, the `fetch` method may benefit from another set of eyes (@mds1?), in the future we will probably replace the rpc calls with queries to the graph, but for now I was wondering if we could better filter the `GrantDonation`'s to pull down only a specific rounds events?

If we're happy to optimise later, then this PR is good for review 🙌  @apbendi @phutchins @mds1 @frankchen07 @wildmolasses @jferas - any feedback, please let us know and we'll get it sorted asap! 👍 

Thanks for all your help here @thelostone-mc!!

File: dcurve/src/internal/clr.ts [hunk 1]
```diff
@@ -0,0 +1,151 @@
...
+      // record the distribution used to generate the merkle
+      distribution.payoutDistribution = Object.values(payoutObj) as PayoutMatch[];
+      // prevent attempting to generate a merkle tree if we're missing the minimum number of leafs
+      if (distribution.payoutDistribution.length > 1) {
+        // generate the merkle tree and record the root
+        distribution.merkle = generateMerkle(distribution.payoutDistribution, grantRoundContributions.currDecimals);
+        distribution.hash = getMerkleRoot(distribution.merkle);
+      } else {
+        // empty state when we can't create a tree
+        distribution.merkleError = 'Missing required context to build tree';
```
mds1, line 81: Why empty state instead of throwing an error? E.g. if I'm a user do I need to manually check that `merkleError` is undefined to safely use this script? Not saying this is bad, just trying to understand the decision here!

dcurve/src/internal/fetch.ts
```diff
@@ -0,0 +1,73 @@
...
+ */
+export const fetch = async (args: GrantRoundFetchArgs) => {
+  // fetch grantRoundManager
+  const roundManager = new Contract(args.grantRoundManager, GRANT_ROUND_MANAGER_ABI, args.provider);
+  // fetch the grantList
+  const registry = new Contract(args.grantRegistry, GRANT_REGISTRY_ABI, args.provider);
+
+  // set-up
+  const values = await Promise.all([
+    registry.getAllGrants(),
```
mds1, line 22: Just a note this may fail if the number of grants is too large to return in a single call, in which case this will need to be chunked with `getGrants()`. Doesn't matter for now, but something to be aware as a potential future source of error

dcurve/src/internal/fetch.ts
```diff
@@ -0,0 +1,73 @@
...
+    roundManager.donationToken(),
+    roundManager.queryFilter(roundManager.filters.GrantDonation()),
```
mds1, line 24: Similarly, Infura and Alchemy nodes will throw if there are more than 10,000 `GrantDonation` logs returned by a query. Ways to prevent this:
- Provide a start block and end block
- Use your own node with no limits
- Add a method to repeatedly attempt with smaller chunks on failure (see how Rotki does this [here](https://github.com/rotki/rotki/pull/1795/files))

I don't think this change needs to be done in this PR, but maybe worth documenting some of these comments in the README or issues

dcurve/src/internal/fetch.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @param {GrantRoundFetchArgs}
+ */
+export const fetch = async (args: GrantRoundFetchArgs) => {
+  // fetch grantRoundManager
+  const roundManager = new Contract(args.grantRoundManager, GRANT_ROUND_MANAGER_ABI, args.provider);
+  // fetch the grantList
+  const registry = new Contract(args.grantRegistry, GRANT_REGISTRY_ABI, args.provider);
+
+  // set-up
+  const values = await Promise.all([
```
mds1, line 21: Nit, but I prefer something like this for readability/brevity, plus then you can get rid of definitions later too
```suggestion
  const [allGrants, donationToken, grantDonations] = await Promise.all([
```

dcurve/src/internal/fetch.ts
```diff
@@ -0,0 +1,73 @@
...
+  grantDonations.forEach(async (contribution) => {
+    // get tx details to pull contributor details from
+    const tx = await contribution.getTransaction();
+
+    // check that the contribution is valid
+    const inRound = contribution?.args?.rounds.indexOf(args.grantRound) !== -1;
+    const isIgnoredGrant =
+      args.ignore && args.ignore.grants && args.ignore.grants?.indexOf(contribution?.args?.grantId.toNumber()) !== -1;
+    const isIgnoredContributor =
+      args.ignore && args.ignore.contributionAddress && args.ignore.contributionAddress?.indexOf(tx.from) !== -1;
```
mds1, line 52: Thoughts on `.includes()` instead of `.indexOf()`? The former is a bit more readable IMO.

Also a nit, but the first uses `?` and the second and third handle the checks separately, e.g. `args.ignore &&` instead of `args?.ignore?....`. Guessing that's from a TS complaint?

dcurve/src/internal/fetch.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    // only include transactions from this grantRound which havent been ignored
+    if (inRound && !isIgnoredGrant && !isIgnoredContributor) {
+      contributions.push({
+        grantId: contribution?.args?.grantId.toNumber(),
+        amount: contribution?.args?.donationAmount / 10 ** args.supportedTokens[donationToken].decimals,
```
mds1, line 58: Current approach may result in math errors for donation amounts above `Number.MAX_SAFE_INTEGER` (which will be most donation amounts), so we should use ethers' `formatUnits` here instead to go from BigNumber to human number

```suggestion
        amount: formatUnits(contribution?.args?.donationAmount, args.supportedTokens[donationToken].decimals)
```

dcurve/src/internal/fetch.ts
```diff
@@ -0,0 +1,73 @@
...
+        grantAddress: grantsList[contribution?.args?.grantId.toNumber()],
```
mds1, line 59: We have grant ID a few lines up. This one is called `grantAddress` but also returns the grant ID. Guessing the value here should be the grant's `payee` address and not their grant ID?
```suggestion
        grantAddress: grantsList[contribution?.args?.grantId.toNumber()],
```

dcurve/src/internal/fetch.ts
```diff
@@ -0,0 +1,73 @@
...
+    const inRound = contribution?.args?.rounds.indexOf(args.grantRound) !== -1;
+    const isIgnoredGrant =
+      args.ignore && args.ignore.grants && args.ignore.grants?.indexOf(contribution?.args?.grantId.toNumber()) !== -1;
+    const isIgnoredContributor =
+      args.ignore && args.ignore.contributionAddress && args.ignore.contributionAddress?.indexOf(tx.from) !== -1;
+
+    // only include transactions from this grantRound which havent been ignored
+    if (inRound && !isIgnoredGrant && !isIgnoredContributor) {
+      contributions.push({
+        grantId: contribution?.args?.grantId.toNumber(),
```
mds1, line 57: In the cart we handle grantId as a string because it's saved in localStorage which only deals with strings. I'm fine with either though, but we should just agree and be consistent to avoid issues. Let me know your preference!

dcurve/src/internal/fetch.ts
```diff
@@ -0,0 +1,73 @@
...
+        address: tx.from,
+      });
+    }
+  });
+
+  // return contributions as GrantRoundContributions
+  return {
+    grantRound: args.grantRound,
+    totalPot:
+      (await donationTokenContract.balanceOf(args.grantRound)) / 10 ** args.supportedTokens[donationToken].decimals,
```
mds1, line 69: Same comment about `formatUnits` instead of this math

dcurve/src/internal/fetch.ts
```diff
@@ -0,0 +1,73 @@
...
+    currDecimals: 18,
```
mds1, line 70: What's this? Unclear why it's hardcoded to 18

dcurve/src/internal/merkle.ts
```diff
@@ -0,0 +1,98 @@
+import { parseUnits } from 'ethers/lib/utils';
+import {
+  MerkleDistributorInfo,
+  NewFormat,
+  parseBalanceMap,
+} from '@dgrants/utils/src/merkle-distributor/parse-balance-map';
+import { PayoutMatch } from '../types';
+
+// export MerkleDistributorInfo
```
mds1, line 9: Delete comment?

dcurve/src/internal/fetch.ts
mds1: Oh I see this used in the Merkle tree also, so it seems the idea is:
1. Get human readable amounts from this method
2. Pass them into the Merkle tree, along with `currDecimals`, to BigNumber-ify them

Hmm, maybe this method should take an optional payout token address and return it's decimal value here if provided and null or undefined otherwise? And maybe `matchTokenDecimals` would be more descriptive than `currDecimals`?

dcurve/src/internal/merkle.ts
```diff
@@ -0,0 +1,98 @@
...
+ */
+export const generateMerkle = (distribution: PayoutMatch[], currDecimals: number) => {
+  // collect the input
+  const merkleInput: NewFormat[] = [];
+
+  // foreach payout parse the earnings
+  distribution.forEach((payoutMatch: PayoutMatch) => {
+    merkleInput.push({
+      address: payoutMatch.address,
+      earnings: parseUnits(payoutMatch.match.toString(), currDecimals).toHexString(),
```
mds1, line 27: We should clearly document (in the README?) that the input values in `payoutMatch` are human-readable values, and not parsed based on decimals, since this deviates from the typical interface

dcurve/src/internal/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * util function which adds anon contribution of given value to the
+ * grantRoundContributions for given grant
+ *
+ * @param grantId
+ * @param grantRoundContributions
+ * @param amount
+ * @returns grantRoundContributions
+ */
+export const addAnonContribution = (
```
mds1, line 12: I don't fully understand the purpose of this method, but not a big deal since it seems this is a detail of the algorithm implementation

mds1 (dismissed):
This is awesomeeeeeee great job! Love all the code comments too, really helped me understand what's going on 😍

Personally I like avoiding reliance on the Graph for as long as possible because it does add another dependency / point of failure. Added some comments about the query options, and one other thing is we can cache events in IndexedDB if we want (caching in localStorage will fill it up too fast)

dcurve/package.json
```diff
@@ -0,0 +1,37 @@
...
+  ],
+  "main": "src/index.ts",
+  "files": [
+    "src"
+  ],
+  "scripts": {
+    "clean": "rimraf dist",
+    "dev": "echo 'TODO'",
+    "build": "echo 'TODO'",
+    "test": "echo 'TODO'",
```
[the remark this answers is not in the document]
thelostone-mc: yup ! we'd want to add this in soon

dcurve/src/internal/utils.ts
thelostone-mc: Updating the description and renaming the method to make it cleaner

dcurve/src/internal/fetch.ts
thelostone-mc: Left a todo! will create an issue for this as well later

dcurve/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+### calc/\*.ts
+
+This folder lists out the different QF algorithms supported by dcurve
+and would need to be set when creating a new instance of CLR object
+
+| QF algorithm | How to import                               | Description                                  |
+| ------------ | ------------------------------------------- | -------------------------------------------- |
+| linear       | `import { linear } from @dgrants/dcurve;`   | This is QF without pairwise                  |
+| pairwise     | `import { pairwise } from @dgrants/dcurve;` | This is QF which uses pairwise (coming soon) |
```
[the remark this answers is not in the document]
thelostone-mc: @frankchen07 would you be able to add some comments here

dcurve/src/internal/fetch.ts
thelostone-mc: Adding todo

dcurve/src/internal/fetch.ts
thelostone-mc: cleaned it up a bit + renamed the variable so it makes sense

app/vite.config.ts
```diff
@@ -21,4 +21,8 @@ export default defineConfig({
   optimizeDeps: {
     exclude: ['electron'],
   },
+  define: {
+    'process.env': process.env,
+    global: 'window',
```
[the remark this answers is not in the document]
gdixon: Yes, looks like #86 is solving for the same case 🙌 we can drop it from here 👍

dcurve/src/internal/clr.ts
```diff
@@ -0,0 +1,151 @@
...
+  InitArgs,
+  PayoutMatch,
+  PayoutMatches,
+} from '../../src/types';
+import { generateMerkle, getMerkleRoot } from './merkle';
+import { addAnonContribution, getGrantMatch } from './utils';
+
+// polyfill Buffer for use in browser
+import { Buffer } from 'buffer/';
+// set to window (TODO: make this work in vite.config.ts)
```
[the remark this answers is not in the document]
gdixon: Yes, I'm not sure how best to polyfill this? Is `vite.config` the right place to be doing it?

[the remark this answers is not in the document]
mds1: In an ideal world we'd remove the dependency on Buffer, and if we can't do that, yea either `vite.config.ts`, `index.html`, or `main.ts` would be the place to polyfill it

Previously, the app accidentally introduced a dependency on Buffer when regenerating the lockfile, and I could not figure out how to polyfill it ([1](), [2](https://github.com/blocknative/onboard/issues/650)), so I just downgraded onboard.js to the version that doesn't require it.

dcurve/src/internal/clr.ts
```diff
@@ -0,0 +1,151 @@
...
+export class CLR {
+  _options: InitArgs;
+
+  constructor(options: InitArgs) {
+    this._options = options;
+  }
+
+  calculate(grantRoundContributions: GrantRoundContributions, _options?: Record<string, unknown>): GrantsDistribution {
+    // allow the options to be overridden
+    const options = Object.assign(this._options, _options || {});
```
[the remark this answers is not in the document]
gdixon: This was done this way to allow the calc implementations to have an extended option set which could be defined both globally and on a per calc basis. 

This was maybe more important when we intended to allow the hash mechanism to be defined as an option, but I still think there may be extra options we'd like to pass in to some calculations and not every calculation (for example the pairwise algorithm accepts a `threshold` which is specific to the round, this way we don't need a new instance to calculate each distribution, we can just feed the `threshold` on `calculate`/`predict`).

You are right though, creating a new instance would suffice! I'm happy to go either way here - what would your preference be? @mds1 @thelostone-mc

dcurve/src/internal/clr.ts
gdixon: `generateMerkle` will throw if its not provided enough data, I wrapped it because we can know ahead of time if the merkle-tree will fail - but I'm happy to back out those checks and let it throw?

dcurve/src/internal/fetch.ts
gdixon: We've made of these changes 🙌 - I always thought includes was slower, but I'm happy to be wrong - it does read better 🙌

dcurve/src/internal/fetch.ts
gdixon: I've converted all of the `grantId`'s to strings here too - yeah I agree, its better if everything is consistent 🙌

dcurve/src/internal/clr.ts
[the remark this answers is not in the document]
mds1: Got, thanks for the info. I'm fine with whatever you both think is best, since you're more familiar with the match algorithm/potential usages than I am!

dcurve/src/internal/fetch.ts
gdixon: > What's this? Unclear why it's hardcoded to 18

That should have been `args.supportedTokens[matchingToken].decimals` my bad!!

---

I think we will always want to avoid bigNumber for the clr calculation and only bring it back to generate the merkle - I think what you are suggesting here is similar to how its working, or am I missing something?

mds1: > I think we will always want to avoid bigNumber for the clr calculation and only bring it back to generate the merkle - I think what you are suggesting here is similar to how its working, or am I missing something?

Agreed, yea l I think we are on the same page here! Sorry I may have been unclear/misunderstood previously

dcurve/README.md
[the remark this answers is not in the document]
frankchen07: sure - we can add this into the readme:

The non-pairwise, original quadratic matching formula makes the inherent assumption that all agents participating in a grant round are fully uncoordinated. Therefore, we follow formula where the amount received by a grant is proportional to the square of the sum of the square roots of contributions received less the sum of the total contributions for the grant.

The pairwise quadratic matching formula takes into account that some agents are coordinating while contributing to grants and that this coordination should be penalized. 

This can be represented mathematically as a coordination coefficient, 1 if the agents are fully uncoordinated and 0 (no matching should occur) if the agents are fully coordinated. This is theoretical of course, and the applicable concept is that the more coordinated a pair of contributors are, the higher the penalty. In order to discover the level of coordination between users for each grant, this means we must take a permutation of all contributors who contributed to a grant and look across how many grants they both contributed to, to assess how much penalty a user pair would incur. 

The trade off is the calculation speed of the formula and the inherent collusion resistance. Non-pairwise gives us faster calculation speeds and the possibility of accurately estimating CLR match amounts. Pairwise, to some degree, protects against collusion tactics, but the complexity and overhead of running user to user permutations puts limitations on predicting match amounts and quickly running calculations.

dcurve/src/internal/clr.ts
corydickson: I'm inclined to let it throw here unless we imagine some DAO UI interface to consume this.

app/src/store/data.ts
```diff
@@ -156,11 +159,33 @@ export default function useDataStore() {
...
 
+    // test/example
+    const clr = new CLR({
+      calcAlgo: linear,
+      includePayouts: true,
+    } as InitArgs);
+    // get contributions and grantRound details
+    const grantRoundArgs = await fetch({
+      provider: provider.value,
+      grantRound: '0x8B4091997E3EbB87be90cEd3e50d8Bb27e1DC742',
```
corydickson, line 170: kind of a nit, just wondering why it's not a constant here.

gdixon: This was just to demo/test that things are working, we will likely want to hook the `prediction` call in rather than `calculate` and we can probably avoid calling `fetch` from the `app` itself too. 

We can remove these calls prior to merging 🙌

dcurve/README.md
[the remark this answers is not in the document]
thelostone-mc: sweet thanks for this @frankchen07

dcurve/package.json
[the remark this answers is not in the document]
mds1: @thelostone-mc Is there an issue for this? Just looked but might have missed it

dcurve/src/internal/clr.ts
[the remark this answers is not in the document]
mds1: Being tracked in https://github.com/dcgtc/dgrants/issues/127

dcurve/src/internal/clr.ts
```diff
@@ -0,0 +1,151 @@
...
+import { Buffer } from 'buffer/';
+// set to window (TODO: make this work in vite.config.ts)
+Object.defineProperty(window, 'Buffer', { value: Buffer });
+
+export class CLR {
+  _options: InitArgs;
+
+  constructor(options: InitArgs) {
+    this._options = options;
+  }
```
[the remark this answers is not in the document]
mds1: @gdixon @thelostone-mc Looks like this one wasn't done correctly. The current approach will create a property called `options` which is readonly, and a property called `_options` which is not readonly. It seems we want the below change to make the class only have a readonly `_options` property

```typescript
 _options: InitArgs; // REMOVE THIS LINE

  constructor(readonly options: InitArgs) { // ADD UNDERSCORE BEFORE options
    this._options = options; // REMOVE THIS LINE
  }
```

So the end result would be

```typescript
 constructor(readonly _options: InitArgs) {}
```

which gives a readonly `_options` property

dcurve/tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
```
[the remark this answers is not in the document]
mds1: Looks like this still needs to be run. Perhaps someone doesn't have the git hooks running?

dcurve/src/internal/clr.ts
mds1: And decision here? I'm fine with whatever you all decide! @gdixon @thelostone-mc @corydickson

dcurve/src/internal/fetch.ts
mds1: Being tracked in https://github.com/dcgtc/dgrants/issues/125

dcurve/src/internal/fetch.ts
mds1: Being tracked in https://github.com/dcgtc/dgrants/issues/126

dcurve/src/internal/fetch.ts
mds1: I wasn't sure about speed actually so did a quick search. Seems performance in nearly identical, and the only difference is how NaN is treated
- https://ryanpeden.com/javascript-indexof-vs-includes-performance/
- https://stackoverflow.com/questions/35370222/array-prototype-includes-vs-array-prototype-indexof

dcurve/src/internal/fetch.ts
```diff
@@ -0,0 +1,78 @@
...
+    const isIgnoredGrant = args?.ignore?.grants?.includes(grantId);
+    const isIgnoredContributor = args?.ignore?.contributionAddress?.indexOf(tx.from) !== -1;
+
+    // only include transactions from this grantRound which havent been ignored
+    if (inRound && !isIgnoredGrant && !isIgnoredContributor) {
+      contributions.push({
+        grantId: grantId,
+        amount: parseFloat(
+          formatUnits(contribution?.args?.donationAmount, args.supportedTokens[donationToken].decimals)
+        ),
```
mds1, line 62: Can we use `formatUnits(...).toNumber()` here instead? Only because `toNumber()` should be more robust since it's part of the class instance and will throw if it needs to, whereas `parseFloat` is more likely result in unexpected behavior/NaNs for some edge cases

dcurve/src/internal/fetch.ts
```diff
@@ -0,0 +1,78 @@
...
+        grantAddress: grantsDict[grantId],
+        address: tx.from,
+      });
+    }
+  });
+
+  // return contributions as GrantRoundContributions
+  return {
+    grantRound: args.grantRound,
+    totalPot: parseFloat(
```
mds1, line 72: Same comment on `parseFloat`

dcurve/src/internal/merkle.ts
mds1: Was this added somewhere

dcurve/src/internal/fetch.ts
thelostone-mc: @mds1 `formatUnits` returns a `string` ! We could do a `Number(...)`

mds1: Ah right I was thinking it returned a BigNumber. Hmm, I think `BigNumber.from(stuff).toNumber()` would be best at helping us catch any sneaky bugs/bad values that made it this far

phutchins (approved):
All looks good!

dcurve/src/internal/fetch.ts
gdixon: `BigNumber` will only accept integers and `formatUnits` will always return a decimal string with at least one decimal - the biggest number we can safely parse here would be `9007199254740991` - if we think we'll have to deal with any numbers larger than that then we may have to refactor to use `BigNumber` throughout - what do you think @mds1 @thelostone-mc?

mds1: Ahhh yes yes ok, we can stick with parseFloat for now, sorry! Since we're dealing with regular numbers we likely won't exceed that `2^53-1` limit so it's probably fine! I was just trying to think of edge cases/be cautious

dcurve/src/internal/merkle.ts
gdixon: I've added this to the readme now 👍 (in 1016d54)

dcurve/src/internal/fetch.ts
gdixon: Yeah, its best to be cautious - lets see how we go - we know that in this first round we'll be in control of the matching currency and the donation currencies, and I don't think we expect to have a match or a contribution that exceeds 9 quadrillion - so we should be safe!

dcurve/package.json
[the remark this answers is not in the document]
gdixon: @mds1 - I've just opened #137 and added some questions there to start the conversation - it might be better to track this in a discussion until we iron out the details though?

dcurve/tsconfig.json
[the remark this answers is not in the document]
gdixon: I ran prettier manually just to make sure and I didn't get any formatting changes, is this looking right to you now?

dcurve/src/internal/clr.ts
gdixon: We still need to clarify how we'll consume/expose the merkle/payout distribution, so I'm tempted to leave this in and make that decision in a follow-up PR? @apbendi any thoughts on this?

PR #96, opened 2021-08-11
Bulk donation support

### LINKED ISSUE #76: Support Bulk Grant Contributions in a Single Transaction
The current `swapAndDonate` method supports:

* Donating an arbitrary token
* To a single Grant
* In multiple Rounds

We want to update this to support:

* Donating an arbitrary token
* To *multiple* Grants
* In *multiple* denominations (e.g. 1 Grant receives 5 Dai and another 10 Dai)
* In multiple Rounds

To do this efficiently, it probably makes sense to first execute a single swap to `donationToken`, then transfer `donationToken`s  to N Grant payee addresses. The assumption here is that 1 swap and N transfers is going to be less expensive than N swaps.

The caller should provide a proportion for each grant receiving a donation. The proportions should be applied to the total `donationToken` received after the swap.

mds1:
Closes #76

The `GrantRoundManager` contract, and in particular the `donate` method, feel a bit clunky now, but I'm not totally sure how to clean it up. There's a few things that we could handle in separate issues to make it cleaner, so curious to hear thoughts around this:
- Making the `payee` address the grant ID (#54) removes a validation check
- Add an `assertValidContribution(address token)` method to the `GrantRound` can be used to combine the `isActive` and `donationToken` checks into a single call
- Making the `GrantRoundManager` inherit from `SwapRouter` (#74) would remove approvals and transfers that are currently needed
- Other ideas?

Current code coverage results are below

image

### DIFF
--- a/contracts/.solhint.json
+++ b/contracts/.solhint.json
@@ -2,7 +2,7 @@
   "extends": "solhint:recommended",
   "plugins": ["prettier"],
   "rules": {
-    "code-complexity": ["error", 7],
+    "code-complexity": ["error", 9],
     "compiler-version": ["error", "^0.8.0"],
     "const-name-snakecase": "off",
     "constructor-syntax": "error",
@@ -10,6 +10,7 @@
     "max-line-length": "off",
     "not-rely-on-time": "off",
     "prettier/prettier": ["error", { "endOfLine": "auto" }],
-    "reason-string": "off"
+    "reason-string": "off",
+    "reentrancy": "off"
   }
 }
--- a/contracts/contracts/GrantRoundManager.sol
+++ b/contracts/contracts/GrantRoundManager.sol
@@ -9,20 +9,12 @@ import "./GrantRegistry.sol";
 import "./GrantRound.sol";
 
 contract GrantRoundManager {
+  // --- Libraries ---
   using Address for address;
   using BytesLib for bytes;
   using SafeERC20 for IERC20;
 
-  /// @notice Donation inputs and Uniswap V3 swap inputs: https://docs.uniswap.org/protocol/guides/swaps/multihop-swaps
-  struct Donation {
-    uint96 grantId; // grant ID to which donation is being made
-    GrantRound[] rounds; // rounds against which the donation should be counted
-    bytes path; // swap path, or if user is providing donationToken, the address of the donationToken
-    uint256 deadline; // unix timestamp after which a swap will revert, i.e. swap must be executed before this
-    uint256 amountIn; // amount donated by the user
-    uint256 amountOutMinimum; // minimum amount to be returned after swap
-  }
-
+  // --- Data ---
   /// @notice Address of the GrantRegistry
   GrantRegistry public immutable registry;
 
@@ -32,21 +24,40 @@ contract GrantRoundManager {
   /// @notice Address of the ERC20 token in which donations are made
   IERC20 public immutable donationToken;
 
+  /// @notice Used during donations to temporarily store swap output amounts
+  mapping(IERC20 => uint256) internal swapOutputs;
+
   /// @notice WETH address
   address public constant WETH = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
 
+  /// @notice Scale factor
+  uint256 internal constant WAD = 1e18;
+
+  /// @notice Defines the total `amount` of the specified `token` that needs to be swapped to `donationToken`. If
+  /// `path == donationToken`, no swap is required and we just transfer the tokens
+  struct SwapSummary {
+    address token;
+    uint256 amount;
+    uint256 amountOutMinimum; // minimum amount to be returned after swap
+    bytes path;
+  }
+
+  /// @notice Donation inputs and Uniswap V3 swap inputs: https://docs.uniswap.org/protocol/guides/swaps/multihop-swaps
+  struct Donation {
+    uint96 grantId; // grant ID to which donation is being made
+    IERC20 token; // address of the token to donate
+    uint256 ratio; // ratio of `token` to donate, specified as numerator where WAD = 1e18 = 100%
+    GrantRound[] rounds; // rounds against which the donation should be counted
+  }
+
+  // --- Events ---
   /// @notice Emitted when a new GrantRound contract is created
   event GrantRoundCreated(address grantRound);
 
   /// @notice Emitted when a donation has been made
-  event GrantDonation(
-    uint96 indexed grantId,
-    IERC20 indexed tokenIn,
-    uint256 amountIn,
-    uint256 amountOut,
-    GrantRound[] rounds
-  );
+  event GrantDonation(uint96 indexed grantId, IERC20 indexed tokenIn, uint256 donationAmount, GrantRound[] rounds);
 
+  // --- Constructor ---
   constructor(
     GrantRegistry _registry,
     ISwapRouter _router,
@@ -60,6 +71,8 @@ contract GrantRoundManager {
     router = _router;
     donationToken = _donationToken;
   }
+
+  // --- Core methods ---
 
   /**
    * @notice Creates a new GrantRound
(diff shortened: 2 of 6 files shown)


contracts/.solhint.json
```diff
@@ -2,14 +2,15 @@
   "extends": "solhint:recommended",
   "plugins": ["prettier"],
   "rules": {
-    "code-complexity": ["error", 7],
+    "code-complexity": ["error", 9],
```
thelostone-mc, line 5: haha the fav change in this PR <3

contracts/contracts/GrantRoundManager.sol
```diff
@@ -32,35 +24,70 @@ contract GrantRoundManager {
...
+
+  /// @notice Used during donations to temporarily store donation ratio sums to ensure totals are always 100%
+  mapping(IERC20 => uint256) internal donationRatios;
+
   /// @notice WETH address
-  address public constant WETH = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
+  IERC20 public constant WETH = IERC20(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
+
+  /// @notice Scale factor
+  uint256 internal constant WAD = 1e18;
```
corydickson, line 37: decimals here may lead to misleading amounts for tokens like USDC which only go to 1e6

contracts/contracts/GrantRoundManager.sol
```diff
@@ -32,35 +24,70 @@ contract GrantRoundManager {
...
+    uint256 amountIn;
+    uint256 amountOutMin; // minimum amount to be returned after swap
+    bytes path;
+  }
 
+  /// @notice Donation inputs and Uniswap V3 swap inputs: https://docs.uniswap.org/protocol/guides/swaps/multihop-swaps
+  struct Donation {
+    uint96 grantId; // grant ID to which donation is being made
+    IERC20 token; // address of the token to donate
+    uint256 ratio; // ratio of `token` to donate, specified as numerator where WAD = 1e18 = 100%
```
corydickson, line 52: for swaps with different decimals we could floor the amount for the tokens with the least amount of significant figures

contracts/contracts/GrantRoundManager.sol
```diff
@@ -32,35 +24,70 @@ contract GrantRoundManager {
...
     donationToken = _donationToken;
+
+    // Token approvals of common tokens
+    // TODO inherit from SwapRouter to remove the need for this approvals and extra safeTransferFrom before swap
+    IERC20(0x6B175474E89094C44Da98b954EedeAC495271d0F).safeApprove(address(_router), type(uint256).max); // DAI
+    IERC20(0xDe30da39c46104798bB5aA3fe8B9e0e1F348163F).safeApprove(address(_router), type(uint256).max); // GTC
+    IERC20(0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48).safeApprove(address(_router), type(uint256).max); // USDC
+    IERC20(0xdAC17F958D2ee523a2206206994597C13D831ec7).safeApprove(address(_router), type(uint256).max); // USDT
+    IERC20(0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599).safeApprove(address(_router), type(uint256).max); // WBTC
+    IERC20(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2).safeApprove(address(_router), type(uint256).max); // WETH
```
corydickson, line 86: I think we save gas by using the already defined public constant here

contracts/contracts/GrantRoundManager.sol
mds1: Hm, how so? Even if USDC amount where 1e6 =1 USDC, say you want 25% of 1 USDC, then you'd have `1e6 * 0.25e18 / 1e18 = 250000 = 0.25 USDC`

contracts/contracts/GrantRoundManager.sol
mds1: Seems related to the above comment, but I'm not sure what you mean here and where you need to floor things that isn't already floored by integer division in Solidity?

contracts/contracts/GrantRoundManager.sol
```diff
@@ -99,66 +126,97 @@ contract GrantRoundManager {
...
+    for (uint256 i = 0; i < _donations.length; i++) {
+      // Validate grant exists
+      require(_donations[i].grantId < registry.grantCount(), "GrantRoundManager: Grant does not exist in registry");
+
+      // Used later to validate ratios are correctly provided
+      donationRatios[_donations[i].token] += _donations[i].ratio;
+
+      // Validate round parameters
+      GrantRound[] calldata _rounds = _donations[i].rounds;
+      require(_rounds.length > 0, "GrantRoundManager: Must specify at least one round");
```
apbendi, line 155: I wonder if this is true. While it would possible to donate to a Grant by simply transferring directly to the Payee, we've also discussed that we'd want an event to be emitted even for non-round-associated contributions. So the question is, do we want a separate method for that, one that doesn't swap the donated token? Or do we just use the same method, allowing for no-round donations AND still swapping to the RoundManager's donation token?

contracts/contracts/GrantRoundManager.sol
corydickson: right but what if someone tries to swap 0.2500009 USDC and submits that to the swap method on uni won't it revert?

contracts/contracts/GrantRoundManager.sol
corydickson: maybe I'm missing something, but it's not related to the calculation per say but when we transfer an amount, emitting one value but really another has been sent to the contract

contracts/contracts/GrantRoundManager.sol
mds1: Do you have an example of how we might get such a value? When we divide by 1e18 at the end, we'll either be left 250,000 (0.25 USDC) or 250,001 (0.250001 USDC) because there's only 6 decimals of precision when do you `tokenDecimalsUsdc * xe18 / 1e18`

contracts/contracts/GrantRoundManager.sol
mds1: The `_donationAmount` that is transferred is the same value as we emit in the `GrantDonation` method. Sorry, I don't think I understand the issue you are pointing out 😬

contracts/contracts/GrantRoundManager.sol
```diff
@@ -32,35 +24,70 @@ contract GrantRoundManager {
...
     require(_donationToken.totalSupply() > 0, "GrantRoundManager: Invalid token");
 
+    // Set state
     registry = _registry;
     router = _router;
     donationToken = _donationToken;
+
+    // Token approvals of common tokens
+    // TODO inherit from SwapRouter to remove the need for this approvals and extra safeTransferFrom before swap
+    IERC20(0x6B175474E89094C44Da98b954EedeAC495271d0F).safeApprove(address(_router), type(uint256).max); // DAI
```
apbendi, line 81: Just want to make sure I understand: as currently written, the donate method will fail if it's not using one of these pre-approved tokens, and there is no mechanism for the contract to approve new ones. Correct?

Inheriting from SwapRouter means we'd become a custom v3 router, and users would have to trust (or verify by inspected verified source code) that we did not modify the router functionality in some malicious or unsafe way. Correct?

apbendi (commented):
There's a good amount of complexity in what we're doing here, so really nice job tackling this 👍

I left some comments and questions and will follow up further based on responses. Thanks!

contracts/contracts/GrantRoundManager.sol
mds1: Good call, I think using the same method and removing this `require` is the way to go since everything else should still behave the same IMO. Going to implement that for now unless anyone objects

contracts/contracts/GrantRoundManager.sol
```diff
@@ -32,35 +24,70 @@ contract GrantRoundManager {
   /// @notice Address of the ERC20 token in which donations are made
   IERC20 public immutable donationToken;
 
+  /// @notice Used during donations to temporarily store swap output amounts
+  mapping(IERC20 => uint256) internal swapOutputs;
```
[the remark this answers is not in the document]
mds1: > Mappings can only have a data location of `storage`

([source](https://docs.soliditylang.org/en/v0.8.7/types.html#mapping-types))

Given that, it costs 20k gas to write and you get a refund of 19.9k when clearing in the same transaction as the initial write, so it ends up being cheap ([source](https://github.com/wolflo/evm-opcodes/blob/main/gas.md#a7-sstore)). Open to other ideas on how to do this, though this seemed the cleanest

contracts/contracts/GrantRoundManager.sol
mds1: Correct on all counts! If for whatever reason we don't end up inheriting from SwapRouter, I think we'd want a public method where you pass in a token address and it approves the SwapRouter to spend MaxUint of that token. Otherwise we have to check allowances every donation which would get very costly

contracts/contracts/GrantRoundManager.sol
```diff
@@ -99,66 +126,97 @@ contract GrantRoundManager {
...
+
+      // Do nothing if the swap input token equals donationToken
+      if (_tokenIn == donationToken) {
+        swapOutputs[_tokenIn] = _swaps[i].amountIn;
+        continue;
+      }
+
+      // Transfer input token to this contract if required
+      // TODO inherit from SwapRouter to remove the need for this
+      if (_tokenIn != WETH || msg.value == 0) {
```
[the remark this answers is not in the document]
mds1: When `_tokenIn == WETH`, we need a way to distinguish between an ETH donation, which we forward to the SwapRouter which converts it to WETH, and an actual WETH token donation.

A limitation that results from this is you can't donate WETH and ETH in the same transaction, which I think is an edge case that's ok if we don't support. Wonder if Uniswap V3 router would let you multicall an ETH->DAI and WETH->DAI swap in the same transaction. My guess it supports that via multicall where you deposit ETH->WETH in the router as the first call, then execute the swap as the next call. Perhaps pulling out into an issue?

contracts/contracts/GrantRoundManager.sol
```diff
@@ -99,66 +126,97 @@ contract GrantRoundManager {
...
+    for (uint256 i = 0; i < _donations.length; i++) {
+      // Get data
+      GrantRound[] calldata _rounds = _donations[i].rounds;
+      uint96 _grantId = _donations[i].grantId;
+      IERC20 _tokenIn = _donations[i].token;
+      uint256 _donationAmount = (swapOutputs[_tokenIn] * _donations[i].ratio) / WAD;
+      require(_donationAmount > 0, "GrantRoundManager: Donation amount must be greater than zero"); // verifies that swap and donation inputs are consistent
+      address _payee = registry.getGrantPayee(_grantId);
+
+      // Execute transfer
```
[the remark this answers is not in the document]
mds1: That was already removes from the `GrantRound` contract

contracts/test/GrantRoundManager.test.ts
```diff
@@ -166,93 +175,282 @@ describe('GrantRoundManager', () => {
...
-        grantId: 0,
-        rounds: [mockRound.address],
-        path: await encodeRoute(['dai', 'eth', 'gtc']),
-        deadline: farTimestamp, // arbitrary date far in the future
-        amountIn: '1',
-        amountOutMinimum: '0',
-      };
-
-      // Fund the first user with tokens
+      swap = { amountIn: '1', amountOutMin: '0', path: await encodeRoute(['dai', 'eth', 'gtc']) };
```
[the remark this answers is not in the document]
mds1: FIxed in 4fef97a4c52eb265f8a4db4ef8b39e05dd5ea5c0

contracts/contracts/GrantRoundManager.sol
mds1: Updated in 4fef97a4c52eb265f8a4db4ef8b39e05dd5ea5c0

contracts/contracts/GrantRoundManager.sol
```diff
@@ -99,66 +126,97 @@ contract GrantRoundManager {
...
+   * @dev Caller must ensure the input tokens to the _swaps array are unique
+   * @dev Does not verify
    */
-  function swapAndDonate(Donation calldata _donation) external payable {
+  function donate(
+    SwapSummary[] calldata _swaps,
+    uint256 _deadline,
+    Donation[] calldata _donations
+  ) external payable {
     // --- Validation ---
```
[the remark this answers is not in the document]
mds1: Split up in 4fef97a4c52eb265f8a4db4ef8b39e05dd5ea5c0

contracts/contracts/GrantRoundManager.sol
apbendi: Cool, I agree, as it will also make the frontend checkout logic must easier as well.

contracts/contracts/GrantRoundManager.sol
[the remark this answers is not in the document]
apbendi: 🤦  Duh I should have known that. I think our only option would be to use something like dynamic arrays of tuples. This would have O(N) lookup time. I'm curious where the break-even point would be on N, but given we can get refunded for the storage slots it's probably pretty low. It'd also be a much uglier implementation.

Another option would be to require-by-convention that donations be grouped by token. Then you could do everything by array index, with inner while loops that increment the index when the token changes. Again, more complex implementation. Also  adds work on the frontend (albeit pretty minor). Might be worth exploring, but feels like premature optimization at the moment.

Let's keep it as-is and revisit whenever we get around to gas golfing.

contracts/contracts/GrantRoundManager.sol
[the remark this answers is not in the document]
apbendi: Got it. Created #98 to track this.

contracts/contracts/GrantRoundManager.sol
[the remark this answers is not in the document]
mds1: For context on the function naming, I used the same convention as the `app` of naming things `<parent-purpose>` to make it easier to understand something's purpose / where it fits in. So here the parent method is `donate` and we have 3 sub-methods of validation, swaps, and execution. I don't feel strongly about the naming, so I have no objections to renaming them if that reasoning doesn't convince you

Re: modifier, I'd personally vote against that since it's only used one time, so I'm not sure pulling it out adds any value, and just costs extra gas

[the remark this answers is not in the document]
mds1: Though I do think we should add underscores to prefix internal methods regardless, but will hold off pending a final decision on naming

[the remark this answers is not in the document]
apbendi: Wasn't aware we had that convention on the frontend. Does that come from Vue?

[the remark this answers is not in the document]
mds1: Yea it's based off their style guide: https://vuejs.org/v2/style-guide/#Tightly-coupled-component-names-strongly-recommended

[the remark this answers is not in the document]
apbendi: Got it. I can see how that convention makes sense for Components, but not sure it fits well for helper methods. The point of breaking the original function up was for code clarity, so easily grokable names seems pretty important. That said, don't feel strongly enough to block the PR or anything! I think I've made my case and will let you exercise your judgement!

contracts/contracts/GrantRoundManager.sol
apbendi: Got it, sounds good. I think we want to go this route (ha!). Created this issue to track as a separate PR: https://github.com/dcgtc/dgrants/issues/99

apbendi (approved):

contracts/contracts/GrantRoundManager.sol
thelostone-mc: Just a clarification -> Even if we inherited the list -> wouldn't we still have to do an approve for every token on the list right ? (the only diff is we won't be hardcoding every token within this contract )

contracts/contracts/GrantRoundManager.sol
```diff
@@ -99,66 +126,119 @@ contract GrantRoundManager {
...
-      require(
-        donationToken == _rounds[i].donationToken(),
-        "GrantRoundManager: GrantRound's donation token does not match GrantRoundManager's donation token"
-      );
+  /**
+   * @dev Validates the the inputs to a donation call are valid, and reverts if any requirements are violated
+   * @param _donations Array of donations that will be executed
+   */
+  function _validateDonations(Donation[] calldata _donations) internal {
+    // TODO consider moving this to the section where we already loop through donations in case that saves a lot of
```
thelostone-mc, line 160: are we saying 
- keep the function 
- let the argument be Donation _donation
- it gets invoked  as the first thing in _transferDonations 

It does seem a smart way to save gas but this would mean -> we'd do it after the swap happens which is I am not very comfortable about (but the argument could be made -> if you invoke it wrong -> it's not the contract's fault )

contracts/contracts/GrantRoundManager.sol
thelostone-mc: @corydickson I'm just running the example in my head and i don't see an issue either.
When donate is invoked for a given token -> sum of ratios should be 1e18.

So if we have a token with 16 decimals and contributions are made to 3 diff grants -> 5 , 5, 10
The ratio would be 25%, 25%, 50%.
The frontend would backfill the last 2 decimals with 0 and send it. 


So once we do the swap to 18 decimal token -> you'd still have the same ratio of 25%, 25%, 50%  right ?

The only scenario where this might mess things up is if we had a token > 18 decimals  (which is not something we need to be bothered about)

thelostone-mc (approved):

contracts/contracts/GrantRoundManager.sol
mds1: Yea pretty much! The benefits of the current approach are that you save gas in the event of a failure (by reverting before the swap), but downside is more gas overall since you repeat a loop. I don't think the extra gas is significant, so I think we should we leave this as-is for readability, and come back to it later when we focus on optimizing gas usage

contracts/contracts/GrantRoundManager.sol
mds1: So here is the flow when you normally use Uniswap as a user:
1. Approve router to spend your tokens
2. Call swap on router, which takes your tokens and executes swap 
3. (1 total approval, 1 transferFrom)

Here is our current flow:
1. Approve manager to spend your tokens
2. Manager approves router to spend it's tokens
3. Call donate on manager, which takes your tokens, the router takes tokens from manager to execute swap (no more approvals)
4. (2 total approvals, 2 transferFroms)

So once we inherit from SwapRouter, we'll migrate to the top flow, which will reduce approvals and transfers

PR #103, opened 2021-08-12
Implementing the designed navigation to the layoutheader and hookup the existing logic

scco:
- had to get rid of the loop for main navigation items and the variable - as the navigation have custom seperators. 
- simplifyed  the if-else logic to display the cart, user-nav and wrong-network so it works as i want in the menu


apbendi:
Closing this PR for the process described in https://github.com/dcgtc/dgrants/issues/112

PR #104, opened 2021-08-12
feat: add utils package

thelostone-mc:
#### Description 

Introduced utils package with 

- readme 
- src/merkle-distributor


mds1 (changes requested):
Looks great! I haven't tested it but will defer to @wildmolasses on verifying it's all setup properly. Only left super tiny comments:
- Some phrasing suggestions, you can ignore these if you prefer
- Use permalinks for the code references (includes commit hash in URL) instead of linking to master

corydickson (approved):

PR #109, opened 2021-08-16
Extend GrantRoundManager from Uniswap v3's SwapRouter

mk1r:
- downgrade compiler version from 0.8.6 down to 0.7.6 for Uniswap v3 SwapRouter
- downgraded OpenZeppelin version from 4.2 to 3.4.1 and using solc 0.7
- Inheriting SwapRouter instead of injecting it through the constructor
- It's compiling well and working with yarn dev
- 46 tests are passing, but 1 (out of 47) is failing

test to check: bulk donation, single token; input tokens DAI, ETH, GTC

update: all tests are passing


contracts/contracts/GrantRoundManager.sol
```diff
@@ -18,22 +18,16 @@ contract GrantRoundManager {
...
 
-  /// @notice Used during donations to temporarily store swap output amounts
   mapping(IERC20 => uint256) internal swapOutputs;
 
-  /// @notice Used during donations to temporarily store donation ratio sums to ensure totals are always 100%
   mapping(IERC20 => uint256) internal donationRatios;
 
   /// @notice WETH address
-  IERC20 public constant WETH = IERC20(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
+  IERC20 public WETH;
```
mds1, line 29: Let's get rid of this WETH and just use the `WETH9` that's already part of the `SwapRouter`. Currently the contract has `IERC20 WETH` and `address WETH9` which is a bit redundant/unnecessary

This will also require us to change `WETH` to `IERC20(WETH9)` in other places

contracts/contracts/GrantRoundManager.sol
```diff
@@ -63,27 +57,27 @@ contract GrantRoundManager {
...
+  ) SwapRouter(_factory, _WETH9) {
     // Validation
     require(_registry.grantCount() >= 0, "GrantRoundManager: Invalid registry");
-    require(address(_router).isContract(), "GrantRoundManager: Invalid router"); // Router interface doesn't have a state variable to check
     require(_donationToken.totalSupply() > 0, "GrantRoundManager: Invalid token");
 
     // Set state
     registry = _registry;
-    router = _router;
+    WETH = IERC20(_WETH9);
```
mds1, line 70: As above, we'll be able to delete this

contracts/contracts/GrantRoundManager.sol
```diff
@@ -63,27 +57,27 @@ contract GrantRoundManager {
...
-    IERC20(0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48).safeApprove(address(_router), type(uint256).max); // USDC
-    IERC20(0xdAC17F958D2ee523a2206206994597C13D831ec7).safeApprove(address(_router), type(uint256).max); // USDT
-    IERC20(0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599).safeApprove(address(_router), type(uint256).max); // WBTC
-    WETH.safeApprove(address(_router), type(uint256).max); // WETH
+    IERC20(0x6B175474E89094C44Da98b954EedeAC495271d0F).safeApprove(address(this), type(uint256).max); // DAI
+    IERC20(0xDe30da39c46104798bB5aA3fe8B9e0e1F348163F).safeApprove(address(this), type(uint256).max); // GTC
+    IERC20(0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48).safeApprove(address(this), type(uint256).max); // USDC
+    IERC20(0xdAC17F958D2ee523a2206206994597C13D831ec7).safeApprove(address(this), type(uint256).max); // USDT
+    IERC20(0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599).safeApprove(address(this), type(uint256).max); // WBTC
+    WETH.safeApprove(address(this), type(uint256).max); // WETH
```
mds1, line 80: We can get rid of all these approvals (and the comments above them). These were only needed because we had to approve the router to spend this contract's tokens, but now this contract is the router so this is not required

contracts/contracts/GrantRoundManager.sol
```diff
@@ -204,15 +198,15 @@ contract GrantRoundManager {
...
+      ExactInputParams memory params = ExactInputParams(
         _swaps[i].path,
         address(this), // send output to the contract and it will be transferred later
         _deadline,
         _swaps[i].amountIn,
         _swaps[i].amountOutMin
       );
       uint256 _value = _tokenIn == WETH && msg.value > 0 ? msg.value : 0;
-      swapOutputs[_tokenIn] = router.exactInput{value: _value}(params); // save off output amount for later
+      swapOutputs[_tokenIn] = this.exactInput{value: _value}(params); // save off output amount for later
```
mds1, line 209: `this.exactInput` makes an external call back to this contract, which mainly just wastes gas. Instead let's change the `exactInput` method from `external` to `public` so we can remove the `this.` keyword. Doing that means we'll also be able to remove the `value` passed with the call, since that's only needed since we're forwarding the ETH to an external call

contracts/contracts/SwapRouter.sol
```diff
@@ -0,0 +1,241 @@
...
+import "@uniswap/v3-periphery/contracts/base/Multicall.sol";
+import "@uniswap/v3-periphery/contracts/base/SelfPermit.sol";
+import "@uniswap/v3-periphery/contracts/libraries/Path.sol";
+import "@uniswap/v3-periphery/contracts/libraries/PoolAddress.sol";
+import "@uniswap/v3-periphery/contracts/libraries/CallbackValidation.sol";
+import "@uniswap/v3-periphery/contracts/interfaces/external/IWETH9.sol";
+
+/// @title Uniswap V3 Swap Router
+/// @notice Router for stateless execution of swaps against Uniswap V3
+contract SwapRouter is
```
mds1, line 22: Curious why you pulled this into its own file instead of using the version from `node_modules`? Did you make any changes from the original SwapRouter? 

Given that we should make some tweaks like going from `external` to `public` as mentioned above, I'm ok with this, but we should clearly document in the comments of this file what changes we made from the default SwapRouter

contracts/package.json
```diff
@@ -33,7 +33,7 @@
     "shelljs": "^0.8.4",
     "solhint": "^3.3.4",
     "solhint-plugin-prettier": "^0.0.5",
-    "solidity-bytes-utils": "0.8.0",
+    "solidity-bytes-utils": "https://github.com/GNSPS/solidity-bytes-utils#feat/update-0.7.0",
```
mds1, line 36: Ah we can remove this dependency now, as the Uniswap module already has an 0.7.x compatible version of this

mds1 (changes requested):
Looking good! Added some comments about some additional cleanup

Regarding the test failure: One of my comments is about removing `this.` to avoid an unnecessary external call. After that, you'll see a comment that says `TODO inherit from SwapRouter to remove the need for this`. If you remove that line, your error in the swap test might go away. Basically we no longer need to first transfer tokens from the user to our contract before executing the swap, and instead we can rely on the swap to execute that transfer

contracts/package.json
mk1r: alright cool, and yeah I couldn't find the version I needed in the npm / yarn registry (so got it from github)

contracts/contracts/SwapRouter.sol
mk1r: Yeah so I looked there first, but I didn't see the contract there for some reason. This file should probably be moved elsewhere though until we have our own router. I didn't modify it all so far - will be sure to add comments when I do.

contracts/contracts/GrantRoundManager.sol
mk1r: tried this, but errors out on function call options (used for external functions, not public functions)

contracts/contracts/GrantRoundManager.sol
```diff
@@ -18,22 +18,16 @@ contract GrantRoundManager {
...
   mapping(IERC20 => uint256) internal swapOutputs;
 
-  /// @notice Used during donations to temporarily store donation ratio sums to ensure totals are always 100%
   mapping(IERC20 => uint256) internal donationRatios;
 
   /// @notice WETH address
-  IERC20 public constant WETH = IERC20(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
+  IERC20 public WETH;
 
-  /// @notice Scale factor
```
thelostone-mc, line 36: would we wanna still keep this / document it better @mds1 ?

contracts/contracts/GrantRoundManager.sol
thelostone-mc: would it make sense to stick to using `this` and leave a TODO to tweak the exactInput into public function and fix the other things in the custom SwapContract which break as a result ?

contracts/contracts/SwapRouter.sol
thelostone-mc: would it be worth adding a dev note stating we forked the code from the link and we're using that

contracts/contracts/GrantRoundManager.sol
mds1: Oh yea, we should keep this comment. Maybe we phrase it as

```
/// @notice Scale factor on percentages when constructing `Donation` objects. One WAD represents 100%
```

contracts/contracts/GrantRoundManager.sol
mds1: Hmm, my initial reaction is that it shouldn't be much more work to do things the "right" way here, but if keeping `this` is a lot simpler then I'm ok with that for now

contracts/contracts/SwapRouter.sol
mds1: Good call, yea in the comments we should:
- Link to the GitHub permalink (i.e. the URL containing the commit hash) this was copied from
- Document all changes from the original source (including that our prettier config was ran against it)

contracts/contracts/GrantRoundManager.sol
mk1r: was getting some errors with the notice tags after changing compiler versions: ```contracts/GrantRoundManager.sol:28:3: DocstringParsingError: Documentation tag @notice not valid for non-public state variables.```

contracts/contracts/SwapRouter.sol
mk1r: should comments be somewhere at the top of the file? left a comment above the function but with these new changes, at the top might make more sense

contracts/contracts/GrantRoundManager.sol
mk1r: yes we're all good here - thanks for the input though thelostone-mc

contracts/contracts/GrantRoundManager.sol
mds1: Ah sorry, yes that makes sense, sorry should be `@dev` then.

`@notice` is used to document public-facing methods/variables, and `@dev` for internal methods/variables (or for additional additional developer-related info somewhere)

contracts/contracts/SwapRouter.sol
mds1: Yea if you could put a natspec comment at the top of the file that would be great! Check out `GrantRegistry.sol` and `MerkleGrantRoundPayout.sol` (you'll need to rebase against `main` for this one) for examples where we do this. The latter was also based on another Uniswap contract so might be a good comment template

mds1:
@mk1r I saw you left a message in Discord, but it seems either me or Discord is having problems loading Discord messages so I didn't see your message. From the commit messages it seems this might be ready for final review? If so:
- Rebase against main just to make sure there's no issues
- Run `yarn lint` in the contracts directory and resolve those issues so the CI passes (you can see the last run on this branch failed). They should all be small things, e.g. editing `solhint.json` to fix the expected solc version, but if you need help with any of it let me know

mk1r:
@mds1 yes, still seeing some issues there - thanks for the suggestions, resolved lint errors and added in natspec comments

just gotta rebase now

mds1:
@mk1r If you are also have Discord problems, the solution was go to the “role-selector” channel and click whichever roles you want, and then it will be back to normal!

mk1r:
@mds1 cool did that now, but it seems the channel has completely disappeared for me

mds1:
Hmm, I just clicked all 3 roles so I can see all channels. I'm not sure exactly which one corresponds to the "ws-decentralize-gitcoin" channel, so maybe just try selecting them all

mk1r:
@mds1 yes I tried selecting all too haha - will check it agian

mk1r:
marked changes request as resolved, ready to be merged

contracts/contracts/SwapRouter.sol
```diff
@@ -0,0 +1,249 @@
...
+import "@uniswap/v3-periphery/contracts/base/SelfPermit.sol";
+import "@uniswap/v3-periphery/contracts/libraries/Path.sol";
+import "@uniswap/v3-periphery/contracts/libraries/PoolAddress.sol";
+import "@uniswap/v3-periphery/contracts/libraries/CallbackValidation.sol";
+import "@uniswap/v3-periphery/contracts/interfaces/external/IWETH9.sol";
+
+/**
+ * @title Uniswap V3 Swap Router
+ * @notice Router for stateless execution of swaps against Uniswap V3
+ * @dev code sourced from https://github.com/Uniswap/uniswap-v3-periphery/blob/main/contracts/SwapRouter.sol
```
mds1, line 23: Recommend using permalink in case the current `main` ever changes

```suggestion
 * @dev code sourced from https://github.com/Uniswap/uniswap-v3-periphery/blob/1f5b385675eee30a77ee10f4dfd6aa0aaa766026/contracts/SwapRouter.sol
```

mds1: Since this is quick, did this for you in 5166c1e0d513569b12b42ce41595121e3bcfaaaf

thelostone-mc (approved):

thelostone-mc:
Closing this out as we have a messy rebase and getting #120 in instead

PR #111, opened 2021-08-16
Cart page setup

### LINKED ISSUE #94: Implement Cart Checkout (spec)
### Adding To Cart

- [x] When on the grant detail view, there should be an "Add to Cart" button
- [x] When viewing a list of grants, there should an Cart icon for add to cart
- [x] When either aforementioned button is clicked, the following data should be added to local storage:
     - [x] The grantId
     - [x] The donation token address (default to Dai)
     - [x] The donation token amount (default to 5 in raw decimal format)
- [x] There should be utility methods that abstract writing/reading cart data from local storage
- [x] The data should be returned in the following format:

#### Sample Schema

````js
[
    {
        grantId: 5,
        contributionTokenAddress: "0x6b175474e89094c44da98b954eedeac495271d0f",
        contributionAmount: "5000000000000000000",
    },
    {
        grantId: 7,
        contributionTokenAddress: "0xde30da39c46104798bb5aa3fe8b9e0e1f348163f",
        contributionAmount: "1337000000000000000",
    },
]
````

### Cart Page

- [x] The cart page should list the grants the user has added
    - [ ] It should load metadata for the grant via IPFS; dependent on #86 
    - [x] It should display the token and the human readable amount
- [x] The cart page should show a summarizations of how much of each token is being donated in total (e.g. "Donating 25 Dai + .4 ETH")
- [x] The cart page should allow the user to update their token and amount individually for each row
     - [x] The UI should list the following tokens in the UI (for now): DAI, ETH, USDC, GTC, UNI
     - [x] The list of supported tokens should use the [token list](https://uniswap.org/blog/token-lists/) format, and should be generalized to work with tokens of any name/decimals/symbols etc....
- [x] The cart page should allow the user to remove items from their cart
- [x] The cart page should allow users to clear their cart

### Additional Notes

* The implementation should tend toward @scco's [figma designs](https://www.figma.com/file/sUQSYyPfooqTcVdajfZm0t/dgrants?node-id=92%3A80), but do not need to match it precisely for now; more precise styling will be handled in a separate task
* We can also leave matching estimates and USD conversions out of the initial implementation and address them as separate features

### PR DESCRIPTION
**Pull request by mds1**
Closes #94

Notes:
- Use the new "Cart" menu item in the header to view your cart
- Visiting the `/cart` page directly is not well supported. Waiting on #86 to be merged to better support that, since like that PR, here we have a similar requirement of "needing `poll` to execute at least once
- Fulfills spec in #94 and adds new `BaseSelect` component to handle dropdowns menu
- Layout is loosely based off the Figma mockups but I didn't spend too much time on this

### CODE REVIEW

File: app/src/utils/utils.ts [hunk 1]
```diff
@@ -3,21 +3,35 @@
...
 export function formatAddress(address: string) {
   if (!address || address.length !== 42) return null;
   return `${address.slice(0, 6)}...${address.slice(38)}`;
 }
 
-// Navigates to the specified page and pushes a new entry into the history stack
-export async function pushRoute(to: RouteLocationRaw) {
-  await router.push(to);
+// Expects a unix timestamp and will return a human readable message of how far in the past/future it is
+export function daysAgo(val = 0) {
```
thelostone-mc, line 19: push this into utils package ?

app/src/utils/utils.ts
```diff
@@ -28,22 +42,69 @@ export function isValidAddress(val: string | undefined) {
...
 
-// Expects a unix timestamp and will return a human readable message of how far in the past/future it is
-export function daysAgo(val = 0) {
-  // Use a formatter to establish "in 10 days" vs "10 days ago"
-  const formatter = new Intl.RelativeTimeFormat();
-  // Number of days since now
-  const deltaDays = (val * 1000 - Date.now()) / (1000 * 3600 * 24);
+// --- Grants + Cart ---
+const CART_KEY = 'cart';
+const DEFAULT_CONTRIBUTION_TOKEN_ADDRESS = '0x6B175474E89094C44Da98b954EedeAC495271d0F'; // DAI
```
thelostone-mc, line 47: should we use from token.dai.address?

app/src/utils/utils.ts
```diff
@@ -28,22 +42,69 @@ export function isValidAddress(val: string | undefined) {
...
+const DEFAULT_CONTRIBUTION_AMOUNT = 5; // this is converted to a parsed BigNumber at checkout
```
thelostone-mc, line 48: would it make sense having a constant files and dumping all these there ?

app/src/utils/utils.ts
```diff
@@ -28,22 +42,69 @@ export function isValidAddress(val: string | undefined) {
...
-  // Format "days ago" as string
-  return formatter.format(Math.round(deltaDays), 'days');
+// Loads cart data
+export function loadCart(): CartItemOptions[] {
+  // Return empty array if nothing found
+  const rawCart = localStorage.getItem(CART_KEY);
+  if (!rawCart) return [];
+
+  // Parse the data. If the data is an array, return it,
+  const cart = JSON.parse(rawCart);
```
thelostone-mc, line 57: would we want to wrap this in a try catch?

app/src/utils/utils.ts
```diff
@@ -28,22 +42,69 @@ export function isValidAddress(val: string | undefined) {
...
 }
 
-// Expects a unix timestamp and will return a human readable message of how far in the past/future it is
-export function daysAgo(val = 0) {
-  // Use a formatter to establish "in 10 days" vs "10 days ago"
-  const formatter = new Intl.RelativeTimeFormat();
-  // Number of days since now
-  const deltaDays = (val * 1000 - Date.now()) / (1000 * 3600 * 24);
+// --- Grants + Cart ---
+const CART_KEY = 'cart';
```
thelostone-mc, line 46: do we wanna wrap this within a dgrants-core object ? cause cart is a generic name so by wrapping it within a dgrants object, we avoid random conflicts with other things (like imagine gitcoin cgrants vs dgrants) 
Or the other option is we use a prefix `dgrant_` for any localStorage variable we create

app/src/utils/utils.ts
```diff
@@ -28,22 +42,69 @@ export function isValidAddress(val: string | undefined) {
...
+const DEFAULT_CONTRIBUTION_TOKEN_ADDRESS = '0x6B175474E89094C44Da98b954EedeAC495271d0F'; // DAI
+const DEFAULT_CONTRIBUTION_AMOUNT = 5; // this is converted to a parsed BigNumber at checkout
 
-  // Format "days ago" as string
-  return formatter.format(Math.round(deltaDays), 'days');
+// Loads cart data
+export function loadCart(): CartItemOptions[] {
```
thelostone-mc, line 51: This file is starting to grow in size -> would it make sense to move all of this into `app/src/utils/cart.js`

thelostone-mc (commented):
Left a few comments + questions.
overall it's looking good :D

app/src/components/BaseSelect.vue
```diff
@@ -0,0 +1,102 @@
+<template>
+  <Listbox as="div" :modelValue="modelValue" @update:model-value="$emit('update:modelValue', $event)">
+    <!-- <ListboxLabel class="block text-sm font-medium text-gray-700"> {{ description }} </ListboxLabel> -->
```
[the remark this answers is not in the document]
mds1: Originally I left it in with intent to uncomment if/when we want the option to add labels above dropdowns, but it's simple enough that agreed we should remove it

app/src/utils/utils.ts
mds1: I think the question to answer here is "what goes into the shared utils package, and what goes into the shared types package?" I've been organizing things as if the answer is "only utils/types that _are_ shared between other packages". If we think that answer should be "any utils/types that _may_ be helpful to someone integrating their own frontend", then I agree we should move it. 

Given that, I'd say we leave this as-is and figure this out in a separate discussion, which I created here: https://github.com/dcgtc/dgrants/discussions/114

(Side note: I rearranged this file a bit which is why it looks like there's a lot of changes).

app/src/utils/utils.ts
mds1: That `token` mapping from symbol to token info only exists in the `contracts` package, and I don't see a need for mapping symbol to token in the app directory, so my opinion is that we can keep this as is.

app/src/utils/utils.ts
mds1: The reason I put them here instead of the existing `app/src/utils/constants.ts` file is that these don't ever need to be used outside this file, so we can just define them here for simplicity/readability. Open to being convinced otherwise though!

app/src/utils/utils.ts
mds1: What exactly do you mean by a "dgrants-core object"? Local storage is domain-specific, so the naming can only conflict with local storage keys within the dgrants app, and never with cgrants, so by definition (almost) anything in localstorage must be dgrants specific. 

The only exception is that browser extensions can read/write from local storage, though I'm not worried about a conflict there.

If we do want to prefix local storage keys, we also have to update the key names defined in `app/src/store/settings.ts`, and therefore it's probably worth it's own PR

app/src/utils/utils.ts
mds1: Pulling them out into `cart.js` seems like a good idea

mds1: Done in 97fc542e1c60e1958e367ff22e3c96ef23d156de

app/src/utils/utils.ts
mds1: Done in 97fc542e1c60e1958e367ff22e3c96ef23d156de

app/src/components/BaseSelect.vue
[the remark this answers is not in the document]
mds1: Done in 97fc542e1c60e1958e367ff22e3c96ef23d156de

app/src/utils/utils.ts
thelostone-mc: Ah yup you're right ! ignore my comment

app/src/utils/utils.ts
thelostone-mc: lgtm

app/src/utils/utils.ts
thelostone-mc: Haha alrighty ! thanks for creating the discussion

thelostone-mc (approved):
LGTM 🎉

gdixon (approved):
Nice work Matt!! 🚀

apbendi (commented):
Looking and working great! A couple of things we should also have come to mind:

1. When a grant is already in your cart, the button should say "Remove from cart" with the corresponding functionality
2. Clicking the cart row (or at least the grant ID for now— eventually will be the title) should link back to the grant detail view
3. I think we can safely remove the "Save as collection" link for now, as collections will not be in the MVP

mds1:
@apbendi Just addressed those changes, and also created #117 to track some changes we should make. Based on how things unfolded, I didn't realize that architecture would be better until towards the end of the PR, and at this point I think we should merge this and handle that separately to keep concerns nicely separated and ensure we don't have this PR open for too long / need to get re-reviewed and tested

apbendi (approved):
Looks great! Nice work 👍

Issue #121, opened 2021-08-20
Load and Show Grant Metadata in the Cart


apbendi:
Currently the cart shows the Grant ID and the metadata URL in the row. Now that the app actually loads and shows the metadata for grants in other contexts, we want to show it in the cart as well. For this issue, we should remove the ID/URL and show the grant title in the cart row.

apbendi:
closed by #123

PR #123, opened 2021-08-20
Refactor cart store + use grant metadata in cart

### LINKED ISSUE #117: Move `cart` into it's own store
Right now a user's `cart` is a self-contained reactive object in `Cart.vue`. The cart ref and the cart data in localStorage must always be in sync, and the helper methods in `src/utils/cart.ts` can only modify localStorage, so we have a few hacks to ensure things stay in sync:
- In `Cart.vue` we have a watcher, as well as a few special helper methods called `clearCartAndUpdateState`, `removeItemAndUpdateState`, and `syncCart`
- In `GrantRegistryList.vue` and `GrantRegistryGrantDetail.vue`, we have a special `forceRender` ref used in conjunction with an `updateCart` method to force a re-render of the component when localStorage changes

This architecture and reactivity approach is confusing and doesn't scale well. Instead, all cart interactions and helpers should be moved from their respective files into a new store called `src/store/cart.ts`. The cart helpers in `src/utils/cart.ts` can be moved into this file also

This store will ensure that all methods that modify the cart modify both local storage AND the store simultaneously, so we can more naturally get reactivity without the current hacky approaches.

Additionally, the cart helpers in the existing `cart.ts` are a bit inconsistent in that some take a grant ID input, and others take a full object. The PR for this issue is a good time to standardize these interfaces as well to make them consistent

mds1:
Closes #117, #121

- Moves all cart data into it's own store
- Grant metadata resolved from IPFS is now used on the cart page

To test
- Make sure cart and localStorage stay in sync when cart is modified on the grant registry page, the grant detail page, and the cart page
- Make sure grant names (e.g. "Test ipfs 1") are shown on the cart page


apbendi (approved):
Looks and works great. Nice work @mds1 👌

thelostone-mc (approved):

PR #128, opened 2021-08-22
feat: fetch GrantRoundDetails from ipfs

thelostone-mc:


thelostone-mc:
**Q**
- would it make sense to have a function on the registry contract which lets you fetch a list of grantID.
not needed right now but for future

app/src/views/GrantRoundDetails.vue
```diff
@@ -48,11 +50,12 @@
...
       </p>
-      <p>
-        Metadata URL: <a class="link" :href="grantRound.metaPtr" target="_blank">{{ grantRound.metaPtr }}</a>
-      </p>
 
       <button @click="startContributing" class="btn btn-primary mt-6">Contribute to the matching fund</button>
+
+      <!-- Grants Belonging to Grants Round -->
+      <!-- TODO: isGrantRoundView=true -->
+      <GrantRegistryList />
```
thelostone-mc, line 58: Q: 

- Right now we have a single `poll` function that fetches all grants and grants Rounds. 
With this PR when the user visits the grant round page -> they end up being shown a subset of the grants as opposed to all the grants. I'm not clear on how / where to do the filtering so that all grants aren't shown. I'm trying to avoid writing a duplicate of GrantRegistryList as only the data needs to be updated

- I need to pass isGrantRoundView so that the title is hidden when GrantRegistryList is rendered but I'm not sure why I am not able to do that

apbendi: If I'm understanding the `poll` code correctly, it currently fetches every grant in the registry and the round data from on chain. We need to also fetch the round metadata. Since we'll then have all the data required, the filtering happens on the round page. Looking at the `grants` field after the GrantsRound is returned, we should have a list of ids like `[1, 2, 4, 10]`. So the Round page should filter the global registry grant list down to only those found in the `grants` field, and simply show those.

As a side note, I don't think this architecture will work long term, as the number of grants will be an impediment to this zealous up-front fetching. But it should be ok for the mvp round.

mds1: I think there are a few options, depending on whether we still want to keep a `GrantRegistryList` page that lists all grants, regardless of the round the grant is in. For all of them, we'll want to move the `GrantRegistryList.vue` file from `/views` (i.e. pages) to `/components` (i.e. reusable components) and possibly rename it. Then either
1. Change the `grants` variable from something that is read from the data store to a prop that is passed in
2. Add a `filterFn` prop that takes a method used to filter the `grants` data from the list of all grants to some subset, i.e. grants by round, and in the `setup()` method return the filtered grants list (probably as a computed property) instead of just returning grants as-is

> As a side note, I don't think this architecture will work long term, as the number of grants will be an impediment to this zealous up-front fetching. But it should be ok for the mvp round.

Agreed, we have various parts that won't scale well, often with `TODO` comments. Probably should create a "scaling" meta-issue at some point to track these

mds1:
> would it make sense to have a function on the registry contract which lets you fetch a list of grantID.
not needed right now but for future

Do you mean fetch the details (e.g. `payee`) for a list of grantIDs? If so I agree that'd be a useful getter for frontends, but if we don't need it now, maybe not worth adding at the moment and instead creating an issue to track it?

mds1:
@thelostone-mc Looks like CI is failing with `Error: @dgrants/contracts: test/MerkleGrantRoundPayout.test.ts(12,10): error TS2305: Module '"../../types/src/grants"' has no exported member 'Claim'.`

app/src/store/data.ts
```diff
@@ -32,9 +32,12 @@ const roundManager = ref<Contract>();
 // Most recent data read is saved as state
 const lastBlockNumber = ref<number>(0);
 const lastBlockTimestamp = ref<number>(0);
+
 const grants = ref<Grant[]>();
+let grantMetadata = ref<Record<string, GrantMetadataResolution>>({});
```
mds1, line 37: Why the change from `const` to `let`? Refs should pretty much always be `const`, not sure if there is a reason to use `let` here

app/src/store/data.ts
```diff
@@ -32,9 +32,12 @@ const roundManager = ref<Contract>();
...
+
 const grantRounds = ref<GrantRounds>();
-const grantMetadata = ref<Record<string, GrantMetadataResolution>>({});
+let grantRoundMetadata = ref<Record<string, GrantRoundMetadataResolution>>({});
```
mds1, line 40: Same comment about using `const`

app/src/store/data.ts
```diff
@@ -162,28 +165,46 @@ export default function useDataStore() {
...
 
-    const metaPtrs = grants.value.map((grant) => grant.metaPtr);
-    // Initialize metadata for metaPtrs we haven't encountered yet to status "pending"
+    // Fetch Grants Metadata
+    const grantMetaPtrs = grants.value.map((grant) => grant.metaPtr);
+    grantMetadata = await fetchMetaPtrs(grantMetaPtrs, grantMetadata);
+
+    // Fetch GrantRounds Metadata
+    const grantRoundMetaPtrs = grantRounds.value.map((grantRound) => grantRound.metaPtr);
+    grantRoundMetadata = await fetchMetaPtrs(grantRoundMetaPtrs, grantRoundMetadata);
```
mds1, line 174: Ahh, I see why you used `const` instead of `let`. I think it's a misunderstanding of how refs work. Check out the [docs](https://v3.vuejs.org/api/refs-api.html), but basically when you defined a ref, you always make changes to it's `.value` property, not the object itself. 

The current approach is overriding the full ref, and will break the reactivity. Fixing that would give you code something like this:

```suggestion
    // Fetch Grants Metadata
    const grantMetaPtrs = grants.value.map((grant) => grant.metaPtr);
    grantMetadata.value = await fetchMetaPtrs(grantMetaPtrs, grantMetadata);

    // Fetch GrantRounds Metadata
    const grantRoundMetaPtrs = grantRounds.value.map((grantRound) => grantRound.metaPtr);
    grantRoundMetadata.value = await fetchMetaPtrs(grantRoundMetaPtrs, grantRoundMetadata);
```

But, that's not exactly what we want either, as partly explained [here](https://github.com/dcgtc/dgrants/pull/86/#discussion_r691600963). The current approach has two issues:
1. `grantRoundMetadata` is blocked on waiting for `grantMetadata` to resolve
2. Individual metadata pointers may resolve slowly, and the current approach blocks the UI until all metadata is resolved

This is an easy fix for you though, basically just change the snippet to

```suggestion
    // Fetch metadata
    const grantMetaPtrs = grants.value.map((grant) => grant.metaPtr);
    const grantRoundMetaPtrs = grantRounds.value.map((grantRound) => grantRound.metaPtr);
    void fetchMetaPtrs(grantMetaPtrs, grantMetadata);
    void fetchMetaPtrs(grantRoundMetaPtrs, grantRoundMetadata);
```

app/src/views/GrantRoundDetails.vue
```diff
@@ -144,11 +151,12 @@ import {
   hasStatus,
 } from 'src/utils/utils';
 // --- Types ---
-import { GrantRound } from '@dgrants/types';
+import { Grant, GrantRound } from '@dgrants/types';
 import { GrantRound as GrantRoundContract } from '@dgrants/contracts';
 
 function useGrantRoundDetail() {
-  const { grantRounds, poll } = useDataStore();
+  const { grantRounds, grantRoundMetadata: metadata, grants: allGrants, grantMetadata, poll } = useDataStore();
```
mds1, line 158: Not a fan of the variable name `metadata` since it's not clear whether it's for a round or a grant. Perhaps just stick with the original name?

mds1: Ah I see, you redefine `grantRoundMetadata` as a computed property. Maybe `_grantRoundMetadata` then.

But I don't think this and `grants` need to be computed properties. I think you can just set them on mount with `onMounted`, because (I believe) each time you change the URL param to a round the component is destroyed the mounted

app/src/store/data.ts
mds1: Also maybe just add a few more comments around the `fetchMetaPtrs`, because I was confused at first since it was mutating state _and_ returning something. But we only need it to mutate state, and the state being mutated is the ref passed as the `metadata` input. Worth detailing that + specifying the second input should be a `ref`

thelostone-mc: mind need a hand in figuring out how to word it 😅

app/src/views/GrantRoundDetails.vue
thelostone-mc: I've renamed it to `_grantRoundMetadata` for now ! Not sure how do it via `onMounted` 
Any chance you could show me how to go about that ?

app/src/store/data.ts
mds1: Sample below:

```typescript
/**
 * @notice Helper method that fetches metadata for a Grant or GrantRound, and saves the data
 * to the state as soon as it's received
 * @param metaPtrs Array of URLs to resolve
 * @param metadata Name of the store's ref to assign resolve metadata to
 */
async function fetchMetaPtrs(metaPtrs: string[], metadata: Ref) {
  const newMetadata = metaPtrs
    .filter((metaPtr) => !metadata.value[metaPtr])
    .reduce((prev, cur) => {
      return { ...prev, [cur]: { status: 'pending' } };
    }, {});
  // save these pending metadata objects to state
  metadata.value = { ...metadata.value, ...newMetadata };
  // resolve metadata via metaPtr and update state
  Object.keys(newMetadata).map(async (url) => {
    try {
      const data = await resolveMetaPtr(url);
      metadata.value[url] = { status: 'resolved', ...data };
    } catch (e) {
      metadata.value[url] = { status: 'error' };
      console.error(e);
    }
  });
}
```

app/src/views/GrantRoundDetails.vue
mds1: @thelostone-mc and I realized computed makes more sense here, so marking this resolved

mds1 (approved):

PR #131, opened 2021-08-23
ui: Adds fonts, colours and general CSS stylings

gdixon:
This PR adds the basic fonts, colours and stylings as outlined in [richard-ui3](https://github.com/dcgtc/dgrants/tree/richard-ui3)

---

Closes: #116


app/tailwind.config.js
```diff
@@ -5,11 +5,60 @@ module.exports = {
...
+      sans: ['"Source Code Pro"', 'monospace'],
+    },
+    fontSize: {
+      sm: ['100%', '170%'],
+      md: ['110%', '170%'],
+      lg: ['120%', '170%'],
+      xl: ['130%', '170%'],
+      xxl: ['140%', '170%'],
+      headline: ['150%', '170%'],
+    },
```
mds1, line 18: Everything in the app feels really big. The font sizes, button sizes, etc. feel like I've zoomed in on the page. Is this correct/intentional/final?

app/tailwind.config.js
```diff
@@ -5,11 +5,60 @@ module.exports = {
   darkMode: 'class',
```
mds1, line 5: Dark mode seems to be broken, as it only darkens the button text. I think this is because if you rely on default colors Nightwind doesn't convert those. For example you need to explicitly specify the background color as `bg-white` if you want it to become black (though you probably want use Nightwind overrides to make that a dark gray anyway).

Or maybe we're ok with removing dark mode? In which case we can remove the footer buttons and relevant logic from the settings store

app/tailwind.config.js
```diff
@@ -5,11 +5,60 @@ module.exports = {
...
+      4: '4',
+    },
+    boxShadow: {
+      light: '0px 0px 40px 0px rgba(14, 3, 51, .1)',
+      none: 'none',
+    },
+    cursor: {
+      auto: 'auto',
+      default: 'default',
+      pointer: 'pointer',
```
mds1, line 54: Some of the `cursor` and `colors` defined in this config are equivalent to the Tailwind defaults. Are we intentionally redefining them here for explicitness, or should they be removed for brevity / clarity over what exactly is customized/changed from the defaults?

app/index.html
```diff
@@ -5,6 +5,12 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
     <meta name="viewport" content="width=device-width,initial-scale=1.0" />
     <link rel="icon" href="/favicon.ico" />
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Source+Code+Pro:ital,wght@0,400;0,500;1,400&display=swap"
+      rel="stylesheet"
+    />
```
mds1, line 13: Do we need `preconnect` or `crossorigin` on all of these? I'm not sure offhand, so just checking that we aren't missing any attributes here, since some tags have different attributes than others

mds1 (changes requested):
Looking great overall! Aside from the comments below, one other thing is that some alignments seem to be off now, such as the cart page. This is probably ok since we're still styling things, but I only mention it to make sure we didn't break any of the flexbox classes or something

image

app/package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "@dgrants/contracts": "^0.0.1",
     "@dgrants/types": "^0.0.1",
+    "@fusion-icons/vue": "^0.0.0",
```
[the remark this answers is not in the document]
carletex: Adding one more question: Do we still need `@heroicons/vue` or are fusion-icons supposed to be a replacement?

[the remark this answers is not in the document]
mds1: Ah good question. It seems `@heriocons/vue` is still used in a lot of the components. @gdixon if it's not a big lift to replace the heroicons usage with fusion-icons, it'd be awesome to slip that into this PR also

app/src/index.css
```diff
@@ -8,36 +8,105 @@
...
+    @apply focus:outline-none focus:border-grey-500 focus:ring-0;
+    font-size: 100%;
   }
 }
 
 /* Add styles to components */
 @layer components {
+  .indent {
+    text-indent: 1em;
+  }
```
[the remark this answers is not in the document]
scco: > What is this class for? Indenting paragraphs? Maybe some inline comments about when to use various classes would help clear this up

indenting paragraphs is a typographical classic thing for text - e.g. a project description - see in figma on the grant detail page.

other iconpacks : definitly no need. if we miss a custom icon, i will just add it to fusion.


breaking what ever flexbox classes from not-my-code css/html of the page seems to be ok for me - as it will be replaced anyway with my code .  the code fragments for the cart are ready to use .

getting fusion to a higher version number - could graham do as he pushed it to npm for me. 

writing from vacation :)

app/tailwind.config.js
scco: maybe your "base font size" is setup weired.  the idea is to us the browsers base-font-size of 12 - 16px ....  and from here just add % of that to various breakpoints.  so for mobile  100% of base font size .  for xxl 140% .    worked nice on all my test devices so far.  

it maybe helps to set the base-fontsize on the html element to 12-16px .

app/tailwind.config.js
scco: yes i asked phil some weeks ago, if we can get rid of darkmode and he agreed.

app/tailwind.config.js
scco: i just added a specific cursor (  for the dropzone ) .  thats how tailwind told to do - i assume an extend could also work but unsure ( cant test from vacation )

app/src/index.css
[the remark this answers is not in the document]
mds1: Thanks @scco! Now get back to enjoying vacation 😛

app/tailwind.config.js
mds1: Hm, yea we might want to set base fontsize. I don't know if things looked really big for anyone else, but I've never changed my browser font settings 🤷

cc @gdixon

app/src/index.css
```diff
@@ -8,36 +8,105 @@
...
+  .btn {
+    @apply flex items-center gap-x-2 whitespace-nowrap uppercase font-medium bg-grey-500 text-white px-8 py-4 hover:bg-grey-400;
   }
-}
 
-/* Add styles to utilities */
-@layer utilities {
+  .btn svg {
+    @apply stroke-3 stroke-white object-contain h-6 w-6 md:h-8 md:w-8;
+  }
```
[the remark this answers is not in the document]
scco: to style the stroke size and stroke color of a buttons optional inner icon ( and its icon dimension )

graham:
I believe you have the wrong @graham :)

app/tailwind.config.js
gdixon: Wouldn't adjusting the defined font-sizes make more sense than adjusting the base-font-size? If we set a base font then we lose the ability for our user to define their own?

mds1: I'm not actually sure which approach is considered best practice. Adjusting the font size percentages here feels like the way to go, but I thought maybe it wasn't per Richard's comment

scco: if a user tweaks its own base font size (what less then 0,1% of users do) , they also know how to use cmd+ and cmd- . can not see any issue but wana realy see screenshot what mds1 defines as "to big" as maybe its not even a bug but intentional ...

PR #143, opened 2021-08-25
feat: integrate trust-bonus

thelostone-mc:
#### Description 

Implements the logic documented by @frankchen07  to fetch the trust bonus score from Gitcoin API and integrate it into linear CLR calculation

- adds axios as a dependency 
- integrates trust bonus score into linear algorithm 

The dcurve has 3 functions 

- predict  (used on frontend)
  - can fetch trust bonus from Gitcoin /
  - can be passed as an argument (this is to avoid multiple API calls)
- calculate (used at end of round)
  - fetch trust bonus from Gitcoin
  - uploads trust bonus to IPFS and returns metaPtr
- verify (used to verify)
  - inputs: trust bonus IPFS metaPtr , hash to verify , distribution 
  - lets verifier know if hash / merkle root matches or not

Trust bonus uploaded to : http://cloudflare-ipfs.com/ipfs/Qmayu6Sur7zhUMzMp2SQPyvSenXLw1e4iVy6hyzKrPpREr
image

### DIFF
--- a/dcurve/README.md
+++ b/dcurve/README.md
@@ -108,7 +108,7 @@ const grantRoundContributions = fetch(fetchArgs);
 
 // 4. Calculate Distribution
 
-const distribution = clr.calculate(grantRoundContributions);
+const distribution = await clr.calculate(grantRoundContributions);
 
 // 5. Predict match for a grant
 
@@ -117,5 +117,5 @@ const grantPredictionArgs = {
   predictionPoints: [1, 10, 100],
   grantRoundContributions: contributions
 };
-const prediction = clr.predict(predictArgs);
+const prediction = await clr.predict(predictArgs);
 ```
--- a/dcurve/README.md
+++ b/dcurve/README.md
@@ -85,15 +85,13 @@ This file orchestrates the calculation and prediction procedure, exposing two me
 import { fetch, linear, CLR } from @dgrants/dcurve;
 
 // 2. Create instance of CLR
-
 const initArgs = {
   calcAlgo: linear
 };
 const clr = new CLR(options);
 
 // 3. Fetch contributions
-
-const grantRoundFetchArgs = {
+const fetchArgs = {
   provider: provider,
   grantRound: GRANT_ROUND_ADDRESS,
   grantRoundManager: GRANT_ROUND_MANAGER_ADDRESS,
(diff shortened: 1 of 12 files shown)


apbendi:
Hey @thelostone-mc, I'm wondering if the dcurve script should also leave some kind of log or reporting of the TrustBonus scores it uses, since we're going to rely on this centralizing component in the MVP, we can at least "show our work" in the output for transparency.

dcurve/package.json
```diff
@@ -31,6 +31,7 @@
   "dependencies": {
     "@dgrants/contracts": "^0.0.1",
     "@dgrants/utils": "^0.0.1",
+    "axios": "^0.21.1",
```
mds1, line 34: IMO we should use the built-in [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) API instead of axios, unless we need some of the fancier axios features (which it seems we don't), because:
- fewer dependencies -> smaller bundle -> better performance
- fewer dependencies -> less attack surface

dcurve/README.md
```diff
@@ -108,7 +108,7 @@ const grantRoundContributions = fetch(fetchArgs);
 
 // 4. Calculate Distribution
 
-const distribution = clr.calculate(grantRoundContributions);
+const distribution = await clr.calculate(grantRoundContributions);
```
mds1, line 111: Instead of fetching the trust bonus within these calls, does it make sense to make trust bonus part of the input to keep things a little more modular? That way if/when we change the source of the trust bonus, it's less coupled and easier to change. For example something like this:

```typescript
export type Contribution = {
  grantId: string;
  grantAddress: string;
  address: string;
  amount: number;
  trustBonus: number; // add this as part of Contribution objects
};

// step 1
// get trust bonus scores and assign them to your grantRoundContribution objects

// step 2
const distribution = await clr.calculate(grantRoundContributions);
```

dcurve/src/internal/calc/linear.ts
```diff
@@ -25,32 +33,44 @@ export const handle = (clrArgs: CLRArgs): GrantsDistribution => {
         grantAddress: contribution.grantAddress,
         contributions: [],
       };
+
+      contributionAddresses.push(contribution.address);
```
mds1, line 37: Nit: could save a line and make intent more clear. Instead of array-> set -> array, could just do

```typescript
const contributionAddresses: Set<string> = new Set();
// ...
contributionAddresses.add(contribution.address)
// ...
const trustBonusScores = await fetchTrustBonusScore([...contributionAddresses]);

File: dcurve/src/internal/utils.ts [hunk 4]
```diff
@@ -47,3 +48,26 @@ export const getGrantMatch = (grantId: string, grantsDistribution: GrantsDistrib
...
+  const params = {
+    addresses: addresses.join(','),
+  };
+
+  try {
+    const response = await axios.get(url, { params: params });
+    return response.data;
+  } catch (err) {
+    console.error('fetchTrustBonusScore:', err);
+    return [];
```
mds1, line 71: Is silently failing the right move here? Or should we throw if the response status is not 200?

phutchins (approved):
All looks good.

Issue #145, opened 2021-08-25
Expand Grant Metadata to Include Basic Properties


apbendi:
Currently, the minimal grant metadata schema is limited to:

```json
{
    "name": "My Grant",
    "description": "There are many Grants, but this one is mine",
    "logo": "https://cloudflare-ipfs.com/ipfs/QmaKj8z3icu3UFqXK4NXy7Mcch9kCPZJA128gP4pa4fQQM"
}
```

We want to expand the schema to include some of the basic grant properties, most importantly properties that will point externally to other sources of information about the grant, namely:

- [x] The project website
- [x] Twitter Handle
- [x] Github

The updated schema should look like:

```json
{
    "name": "My Grant",
    "description": "There are many Grants, but this one is mine",
    "logo": "https://cloudflare-ipfs.com/ipfs/QmaKj8z3icu3UFqXK4NXy7Mcch9kCPZJA128gP4pa4fQQM",
    "properties": {
        "projectWebsite": "https://gitcoin.co",
        "projectGithub": "https://github.com/dcgtc/dgrants",
        "twitterHandle": "https://twitter.com/GitcoinDAO"
     }
}
```

The frontend should also be updated to accommodate these properties:

- [x] Optional fields on the create/edit grant form with basic validation, including URL for the GitHub & Twitter links
- [x] Properly add the fields to the metadata before uploading it to IPFS
    - [x] Each individual property is optional, but the `properties` top level key should always be present. The app should handle the absence of any individual property gracefully.
- [x] Display these fields on the Grant Detail page

mds1:
Right now we have `logoURI` as optional also, so I just want to clarify which items in the metadata schema are optional vs. required. You could argue `name` is the only required field, but maybe we want to require other fields also? I only mention it here because this feels like a good issue to clarify / update that in

PR #157, opened 2021-08-27
feat: expand grant metadata to include basic properties

mk1r:
resolves issue #145

-  Optional fields on the create/edit grant form with basic validation, including URL for the GitHub & Twitter links
-  Properly add the fields to the metadata before uploading it to IPFS
-  Each individual property is optional, but the properties top level key should always be present. The app should handle the absence of any individual property gracefully.
- Display these properties on the Grant Round details page & Grant Registry details page

### DIFF
--- a/app/src/components/BaseInput.vue
+++ b/app/src/components/BaseInput.vue
@@ -23,7 +23,7 @@
         ]"
         :id="id"
         :name="id"
-        :required="true"
+        :required="required"
         :type="type"
         :readonly="readonly"
         :disabled="disabled"
@@ -53,6 +53,7 @@ export default defineComponent({
     id: { type: String, required: false, default: undefined }, // id, for accessibility
     placeholder: { type: String, required: false, default: undefined }, // input placeholder text
     type: { type: String, required: false, default: 'text' }, // input type
+    required: { type: Boolean, required: false, default: true }, // is required
     readonly: { type: Boolean, required: false, default: false }, // is readonly
     disabled: { type: Boolean, required: false, default: false }, // is disabled
     width: { type: String, required: false, default: 'w-full' }, // input field width
--- a/app/src/views/GrantRegistryGrantDetail.vue
+++ b/app/src/views/GrantRegistryGrantDetail.vue
@@ -191,7 +191,7 @@ import {
   SUPPORTED_TOKENS_MAPPING,
 } from 'src/utils/constants';
 import { Contract, ContractTransaction, formatUnits } from 'src/utils/ethers';
-import { isValidAddress, isValidUrl, isDefined, formatNumber, urlFromTwitterHandle } from 'src/utils/utils';
+import { isValidAddress, isValidUrl, isValidGithubUrl, isDefined, formatNumber, urlFromTwitterHandle } from 'src/utils/utils';
 import { GrantRegistry } from '@dgrants/contracts';
 import { hexlify } from 'ethers/lib/utils';
 import * as ipfs from 'src/utils/ipfs';
--- a/app/src/views/GrantRegistryGrantDetail.vue
+++ b/app/src/views/GrantRegistryGrantDetail.vue
@@ -150,6 +150,8 @@
             description="Your grant's twitter handle"
             id="grant-handle"
             label="Grant twitter"
+            :rules="isValidTwitter"
+            errorMsg="Please enter a valid Twitter handle"
             :required="false"
           />
 
@@ -198,6 +200,7 @@ import {
   isValidAddress,
   isValidUrl,
   isValidGithubUrl,
+  isValidTwitter,
   isDefined,
   formatNumber,
   urlFromTwitterHandle,
@@ -530,6 +533,7 @@ function useGrantDetail() {
     isValidAddress,
     isValidUrl,
     isValidGithubUrl,
+    isValidTwitter,
     isDefined,
     isFormValid,
     grant,
--- a/app/src/views/GrantRegistryGrantDetail.vue
+++ b/app/src/views/GrantRegistryGrantDetail.vue
@@ -31,11 +31,8 @@
 
     <SectionHeader title="Links" />
 
-    <div
-      v-if="isDefined(grantMetadata?.properties?.websiteURI)"
-      class="px-4 md:px-12 py-8 border-b border-grey-100 flex flex-col gap-y-4"
-    >
-      <div class="flex gap-x-4">
+    <div class="px-4 md:px-12 py-8 border-b border-grey-100 flex flex-col gap-y-4">
+      <div v-if="isDefined(grantMetadata?.properties?.websiteURI)" class="flex gap-x-4">
         <span class="text-grey-400">Website:</span>
         <a :href="grantMetadata?.properties?.websiteURI" target="_blank">{{ grantMetadata?.properties?.websiteURI }}</a>
       </div>
--- a/app/src/views/GrantRegistryGrantDetail.vue
+++ b/app/src/views/GrantRegistryGrantDetail.vue
@@ -143,12 +143,10 @@
 
           <!-- Grant twitter handle -->
           <BaseInput
-            v-model="form.handle"
+            v-model="form.twitter"
             description="Your grant's twitter handle"
             id="grant-handle"
             label="Grant twitter"
-            :rules="isValidUrl"
-            errorMsg="Please enter a valid URL"
             :required="false"
           />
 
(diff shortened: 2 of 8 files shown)


app/src/components/BaseInput.vue
```diff
@@ -24,7 +24,7 @@
         "
         :id="id"
         :name="id"
-        :required="true"
+        :required="required"
```
thelostone-mc, line 27: oh nice one ! lol i wonder why we defaulted it to false

thelostone-mc (approved):
codewise LGTM

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -72,6 +72,39 @@
...
+            :required="false"
+          />
+
+          <!-- Grant github -->
+          <BaseInput
+            v-model="form.github"
+            description="Your grant's github"
+            id="grant-github"
+            label="Grant github"
+            :rules="isValidUrl"
```
apbendi, line 92: In addition to validating that this is a URL, I think we should validate specifically that it's a *GitHub* URL. Counter argument here would be that some projects might use GitLab or something else.

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -72,6 +72,39 @@
...
+            :required="false"
+          />
+
+          <!-- Grant twitter handle -->
+          <BaseInput
+            v-model="form.handle"
+            description="Your grant's twitter handle"
+            id="grant-handle"
+            label="Grant twitter"
+            :rules="isValidUrl"
```
apbendi, line 103: While we settled on *storing* a twitter URL in the metadata, I don't think the form should *take* a URL from the user. Instead, I think we should ask them to input their handle— which we would accept with or without the "@" sign— and then normalize it and append it to `https://twitter.com/` to form the expected URL.

apbendi (changes requested):
Hey @mk1r, nice work! Working seamlessly when I create and load new grants.

In addition to the feedback I left on validating GitHub/Twitter URLs, I think we should also update the dev environment fixtures. Preferably, the metadata for the fixtures would include a few different cases (1 with all of these fields, others with only 1 or 2 of them). Would you have any issue with making that update?

app/src/views/GrantRegistryGrantDetail.vue
mk1r: should this validation be done inside GrantRegistryGrantDetail.vue / GrantRegistryNewGrant.vue or in another file (like utils)?

app/src/views/GrantRegistryGrantDetail.vue
mds1: Hmm, I think we should leave this as validating a generic URL. Because there's also Bitbucket, Radicle, and probably others. And Radicle is a decentralized git so there may not be a single 2LD we can look for to cover all Radicle repos

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -118,27 +154,61 @@ function useGrantDetail() {
...
 
-  const form = ref<{ owner: string; payee: string; name: string; description: string }>({
+  const form = ref<{
+    owner: string;
+    payee: string;
+    name: string;
+    description: string;
+    website: string;
+    github: string;
+    handle: string;
```
mds1, line 188: Nit: Maybe `twitter` or `twitterHandle` instead to be more explicit? I can imagine adding e.g. a telegram handle field down the road

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -118,27 +154,61 @@ function useGrantDetail() {
...
+      ? metadata.value[grant.value.metaPtr]?.properties?.projectGithub !== undefined &&
+        metadata.value[grant.value.metaPtr]?.properties?.projectGithub !== ''
+      : false
+  );
+  const hasHandle = computed(() =>
+    grant.value
+      ? metadata.value[grant.value.metaPtr]?.properties?.twitterHandle !== undefined &&
+        metadata.value[grant.value.metaPtr]?.properties?.twitterHandle !== ''
+      : false
+  );
```
mds1, line 175: These all follow the same structure so you might be able to make things a bit more DRY with a helper method that takes the field name, e.g. `projectWebsite` as an input. Maybe that's clunky/harder to read though. I'm fine with either

app/src/views/GrantRoundDetails.vue
```diff
@@ -184,6 +187,26 @@ function useGrantRoundDetail() {
...
+      ? _grantRoundMetadata.value[grantRound.value.metaPtr]?.properties?.projectGithub !== undefined &&
+        _grantRoundMetadata.value[grantRound.value.metaPtr]?.properties?.projectGithub !== ''
+      : false
+  );
+  const hasHandle = computed(() =>
+    grantRound.value
+      ? _grantRoundMetadata.value[grantRound.value.metaPtr]?.properties?.twitterHandle !== undefined &&
+        _grantRoundMetadata.value[grantRound.value.metaPtr]?.properties?.twitterHandle !== ''
+      : false
+  );
```
mds1, line 208: Since these definitions are duplicated again here, they should be moved to be computed properties of the `data.ts` store instead

types/src/grantRounds.d.ts
```diff
@@ -25,6 +25,11 @@ export type GrantRoundMetadata = {
   name: string;
   description: string;
   grants: BigNumberish[];
+  properties: {
+    projectWebsite: string;
+    projectGithub: string;
+    twitterHandle: string;
```
mds1, line 31: Our naming convention feels misleading IMO, e.g. using a name "_project_ Website" for the _round_ metadata is confusing. I think we should either:
1. Use more generic names that don't include the words "round" or "project". For example `websiteURI` and `twitterURI` would be the field names for both GrantRoundMetadata and GrantMetadata. This is my preference, since it's also consistent with `logoURI`. I mentioned this [here](https://github.com/dcgtc/dgrants/pull/158#discussion_r698613804), so we'll need to coordinate with @gdixon if we change the names
2. Rename these to reflect they are for rounds, not projects, e.g. `roundWebsite` instead of `projectWebsite`

app/src/views/GrantRegistryGrantDetail.vue
mds1: Probably utils since you'll need to do the normalization both on grant creation and when editing

types/src/grantRounds.d.ts
phutchins: I lean more toward the first option where we have the elements (websiteURI, twitterURI) under each type (GrantRoundMetadata, GrantMetadata) have consistent names. We already have the context from the object that owns the element.

mds1 (approved):
Reviewed the code and LGTM, thank you @mk1r!

apbendi:
Looking good @mk1r! A couple things I noticed

* Clicking the link on the Grant Detail page does not actually take you to the URL
* The fields are missing on the Edit form

Screen Shot 2021-09-03 at 9 55 15 AM


* The twitter field is not validated, i.e. it should be limited to [valid twitter handles](https://kagan.mactane.org/blog/2009/09/22/what-characters-are-allowed-in-twitter-usernames/comment-page-1/), with or without the "@" symbol prepending it, and should handle it gracefully. For example, when I mistakenly entered the full URL, here's what it looks like:

Screen Shot 2021-09-03 at 9 53 48 AM

mk1r:
@apbendi 
1) which .vue page are you referring to? so I can check
2) it's normalizing the URL (if entered with @) - do we want to restrict URLs being entered there?

apbendi:
@mk1r

1. On the GrantDetail Page, i.e. `/dgrants/2` or any other grant id
2. The form should only accept input that could possibly be a twitter handle, which apparently means letters, digits, and underscores, up to 15 chars. All good on normalizing the @ part.

mk1r:
@apbendi 

sounds good - will add in a rule to restrict

mine looks like this 
Screen Shot 2021-09-03 at 9 21 43 AM

and also, I was asking about the edit form

apbendi:
@mk1r sorry what part of the Edit form were you asking about?

mk1r:
@apbendi just want to see it in the UI and find the corresponding .vue page in the code - but can't find how to edit in the new UI

mk1r:
hey @apbendi take a look when you get a chance

apbendi:
Hey @mk1r, just tried it out. Links work now, which is awesome.

On the twitter handle stuff:

1. It still let me create a grant with a bad twitter handle (I tried `UmbraCash!` which has an illegal character)
2. Even when I used a valid twitter handle, it did NOT show the twitter link on the actual page (I did not fill any other links)

apbendi (approved):
Looking great @mk1r. Nice work 🎉 👏

mk1r:
@apbendi thanks for stress testing

Issue #141, opened 2021-08-25
Show Donation Information on Grant & Grant Details


apbendi:
- [x] Number of contributions
- [x] Value Amount of contributions denominated in the donation token (pulled from [the event](https://github.com/dcgtc/dgrants/blob/main/contracts/contracts/GrantRoundManager.sol#L216))
- [x] List of itemized contributions
    - [x] Amount
    - [x] Donor (address / identicon)

Issue #135, opened 2021-08-23
UI Conversion: Grant Detail Page


apbendi:
#### Description

This task involves building up the GrantDetail Page

**Image:** [HERE](https://user-images.githubusercontent.com/5358146/130651432-21876194-68fe-4605-aadc-53abadaaf187.png)
 **Figma:** https://www.figma.com/file/sUQSYyPfooqTcVdajfZm0t/dgrants?node-id=350%3A2695
 
 Below is the checklist of components which needs to be built out to complete this page 
 
-  [x] **component**: Breadcrumbs
- [ ] **component**: BaseHeader 
  - [x] Title
  - [x] Grant Owner
  - [ ] Last Update
- [x] **component**: GrantDetailsRow
  - [x] Grant Logo (fetched from metaPtr )
  - [x] GrantDetailsRow 
  - [x] ActionsNav
 - [x] Description (fetched from metaPtr)
 - [ ] Links (fetched from metaPtr)
   - [ ] Github 
   - [ ] Website
   - [ ] Figma 
   - [ ] Twitter
 - [x] **component**: ContributionRow

PR #158, opened 2021-08-27
ui: Implements Grant Details page design

gdixon:
This PR Implements the Grant Details page design:

**Image:** [HERE](https://user-images.githubusercontent.com/5358146/130651432-21876194-68fe-4605-aadc-53abadaaf187.png)
 **Figma:** https://www.figma.com/file/sUQSYyPfooqTcVdajfZm0t/dgrants?node-id=350%3A2695
 
---
- [x] Number of contributions
- [x] Value Amount of contributions denominated in the donation token (pulled from [the event](https://github.com/dcgtc/dgrants/blob/main/contracts/contracts/GrantRoundManager.sol#L216))
- [x] List of itemized contributions
    - [x] Amount
    - [x] Donor (address / identicon)
---

- [x] **component**: Breadcrumbs
- [x] **component**: BaseHeader 
  - [x] Title
  - [x] Grant Owner
  - [x] Last Update
- [x] **component**: GrantDetailsRow
  - [x] Grant Logo (fetched from metaPtr )
  - [x] GrantDetailsRow 
  - [x] ActionsNav
 - [x] Description (fetched from metaPtr)
 - [x] Links (fetched from metaPtr)
   - [x] Github 
   - [x] Website
   - [x] Figma 
   - [x] Twitter
 - [x] **component**: ContributionRow
  
  (All of the touch points have been built into the page - but the metadata insertion is unchanged)
  
---

Fixes: #141 
Fixes: #135

---

localhost_3000_dgrants_0

### DIFF
--- a/types/src/app.d.ts
+++ b/types/src/app.d.ts
@@ -5,10 +5,10 @@
 export type RouteTarget = {
   name: string; // the routes name we're navigating to
   params?: {
-    [key: string]: string|number // any params we want to feed into the route
+    [key: string]: string | number; // any params we want to feed into the route
   };
   query?: {
-    [key: string]: string|number // any query string elements we want to feed into the route
+    [key: string]: string | number; // any query string elements we want to feed into the route
   };
 };
 
@@ -36,4 +36,4 @@ export type FilterNavItem = {
 export type FilterNavButton = {
   label: string; // label to display in the button
   action?: (payload: MouseEvent) => void; // action to take when we click the button
-}
+};
--- a/types/src/grants.d.ts
+++ b/types/src/grants.d.ts
@@ -19,11 +19,13 @@ export type GrantMetadata = {
   name: string;
   description: string;
   logoURI?: string;
-  quote?: string;
-  lastUpdated?: string;
-  projectWebsite?: string;
-  projectGithub?: string;
-  twitterHandle?: string;
+  properties?: {
+    websiteURI?: string;
+    githubURI?: string;
+    twitterURI?: string;
+  } & {
+    [key: string]: string; // for extensions/flexibility
+  };
 };
 export type GrantMetadataStatus = 'resolved' | 'pending' | 'error';
 export type GrantMetadataResolution = Partial<GrantMetadata> & {
--- a/app/src/components/Breadcrumb.vue
+++ b/app/src/components/Breadcrumb.vue
@@ -23,7 +23,7 @@ import { Breadcrumb } from '@dgrants/types';
 import { ArrowRight2Icon as ArrowRightSmallIcon } from '@fusion-icons/vue/interface';
 
 export default defineComponent({
-  name: 'BaseHeader',
+  name: 'Breadcrumb',
   props: {
     path: { type: Array as PropType<Breadcrumb[]>, required: true },
   },
--- a/app/src/components/SectionHeader.vue
+++ b/app/src/components/SectionHeader.vue
@@ -4,7 +4,6 @@
       <div class="flex flex-wrap content-center gap-x-8">
         <div class="cursor-pointer uppercase font-medium">
           {{ title }}
-          <span v-if="counter" class="ml-2 text-grey-400">({{ counter }})</span>
         </div>
       </div>
     </div>
@@ -15,10 +14,9 @@
 import { defineComponent } from 'vue';
 
 export default defineComponent({
-  name: 'BaseTitle',
+  name: 'SectionHeader',
   props: {
     title: { type: String, required: false, default: undefined },
-    counter: { type: Number, required: false, default: undefined },
   },
 });
 </script>
--- a/app/src/views/GrantRegistryGrantDetail.vue
+++ b/app/src/views/GrantRegistryGrantDetail.vue
@@ -21,7 +21,7 @@
       </template>
     </GrantDetailsRow>
 
-    <BaseTitle title="Description" />
+    <SectionHeader title="Description" />
 
     <div class="border-b border-grey-100">
       <p class="intent px-4 md:px-12 py-12 mx-auto max-w-6xl">
@@ -29,22 +29,22 @@
       </p>
     </div>
 
-    <BaseTitle title="Links" />
+    <SectionHeader title="Links" />
 
     <div class="px-4 md:px-12 py-8 border-b border-grey-100 flex flex-col gap-y-4">
       <div class="flex gap-x-4">
         <span class="text-grey-400">Website:</span>
-        <a href="#" target="_blank">{{ grantMetadata?.projectWebsite || 'undefined' }}</a>
+        <a href="#" target="_blank">{{ grantMetadata?.properties?.websiteURI || 'undefined' }}</a>
       </div>
 
       <div class="flex gap-x-4">
         <span class="text-grey-400">Github:</span>
-        <a href="#" target="_blank">{{ grantMetadata?.projectGithub || 'undefined' }}</a>
+        <a href="#" target="_blank">{{ grantMetadata?.properties?.githubURI || 'undefined' }}</a>
       </div>
 
       <div class="flex gap-x-4">
         <span class="text-grey-400">Twitter:</span>
-        <a href="#" target="_blank">{{ grantMetadata?.twitterHandle || 'undefined' }}</a>
+        <a href="#" target="_blank">{{ grantMetadata?.properties?.twitterURI || 'undefined' }}</a>
       </div>
     </div>
 
@@ -169,7 +169,7 @@ import { Breadcrumb, FilterNavItem, GrantRound, GrantsRoundDetails } from '@dgra
 // --- Components ---
 import BaseInput from 'src/components/BaseInput.vue';
 import BaseHeader from 'src/components/BaseHeader.vue';
-import BaseTitle from 'src/components/BaseTitle.vue';
+import SectionHeader from 'src/components/SectionHeader.vue';
 import BaseFilterNav from 'src/components/BaseFilterNav.vue';
 import ContributionRow from 'src/components/ContributionRow.vue';
 import GrantDetailsRow from 'src/components/GrantDetailsRow.vue';
@@ -485,7 +485,7 @@ function useGrantDetail() {
 
 export default defineComponent({
   name: 'GrantRegistryGrantDetail',
-  components: { BaseInput, BaseTitle, ContributionRow, BaseHeader, BaseFilterNav, GrantDetailsRow },
+  components: { BaseInput, SectionHeader, ContributionRow, BaseHeader, BaseFilterNav, GrantDetailsRow },
   setup() {
     const { addToCart, isInCart, removeFromCart } = useCartStore();
 


types/src/grants.d.ts
```diff
@@ -18,6 +19,25 @@ export type GrantMetadata = {
...
+// Details of a grant in a round
+export type GrantsRoundDetails = {
+  address: string;
+  metaPtr: string;
+  name: string;
+  matchingToken: TokenInfo;
+  donationToken: TokenInfo;
+  contributions: (false | Event)[];
+  matching: number;
+  prediction10: number;
```
corydickson, line 40: for 0th prediction point, we assume it to be the initial matching balance correct?

types/src/app.d.ts
```diff
@@ -0,0 +1,20 @@
...
+export type FilterItem = {
+  title?: string;
+  counter?: number;
+  tag?: string;
+  active?: number;
+  action?: (payload: MouseEvent) => void;
+  menu?: {
+    title?: string;
+    action?: (payload: MouseEvent) => void;
+    seperator?: boolean;
```
mds1, line 18: Typo

```suggestion
    separator?: boolean;
```

types/src/app.d.ts
```diff
@@ -0,0 +1,20 @@
+// --- App ---
+// Any types which are shared between different parts of the App
+
+export type Breadcrumb = {
+  name: string;
+  href: string;
+};
+
+export type FilterItem = {
```
mds1, line 9: Could you add some comments about what each option is? It's not clear to me just from reading this type definition

types/src/grants.d.ts
```diff
@@ -18,6 +19,25 @@ export type GrantMetadata = {
   name: string;
   description: string;
   logoURI?: string;
+  quote?: string;
+  lastUpdated?: string;
+  projectWebsite?: string;
+  projectGithub?: string;
+  twitterHandle?: string;
```
mds1, line 26: Two comments:
1. I thought we decided on a Twitter URL for consistency instead of the handle, but maybe I'm misremembering?
2. Perhaps `websiteURI`, `githubURI` and `twitterURI` for consistency with `logoURI` in naming conventions?

types/src/app.d.ts
```diff
@@ -0,0 +1,20 @@
+// --- App ---
+// Any types which are shared between different parts of the App
+
+export type Breadcrumb = {
+  name: string;
+  href: string;
```
mds1, line 6: A bit of a nit, but to me `href` indicates a full URL, but here we just link to other parts of the app. Also IMO best practice for routing is to route by name, since that's just internal so we can ensure it never changes, whereas the route slug may change. Therefore I think a type like this makes more sense:

```typescript
export type Breadcrumb = {
  displayName: string; // what is shown in the breadcrumb trail
  routeName: string; // path `name` defined in `app/src/router/index.ts`
};

File: types/src/grants.d.ts (hunk 1 above)
Comment (gdixon) [reply types/src/grants.d.ts:40]: Yes thats rights, the 0th prediction is the `matching` value - which is taken from `predictedGrantMatch` rather than the `predictionDiff` 👍

File: app/src/components/BaseHeader.vue [hunk 6]
```diff
@@ -0,0 +1,72 @@
...
+    owner: { type: String, required: false, default: undefined },
+    lastUpdated: { type: String, required: false, default: undefined },
+    quote: { type: String, required: false, default: undefined },
+    nextPath: { type: String, required: false, default: undefined },
+    lastPath: { type: String, required: false, default: undefined },
+  },
+  components: { ArrowLeftIcon, ArrowRightIcon, Breadcrumb },
+  setup(props) {
+    const by = computed(() => (props.owner ? formatAddress(props.owner) : false));
+    const byWhen = computed(() => (props.lastUpdated ? daysAgo(new Date(props.lastUpdated).getTime() / 1000) : false));
```
mds1, line 67: Nit, but maybe `formattedLastUpdated` (or something similar) is a more clear variable name? `byWhen` sounds like a deadline

app/src/components/BaseHeader.vue
```diff
@@ -0,0 +1,72 @@
...
+    <!-- optional quote -->
+    <p v-if="quote" class="intent italic mt-8 mx-auto max-w-6xl">“{{ quote }}”</p>
+
+    <span>
+      <!-- optional navigation to jump to next item in a list -->
+      <router-link
+        :class="['group', 'inline', 'p-6', 'absolute', 'right-0', !nextPath ? 'cursor-default' : '']"
+        style="top: 50%; transform: translateY(-50%)"
+        :to="nextPath"
+      >
```
mds1, line 27: Should these router links have `v-if="nextPath"` (and same for last path) on them? When we use this component on e.g. the Cart page, we don't want the arrows to show up, but it seems they still will be there. Downside is then you can't get the "visible but disable" look we currently use without an extra prop or a special string for `nextPath`

app/src/components/BaseTitle.vue
```diff
@@ -0,0 +1,24 @@
...
+      </div>
+    </div>
+  </div>
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue';
+
+export default defineComponent({
+  name: 'BaseTitle',
```
mds1, line 18: I think this needs a more clear name, or some comments, as I'm unsure what this component is for. From the name it sounds like this should be part of `BaseHeader`, but from the usage that is not the case

app/src/components/BaseTitle.vue
```diff
@@ -0,0 +1,24 @@
...
+  props: {
+    title: { type: String, required: false, default: undefined },
+    counter: { type: Number, required: false, default: undefined },
```
mds1, line 21: What is this prop for? Seems unused

app/src/components/ContributionRow.vue
```diff
@@ -0,0 +1,61 @@
...
+import { defineComponent } from 'vue';
+import Jazzicon from 'src/components/Jazzicon.vue';
+import { formatUnits } from 'src/utils/ethers';
+import { formatAddress, roundNumber } from 'src/utils/utils';
+
+export default defineComponent({
+  name: 'ContributionRow',
+  props: {
+    contribution: { type: Object, required: false, default: () => ({}) },
+    donationToken: { type: Object, required: false, default: () => ({}) },
```
mds1, line 54: Can we use more specific types here? e.g. `Object as PropType<Something>`. Also seems odd that the type is an object, but the default is a function

app/src/components/GrantDetailsRow.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+  <div class="border-b border-grey-100">
+    <div class="grid grid-cols-4 gap-0">
+      <!-- img -->
+      <div class="col-span-4 md:col-span-2 shadow-light flex">
+        <img class="m-auto" :src="logoURI" />
```
mds1, line 6: `logoURI` is optional so we should probably account for that with some default image

app/src/components/GrantDetailsRow.vue
```diff
@@ -0,0 +1,110 @@
...
+import { ArrowToprightIcon as ShareIcon } from '@fusion-icons/vue/interface';
+import { formatAddress } from 'src/utils/utils';
+import useCartStore from 'src/store/cart';
+import { GrantsRoundDetails } from '@dgrants/types';
+
+export default defineComponent({
+  name: 'GrantDetailsRow',
+  components: { CartIcon, ShareIcon },
+  props: {
+    grant: { type: Object, required: false, default: () => ({}) },
```
mds1, line 98: Same comment that we should use a more specific prop type, might be others I missed too

app/src/utils/utils.ts
```diff
@@ -13,10 +13,19 @@ export function formatAddress(address: string) {
...
   return !!val;
 }
 
+// Accepts number or string and rounds it to the given number of decimals
+export function roundNumber(number: string | number, decimals: number) {
+  const mul = 10 ** decimals;
+  const num = typeof number == 'number' ? number : parseFloat(number);
+
+  return Math.round(num * mul) / mul;
+}
```
mds1, line 27: I have a similar function in my branch like this, but it also adds commas for large numbers. Feel free to use/combine these two if you'd like (`commify` comes from ethers in `src/utils/ethers`). I named it `formatNumber` to avoid naming confusion of rounding numbers with the concept of a "grant round" 🤷

```typescript
// Formats a number for display to the user, by rounding to the specified number of decimals and adding commas
export function formatNumber(value: BigNumberish, decimals: number): string {
  // `BigNumber.from()` can't take decimal inputs
  value = typeof number === 'number' || typeof number === 'string' ? Number(value) : BigNumber.from(value);
  return commify(parseFloat(value.toFixed(decimals)));
}
```

dcurve/src/internal/clr.ts
```diff
@@ -124,26 +124,28 @@ export class CLR {
...
-      const predictedGrantMatch = getGrantMatch(grantId, newDistribution);
-
-      // push the prediction for the predictionPoint
-      predictions.push({
-        predictionPoint: predictionPoint,
-        predictedGrantMatch: predictedGrantMatch,
-        predictionDiff: predictedGrantMatch - currentGrantMatch,
-      } as GrantPrediction);
-    });
+    void (await Promise.all(
```
mds1, line 127: I think you could do this instead, not that it really matters though

```suggestion
    void Promise.all(
```

utils/src/trustBonus.ts
```diff
@@ -31,7 +31,7 @@ export const fetchTrustBonusScore = async (addresses: string[]): Promise<TrustBo
   const url = `https://gitcoin.co/grants/v1/api/trust-bonus?addresses=${addresses.join(',')}`;
 
   try {
-    const response = await fetch(url);
+    const response = await fetch(url, { mode: 'no-cors' });
```
mds1, line 34: Why did we need to add the `no-cors` flag?

mds1 (changes requested):
Looking great! Left a bunch of questions/comments but they're all small and should be easy fixes

app/src/components/BaseHeader.vue
mds1: Looks like the arrows don't show up, and we just get an empty `span` tag, which is ok but would be a little cleaner to use a `v-if` I think so we don't have unused HTML

app/src/components/BaseFilterNav.vue
```diff
@@ -0,0 +1,64 @@
+<template>
```
mds1, line 1: Also getting some errors in the console (likely unrelated to this file, just using it to start a thread)

image

apbendi (approved):
Looking great @gdixon. My concerns have definitely been addressed. I'll let you respond to @mds1 and others' feedback independently, but from my perspective this is approved.

One small non-blocking nit that I would add: can we make the links to etherscan addresses and transactions (the owner, the donations, etc...) open in a new tab?

apbendi:
@gdixon can you check out the errors that @mds1 are reporting to come from this branch, per his pr here: https://github.com/dcgtc/dgrants/pull/160#issuecomment-909438210. Any ideas on these? Can we get those resolved before merging this?

types/src/grants.d.ts
```diff
@@ -18,6 +19,25 @@ export type GrantMetadata = {
   name: string;
   description: string;
   logoURI?: string;
+  quote?: string;
+  lastUpdated?: string;
```
mds1, line 23: Per @apbendi in [Discord](https://discord.com/channels/562828676480237578/882302340655374437/882331205612240916), we should remove these two properties for now

app/src/components/BaseTitle.vue
gdixon: It's an optional counter which sits next to the title, we use this component multiple times on each page to separate sections - I've not seen where the counter is being used (yet) but it is part of @scco's designs - can you provide any more info here Richard?

utils/src/trustBonus.ts
gdixon: Ahh we didn't, I was getting CORS errors so I added this to silence them!

dcurve/src/internal/clr.ts
gdixon: I think we need to await here because the map has side-effects (collating the predictions)

app/src/utils/utils.ts
gdixon: I've copied it from here but altered it a little bit:

```ts
// Formats a number for display to the user, by rounding to the specified number of decimals and adding commas
export function formatNumber(value: BigNumberish, decimals: number): string {
  // `BigNumber.from()` can't take decimal inputs
  value = typeof value === 'number' || typeof value === 'string' ? Number(value) : BigNumber.from(value);
  return commify(parseFloat(String(value)).toFixed(decimals));
}
```

types/src/app.d.ts
gdixon: I've changed this to `routeTarget` so that we can feed in params/query args too 👍

app/src/components/BaseHeader.vue
gdixon: Yes, that sounds better to me too!

app/src/components/ContributionRow.vue
gdixon: I've added a definition for `ContributionEvent` and rolled the `donationToken` into that definition 👍

app/src/components/GrantDetailsRow.vue
gdixon: I've moved the fallback into the template 🙌

app/src/components/BaseFilterNav.vue
gdixon: I think I've got this resolved, can you let me know if this is fixed for you?

types/src/app.d.ts
gdixon: I've added comments here ✊

types/src/grants.d.ts
gdixon: I took these names from the issue @apbendi created, do we want to change these?

app/src/components/BaseTitle.vue
gdixon: So I think this is a section heading -- should we go for something like that? `SectionHeading` / `SectionHeader`?

types/src/grants.d.ts
mds1: My understanding is that this is the type we want to use for this PR

```typescript
type GrantMetadata = {
  name: string;
  description: string;
  logoURI?: string;
  properties: {
    websiteURI?: string;
    githubURI?: string;
    twitterURI?: string;
    [key: string]?: string // for extensions/flexibility
  }
}
```

app/src/components/BaseTitle.vue
mds1: Either of those sounds good to me!

app/src/utils/utils.ts
mds1: LGTM! I'll rebase against your latest branch and use this version also

dcurve/src/internal/clr.ts
mds1: Ah got it, sounds good!

app/src/components/BaseFilterNav.vue
mds1: Yes I no longer see it!

app/src/components/BaseTitle.vue
mds1: We can always remove it for now and add it back later as needed. Personally I'm not a huge fan of shipping unused code that may be used soon ™️

gdixon: It is gone! 

Yeah thats cool, I'm not either - its tricky knowing what to include and exclude when setting up components, it is probably the right move to just implement what we need though 👍

types/src/grants.d.ts
gdixon: Awesome, I've moved to this definition:

```ts
export type GrantMetadata = {
  name: string;
  description: string;
  logoURI?: string;
  properties: {
    websiteURI?: string;
    githubURI?: string;
    twitterURI?: string;
  } & {
    [key: string]: string; // for extensions/flexibility
  };
};
```

app/src/components/BaseTitle.vue
gdixon: I've gone with `SectionHeader` 🙌

mds1 (approved):

PR #162, opened 2021-08-31
Layout header

carletex:
Hey, 

Still a couple of ToDo and a TS linting issue (See review comments).

@mds1 Could you take a look? Also any comments on the Vue stuff are more than welcome :)


app/src/components/LayoutHeader.vue
```diff
@@ -42,30 +42,138 @@
...
+    changeWallet() {
+      // ToDo implement changeWallet
+      console.log('changing wallets');
+    },
+  },
   setup() {
-    const { connectWallet, isSupportedNetwork, userDisplayName } = useWalletStore();
-    return { connectWallet, isSupportedNetwork, navigation, userDisplayName };
+    const { connectWallet, isSupportedNetwork, provider, userDisplayName, userAddress } = useWalletStore();
+    return { connectWallet, isSupportedNetwork, navigation, provider, userDisplayName, userAddress };
```
carletex, line 176: I'm getting this TS error here. 

TS4082: Default export of the module has or is using private name 'UncheckedJsonRpcSigner'.

image

Any ideas?

app/src/components/LayoutHeader.vue
```diff
@@ -42,30 +42,138 @@
...
+      this.balance = formatEther(balance ?? '0');
+    },
+    disconnect() {
+      // ToDo implement disconnect
+      console.log('disconnecting');
+    },
+    changeWallet() {
+      // ToDo implement changeWallet
+      console.log('changing wallets');
+    },
```
carletex, line 172: I could use some help implementing these. Is there a method inside useWallet store that I could use?

app/src/components/Jazzicon.vue
```diff
@@ -0,0 +1,32 @@
+<template>
```
carletex, line 1: Copied this file from Graham's #158

apbendi (commented):
This is looking siiick already! Nice work.

A few things I noticed on a quick first pass:

* Compared to the design, there seems to be too much padding between the address icon and the carat. Also, the location from which the menu is unfolding seems off:

Screen Shot 2021-08-31 at 11 52 54 AM


* Clicking on the cart icon does not bring me to the cart
* Any reason we can't wire up the number of things in the cart to the actual underlying data in local storage?


Also, here are some other minor things that might also be appropriate to put in this PR. Let me know what you think:

* Removing the footer
* Moving the copyright notice, twitter link, and discord link into the About modal (any objection to this @scco / @phutchins ?)
* When the "About" modal is up on the screen, it feels like it should be possible to dismiss it by clicking outside of it, in addition to clicking the "X"

app/src/components/LayoutHeader.vue
mds1: Easy fix: Don't return `provider`! The only things that should be returned from the `setup()` method are things needed in the HTML template, and it seems `provider` is not used in the template (I'd be surprised if you do need it in the HTML template)

app/src/components/LayoutHeader.vue
mds1: We use the onboard.js package to manage wallet state, so:
- For `disconnect`, use [`onboard.walletReset()`](https://docs.blocknative.com/onboard#walletreset)
- For `changeWallet`, use [`await onboard.walletSelect()`](https://docs.blocknative.com/onboard#walletselect). For this one, it should automatically update the `last-wallet` in local storage when you change wallets (see where we call `setLastWallet`), but if not you'll need to do that also

app/src/components/LayoutHeader.vue
```diff
@@ -42,30 +42,138 @@
...
 
 <script lang="ts">
 import { defineComponent } from 'vue';
 import useWalletStore from 'src/store/wallet';
 import { WarningIcon } from '@fusion-icons/vue/interface';
+import { Wallet3Icon as ConnectWalletIcon } from '@fusion-icons/vue/web3';
+import { ArrowBottom2Icon as ArrowBottomIcon } from '@fusion-icons/vue/interface';
+import Jazzicon from 'src/components/Jazzicon.vue';
+import { Cart2Icon as CartIcon } from '@fusion-icons/vue/interface';
+import { formatEther, JsonRpcProvider, Web3Provider } from 'src/utils/ethers';
```
mds1, line 136: A nit, but let's organize all the imports for readability. We don't have hard rule here, but for some inspiration here's how we do it in `Cart.vue`
image

app/src/components/LayoutHeader.vue
```diff
@@ -42,30 +42,138 @@
...
 
 // Header menu bar items
 const navigation = [
   { name: 'Home', href: '/' },
   { name: 'Cart', href: '/cart' },
-  { name: 'Contact', href: '/contact' },
 ];
```
mds1, line 142: A bit of a nit, but to me href indicates a full URL, but here we just link to other parts of the app. Also IMO best practice for routing is to route by name, since that's just internal so we can ensure it never changes, whereas the route slug may change. You can find a route's `name` property defined in `app/src/router/index.ts` 

Therefore you'd have something like this:

```typescript
const navigation = [
  { name: 'Home', name: 'Home' },
  { name: 'Cart', name: 'Cart' },
];
```

More info: https://next.router.vuejs.org/guide/essentials/named-routes.html

app/src/components/LayoutHeader.vue
```diff
@@ -42,30 +42,138 @@
...
+            flex-col
+            gap-y-2
+            uppercase
+            font-medium
+            text-left
+            whitespace-nowrap
+          "
+        >
+          <div>{{ userDisplayName }}</div>
+          <div>{{ balance }} ETH</div>
```
mds1, line 95: @apbendi @phutchins Why do we want to show a user's ETH balance in the header, given that 75% of user's check out with DAI? I think this should be removed as I'm not sure what value it adds

app/src/components/LayoutHeader.vue
```diff
@@ -42,30 +42,138 @@
...
   emits: ['toggleAbout'],
+  data() {
+    return {
+      balance: '0',
+      provider: null as Web3Provider | JsonRpcProvider | null,
+    };
+  },
+  watch: {
+    userAddress: 'updateBalance',
+  },
```
mds1, line 156: Instead of a watcher and async method, this is part of the wallet state, so storing their balance should probably be managed in the `wallet.ts` store. However, IMO this should be removed so before making this change let's see what the decision is around that

mds1:
So we might have never conveyed this to you, so apologies if that's so, but this component is written with the Options API. We use the Composition API because it's easier to keep things organized with large apps, and has much better TypeScript support (which may be why you received the TS error in your above comment). 

I recommend checking out the docs for the Composition API if you're not familiar, and check out some other components too, but basically you'll want to refactor the `<script>` section to something like this:

```typescript
<script lang="ts">
import { defineComponent, ref } from 'vue'; // notice the new `ref` import
// .... rest of imports here

// Header menu bar items
const navigation = [
  { name: 'Home', href: '/' },
  { name: 'Cart', href: '/cart' },
];

// Composition function for wallet management in the header
function useWalletConnection() {
  // Reads/writes to the src/store/wallet.ts store
  const { /* stuff */ } = useWalletStore() // these are reactive so UI will automatically update when store changes

  function disconnect() {
    // call wallet store method to disconnet
  }
  
  function changeWallet() {
    // call wallet store method to change wallet
  }

  return {disconnect, changeWallet,  /* stuff you need in HTML template... */ }
}

export default defineComponent({
  name: 'LayoutHeader',
  components: { ConnectWalletIcon, ArrowBottomIcon, CartIcon, Jazzicon, WarningIcon },
  setup(_props, context) {
    const emitEvent = (eventName: string) => context.emit(eventName); // call `emitEvent('toggleAbout')` in HTML template
    return { ...useWalletConnection(), emitEvent };
  },
});
```

If you need help with this, let me know and I can push a commit that refactors your component to show you what I mean

app/src/components/LayoutHeader.vue
carletex: Makes total sense. Changed it! Using **label** and **name** properties now.

app/src/components/LayoutHeader.vue
carletex: Sure! I see that you'all use more "readability comments" than I'm used to. I'll adapt myself :)

carletex:
Thanks both for all the insightful comments. Learning a lot.

> So we might have never conveyed this to you..

I've actually only had a (great) 15 min overview of the project with Philip. It'd be great for me if we can meet at some point to go deeper into the tech details of the project (I could use discord to ask questions, but I think a call could be more efficient for this first time). 

@apbendi I created this PR (draft) in order to ask Matt some questions (easier to make in-code-review-comments) but still had some work to do. Anyway, I'm glad you reviewed it regardless, because some of your comments weren't on my ToDo list :)

> - [x] Compared to the design, there seems to be too much padding between the address icon and the carat. Also, the location from which the menu is unfolding seems off:

Should be in a better shape now. 

> - [x] Clicking on the cart icon does not bring me to the cart
> - [x] Any reason we can't wire up the number of things in the cart to the actual underlying data in local storage?

Should be done.

> - [x] When the "About" modal is up on the screen, it feels like it should be possible to dismiss it by clicking outside of it, in addition to clicking the "X"

Make sense. Done!

> - [ ] Removing the footer
> - [ ] Moving the copyright notice, twitter link, and discord link into the About modal (any objection to this @scco / @phutchins?)

I'm waiting on confirmation for this.

@mds1 Thank you for your detailed comments! I've answered some of them. I'll wait until we have confirmation on:

> -[ ] @apbendi @phutchins Why do we want to show a user's ETH balance in the header, given that 75% of user's check out with DAI? I think this should be removed as I'm not sure what value it adds

app/src/components/LayoutHeader.vue
carletex: I implemented [disconnectWallet](https://github.com/dcgtc/dgrants/pull/162/files#diff-a870919603ea024ec4349c9ed046dfae2dc99dd8405389e17f9af666c3cd2f00R116-R121). Is it all we need?

Regarding changeWallet, just to be clear: we just want to show the modal with the wallet providers so the user can switch to another one, right? Something like:

```js
await onboard.walletSelect(); 
await onboard.walletCheck();
await configureProvider(); 
// Make sure last-wallet is updated
```

phutchins:
Odd, I thought we were only showing their eth address not balance. That doesn't make sense to me either but I'll have to go back and look at the designs.

app/src/components/LayoutHeader.vue
mds1: > I implemented disconnectWallet. Is it all we need?

Hm, in the linked method, where is `resetState` defined? I can't find what that is. But otherwise that looks right to me, and I think you really only need the `onboard.walletReset()` call

> Regarding changeWallet, just to be clear: we just want to show the modal with the wallet providers so the user can switch to another one, right? Something like:

Yep that is the intention, and yes that snippet looks correct to me, but make sure to test both as confirmation!

app/src/components/LayoutHeader.vue
mds1: Heh yes IMO more comments is the way to go, especially for helping future contributors. Thanks! 🙂

app/src/components/LayoutHeader.vue
mds1: Ah yes thanks for fixing my broken snippet above haha, `label` and `name` sounds great 👌

app/src/store/cart.ts
```diff
@@ -300,13 +300,19 @@ export default function useCartStore() {
     return summary.slice(0, -3); // trim the trailing ` + ` from the string
   });
 
+  /**
+   * @notice Returns the amount of items in the cart
+   */
+  const cartItemsCount = computed(() => Object.entries(cartSummary.value).length);
```
mds1, line 306: `cartSummary` condenses e.g. two cart items of 2 DAI and 3 DAI into a summary of "5 DAI", so using this for the count will return the wrong information. `cart` stores the full cart data, so this should be changed as follows.
```suggestion
  const cartItemsCount = computed(() => cart.value.length);
```

And if you want, for brevity, this getter is simple enough that you could even inline it like this:

```typescript
return {
  cart,
  cartItemsCount: computed(() => cart.value.length),
  // other stuff in the return block
}
```

app/src/utils/ethers.ts
```diff
@@ -27,5 +27,5 @@ export { BigNumber } from '@ethersproject/bignumber';
 export { hexDataSlice } from '@ethersproject/bytes';
 export { Contract } from '@ethersproject/contracts';
 export { JsonRpcProvider, JsonRpcSigner, Web3Provider } from '@ethersproject/providers';
-export { commify, formatUnits, parseUnits } from '@ethersproject/units';
+export { commify, formatEther, formatUnits, parseUnits } from '@ethersproject/units';
```
mds1, line 30: Looks like this export was added but I don't see it used

app/src/components/About.vue
```diff
@@ -1,11 +1,11 @@
...
-    class="h-screen fixed w-full"
-    style="z-index: 10000; background-color: #ffffffdd"
+    class="h-screen fixed w-full z-50 bg-white bg-opacity-80"
+    @click="emitEvent('toggleAbout')"
   >
     <div class="px-4 md:px-12 m-auto bg-white">
-      <div class="border-grey-500 border-2 px-4 md:px-12 relative">
+      <div class="border-grey-500 border-2 px-4 md:px-12 relative" @click.stop>
         <div class="absolute top-6 right-6">
           <XIcon @click="$emit('toggleAbout')" class="icon icon-primary cursor-pointer" />
```
mds1, line 10: Should change `$emit` to the `emitEvent` method

app/src/components/LayoutHeader.vue
carletex: `resetState` is defined on [wallet.ts](https://github.com/dcgtc/dgrants/blob/main/app/src/store/wallet.ts#L50)

```js
// Reset state when, e.g.user switches wallets. Provider/signer are automatically updated by ethers so are not cleared
function resetState() {
  userAddress.value = undefined;
  network.value = undefined;
}
```

Without calling it the UI wasn't updating properly (Still saw my connected wallet)

app/src/components/About.vue
carletex: Missed that one. Fixed!

app/src/utils/ethers.ts
carletex: It was used when were showing the ETH total, but still not clear if we are going to show it. I'll remove it for now.

app/src/store/cart.ts
carletex: Oh, I see it now. Thanks. Fixed.

app/src/components/LayoutHeader.vue
mds1: Oh yes perfect, good call. Sorry I only searched for it in the diff—forgot it was a function we already have

apbendi (approved):
This looks and works great @carletex. Nice work on this 👍

phutchins:
Looks great to me! Very clean.

phutchins (approved):
Looks good!

carletex:
Thanks all!

PR #165, opened 2021-09-01
Add version bump commands to package.json

phutchins:
This PR adds commands to the package.json to handle bumping versions. There are three commands (bump-patch, bump-minor, bump-major) which increase the version specified respectively and the short commit hash.

Screen Shot 2021-09-01 at 3 51 09 PM

Currently all of our packages are in the same repo so they will get the same version bumps regardless of being changed or not. I have added the `--force-publish` flag which forces the version update to happen even if no changes are detected. Once we use lerna to break them out into separate repositories, we can utilize the independent versioning feature.

This setup does not currently manage versions in the top level package.json so that is something to discuss.

The second set of three commands (release-patch, release-minor, and release-major) move from a pre-release version which includes the short commit id to a regular semantic release version.

Screen Shot 2021-09-01 at 3 59 21 PM

I will need to clean up the tags and commits prior to merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,45 @@
+## 0.0.1 (2021-09-10)
+
+
+### Bug Fixes
+
+* button in TransactionStatus component now works ([232adbb](https://github.com/dcgtc/dgrants/commit/232adbb0381440843e4e31c3df4294b8214c84d6))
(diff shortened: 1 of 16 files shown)


apbendi (approved):
LGTM!

gdixon (approved):
All looks good, but do we want to be bumping the versions now or were these for testing?

phutchins:
Have cleaned up the changelogs from testing and am now creating the initial changelog, ,and adding details to the readme on how to use versioning. We can merge after that is complete.

Issue #168, opened 2021-09-02
Occasional Errors Related To Matching Calculations


apbendi:
While testing out the app, The following message occurs— somewhat randomly— on sporadic occasions. Once, for example, it appeared when clicking the header "cart" link and navigated to an empty cart, but this is not reliably reproducible.

<img width="298" alt="Screen Shot 2021-09-02 at 11 04 14 AM" src="https://user-images.githubusercontent.com/1479802/131873273-72669fed-be2b-4256-844b-26d4e259d273.png">

<img width="1053" alt="Screen Shot 2021-09-02 at 11 01 42 AM" src="https://user-images.githubusercontent.com/1479802/131873299-e79aa365-efe7-4956-bb21-49f106bad8ed.png">

PR #169, opened 2021-09-02
fix: Fixes missing match value for when the filter fails

gdixon:
To discover each prediction point we filter the distribution to find the match for the grant - there are situations where the filter will not find a result and instead we should default to a match of 0.

--

Fixes: #168


apbendi (approved):
Seems to have resolved the issue. LGTM. Thanks for tackling this quickly @gdixon!

phutchins (approved):
Looks good.

PR #172, opened 2021-09-03
feat: add GrantRoundPage + add metadata

thelostone-mc:
### Description

- introduced new view GrantRound  which shows details of a given round
- tweak IPFS GrantRound metaPtr to include new data elements
- display new elements on GrantRound Page
- update routes so that 
  - `dgrants/rounds/:address`  shows Grant round Info 
  - `dgrants/rounds/:address/grants` shows the actual Grants in the GrantRound
  
  
### Pending

- Tweaking the UI based on design (this could be in a standalone PR)

Here is the metaPtr structure: http://storageapi.fleek.co/thelostone-mc-team-bucket/round.json

### DIFF
--- a/app/src/views/GrantRound.vue
+++ b/app/src/views/GrantRound.vue
@@ -1,0 +1,305 @@
+<template>
+  <!-- GrantRound details -->
+  <div v-if="!isContributing">
+    <div v-if="grantRound.address">
+      <h1 class="my-6 text-center text-3xl font-extrabold text-gray-900">
+        Details for Grant Round: <span :title="grantRound.address">{{ formatAddress(grantRound.address) }}</span>
+      </h1>
+      <p>Name: {{ grantRoundMetadata?.name }}</p>
+      <p>Description: {{ grantRoundMetadata?.description }}</p>
+      <p>Status: {{ grantRound.status }}</p>
+      <p>Matching: {{ grantRoundMetadata?.matchingAlgorithm }}</p>
+      <p>Website: {{ grantRoundMetadata?.properties?.websiteURI }}</p>
+      <p>Governance: {{ grantRoundMetadata?.properties?.governanceURI }}</p>
+      <p>Twitter: {{ grantRoundMetadata?.properties?.twitterURI }}</p>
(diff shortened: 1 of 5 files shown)


apbendi:
Good start on getting this reworked! I'm ok with merging it separately if that makes it easier to parallelize other work.

Initial comments:

* We should discuss the additional round properties at standup today. Not sure we want/need them for the MVP, but fine with having them if they're implemented and we all agree they're useful.
* When I click "Show Grants" it still shows all the other Round data, which is not what the design does. Is this just because you haven't done the design work? Would it be easy to just remove all that extra metadata on the show-grants page, just so it's clearer right now the difference between the two pages?

apbendi (approved):
Good to go on this one.

app/src/views/GrantRound.vue
```diff
@@ -0,0 +1,305 @@
...
+          :title="grantRound.matchingToken.name"
+          :href="`https://etherscan.io/token/${grantRound.matchingToken.address}?a=${grantRound.address}`"
+          target="_blank"
+          rel="noopener noreferrer"
+        >
+          {{ grantRound.funds.toString() }} {{ grantRound.matchingToken.symbol }}
+        </a>
+      </p>
+      <p>
+        {{ hasStatus('Upcoming')(grantRound) ? 'Will begin' : 'Started' }}:
```
mds1, line 31: Feels clunky that this is two function calls, instead of one function call with a single argument. @gdixon I thought we had changed this a few weeks ago? I think it'd be more clear if this was `hasStatus('Upcoming', grantRound)`

gdixon: I was using currying here because it neatens up the sorts, but if its being used for other things then we can make that change!

gdixon (approved):
All lgtm!! @mds1 - let me know if we want to make the `hasStatus()()` change in this PR or as a follow up!

app/src/views/GrantRound.vue
mds1: Nope, fine by me! Just misremembered what we said when I saw this last time 🙂

PR #177, opened 2021-09-03
improved ui for cart

scco:
i worked on the cart to get my ui in, and everything worked fine. this is a mandatory pull request and should get merged asap to continue working on it ( usdt / matching and so on ) ...

**anyway there is 2 issue here we need to be aware of and fix :**

the cart uses  `<BaseInput>` and `<BaseSelect>` Components (pre-design-created)  what i had to tweak. the `<BaseSelect>` uses some non-native fake-select-input-fields, what are hard to combine with regular native `<input>` ( see different heights on some resolutions )

Bildschirmfoto 2021-09-03 um 19 44 16

Input Fields are always hard to Controll - especialy when you work with paddings, and percentage based sizes - so its not a good idea to mix different approaches to style it. I would 100% prefer if we can go a step back and remove that headless ui components and use regular html `<select>` what would remove that issue and be less overhead / complicated code. i can not do this - as its technical to complex ( but i put an example in the .vue file how i would like to use it if possible ) 

Issue2  is, that i can not see the grants titles in every item in a cart until i update the input field ( eg from 5 to 6 DAI ) . this might need to be inspected and has probably todo with ipfs ?

we are getting a bit closer here in design from what i had in mind.

Bildschirmfoto 2021-09-03 um 19 50 23

### DIFF
--- a/app/src/assets/placeholder_grant.svg
+++ b/app/src/assets/placeholder_grant.svg
@@ -1,0 +1,9 @@
+<svg width="1920" height="1080" viewBox="0 0 1920 1080" fill="none" xmlns="http://www.w3.org/2000/svg">
+<rect width="1920" height="1080" fill="white"/>
+<path d="M984.188 635.875H935.813" stroke="#E2E0E7" stroke-width="6" stroke-miterlimit="10"/>
+<path d="M988.438 535.875L1014.87 583" stroke="#E2E0E7" stroke-width="6" stroke-miterlimit="10"/>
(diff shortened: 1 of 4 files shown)


apbendi (approved):
My only feedback is that on mobile, having the "X" to remove from cart centered and on the right feels a bit odd to me.

Screen Shot 2021-09-07 at 7 56 49 AM

mds1:
Just rebased and pushed. @scco Can you take a look and make sure things still look good?

scco:
looks good to me, i merged it in my local version of this branch and it seems to be fine . thanks a lot !

phutchins (approved):
All looks good. We can do any cleanup needed in subsequent PRs.

PR #180, opened 2021-09-04
grants detail ui enhancement

scco:
- improved the grantsdetailsrow with code from richard-ui3
- removed v-slot from that component for extra interactions
- outsourced interactions into the grantregistrygrantdetail.vue in a interaction nav bar
- removed that css from app.vue ( as it bothers me and some css )
- added a placeholder_grant.svg to use when there is no img from ipfs

### DIFF
--- a/app/src/App.vue
+++ b/app/src/App.vue
@@ -1,8 +1,8 @@
 <template>
   <About :showAbout="showAbout" @toggle-about="toggleAbout" />
-  <div class="flex flex-col min-h-screen">
+  <div>
     <layout-header id="header" @toggle-about="toggleAbout" />
-    <main id="app-main" class="flex-grow bg-white"><router-view /></main>
+    <main id="app-main"><router-view /></main>
   </div>
 </template>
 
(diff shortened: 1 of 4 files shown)


apbendi:
@scco I'm a little concerned with how huge the browser has to be before you get a more standard "desktop" style layout. In my normal browser layout, I'm getting a single column layout, which means I don't even see the full image above the fold:

Screen Shot 2021-09-07 at 8 06 33 AM

mds1:
Code looks good!

Agree with @apbendi that it does seem on this branch the mobile view shows up at a wider width than expected, whereas that is not the case on `main`

Also the fact that there's no max width makes the description look a bit odd on large windows—it feels like it should be left-aligned perhaps?

image

scco:
it's all as it should be, with intention and purpose.
i agree the demo image does not look super good in here, but its a placeholder.

apbendi (approved):

thelostone-mc (approved):

mds1:
Started rebasing this one, but it turned out to be a pain since @scco @thelostone-mc and @mk1r all made changes to `GrantRegistryGrantDetail.vue`, so the version on `main` are causing conflicts.

There's a lot going on in that file and I'm not as familiar with it all since I haven't worked on it, so can you guys all work together on resolving this one? Happy to help as needed

apbendi:
Closed for #185

PR #181, opened 2021-09-07
Gas snapshots

### LINKED ISSUE #163: Assess Gas Costs for a "typical" Proof-of-Concept Donation
With gas prices through the roof, we want to assess— and perhaps snapshot as part of our tests— a few "typical" donation patterns expected during the MVP round. The most likely candidates, in order of importance, are:

- [ ] Donating to a handful of grants (say 5) with just Dai (where Dai is assumed to be the donationToken)
- [ ] Donating to a 5 grants with just ETH
- [ ] Donating to 5 grants with a token other than Dai or ETH

With these snapshots in place, we can look for low-hanging fruit opportunities to gas golf and assess the impact of the changes.

mds1:
Closes #163 

Uses the same approach as Uniswap, described here: https://gist.github.com/gakonst/258a6ccd6bb752248999a13e8dd82a64

Below is a copy of the snapshot results that I manually augmented with USD prices assuming $3500 ETH and 100 gwei gas prices:

```
exports[`dGrants gas tests GrantRegistry createGrant 1`] = `166643` = ($58)

exports[`dGrants gas tests GrantRegistry updateGrantMetaPtr 1`] = `47021` = ($16)

exports[`dGrants gas tests GrantRegistry updateGrantMetaPtr 2`] = `54049` = ($18)

exports[`dGrants gas tests GrantRegistry updateGrantOwner 1`] = `41611` = ($14)

exports[`dGrants gas tests GrantRegistry updateGrantPayee 1`] = `41614` = ($14)

exports[`dGrants gas tests GrantRoundManager createGrantRound 1`] = `923202` = ($323)

exports[`dGrants gas tests GrantRoundManager five donations input token DAI, output token DAI 1`] = `231241` = ($80)

exports[`dGrants gas tests GrantRoundManager five donations input token ETH, output token DAI 1`] = `327195` = ($114)

exports[`dGrants gas tests GrantRoundManager five donations input token GTC, output token DAI 1`] = `394453` = ($138)

exports[`dGrants gas tests GrantRoundManager one donation input token DAI, output token DAI 1`] = `102340` = ($35)

exports[`dGrants gas tests GrantRoundManager one donation input token ETH, output token DAI 1`] = `195411` = ($68)

exports[`dGrants gas tests GrantRoundManager one donation input token GTC, output token DAI 1`] = `249216` = ($87)
```

Created https://github.com/dcgtc/dgrants/issues/182 because CI was failing due to snapshots seemingly not being deterministic

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -54,6 +54,6 @@ jobs:
       - name: Run test
         run: yarn test
         env:
-          INFURA_ID: ${{ secrets.INFURA_ID }}
+          ALCHEMY_API_KEY: ${{ secrets.ALCHEMY_API_KEY }}
       - name: Run lint
         run: yarn lint
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Copy `contracts/.env.template` to `contracts/.env` and edit, providing your own 
 cp contracts/.env.template contracts/.env
 ```
 
-You will have to supply an [Infura](https://infura.io/) Project ID. An account with their free tier is sufficient for development. You'll also need a [Fleek](https://fleek.co/) IPFS endpoint key in your .env file.
+You will have to supply an [Alchemy](https://www.alchemy.com//) API Key. An account with their free tier is sufficient for development. You'll also need a [Fleek](https://fleek.co/) IPFS endpoint key in your .env file.
 
 ### Develop
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ To ensure that everyone is using the same version of nodejs on this project, [vo
 
 ### Set your env files
 
-Copy `app/.env.template` to `app/.env` and edit, providing your own env vars. You will have to supply a [Blocknative](https://www.blocknative.com/) API key and an [Infura](https://infura.io/) Project ID. Both services have free tiers which are sufficient for development.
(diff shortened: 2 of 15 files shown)


thelostone-mc (approved):
This does seem too expensive for mainnet from the looks of it :/

mds1:
>  I don't think we should merge this as-is because it causes running tests locally to fail....I guess one option would be to include UPDATE_SNAPTSHOT=1 in the test command and gitignore the snapshots file. This means we lose the ability to make changes to contracts and see the changes to gas usage in the commit history, so it's not ideal. Any other ideas?

Ah true. I think the easiest is to include UPDATE_SNAPTSHOT=1 in the test command like you said, but _don't_ ignore the snapshots file. This way any PR that affects the contracts will update the snapshot file, and part of the PR will be reviewing that diff to see if the changes are significant or just noise

mds1:
Also try pinning tests at a block which might fix the determinism

apbendi (approved):
Working great with Alchemy and the pinned block number. I'm curious if there will be any downsides as we move further away from the pinned block, i.e. token prices that are way off in the UI or whatever. Cross that bridge when we get there.

One thing to note, we need to communicate really clearly to the team about the need to update their `.env` var before this gets merged, or everyones dev environment will break at once.

Also, let's get the CI fixed before merging.

PR #183, opened 2021-09-08
enhance / fix about modal

scco:
just worked on  this file : components/About.vue ! ignore other files when you see conflicts with main branch.

- fixing the donate to dgrants button to a real <button> for less code + consistency + it had some css bugs in ( related to positioning of next elements after the button ... )
- removing that inline / incode icons as it was not planed in my designs to have icons here. actualy i wanted to have NO icons to have a pragmatic look on the about modal, but i added 2 buttons for github and twitter now.
- removed telegram for now as we do not have a telegram and we are not have telegram yet.

### DIFF
--- a/app/src/components/About.vue
+++ b/app/src/components/About.vue
@@ -1,15 +1,17 @@
 <template>
   <div
     :class="[showAbout ? 'flex' : 'hidden']"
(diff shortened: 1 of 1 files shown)


apbendi (approved):
Looks good 👌

thelostone-mc (approved):

Issue #112, opened 2021-08-17
Master Issue: Convert and Integrate Static Design Elements


apbendi:
This is a tracking issue for what will be a number of discreet tasks, each of which can have its own PR, or be bundled together in small batches.

Using [richard-ui3](https://github.com/dcgtc/dgrants/tree/richard-ui3) as the canonical branch, we'll convert the various static design components he's created into components with props & UI mechanics as required, AND integrate those components into the appropriate portions of the existing app.

Reference the [Figma designs](https://www.figma.com/file/sUQSYyPfooqTcVdajfZm0t/dgrants?node-id=92%3A80) for context as to how/where components should be integrated into the UI.

Checkout [UI.vue](https://github.com/dcgtc/dgrants/blob/richard-ui3/app/src/views/Ui.vue) to see an example built by @scco of the components laid out on a page.

- [x] #116
     - [x] https://github.com/dcgtc/dgrants/pull/107/files#diff-0e8cfeba27f9bd404b3c20c24f4b8dbc1d30afcccd9fa54ddb16dafb69e4e252
     - [x] https://github.com/dcgtc/dgrants/pull/107/files#diff-dd2de3fc01fad5ac9a6c0e977369ede9376bf11738c028cd108f47ef5f797989
     - [x] https://github.com/dcgtc/dgrants/pull/107/files#diff-25759d004f59de124ac11fa8264f622e252812898c32ea7b145ed2351366ead3
     - [x] https://github.com/dcgtc/dgrants/pull/107/files#diff-e801e606ad9817fd2e8265cc54fe2fd2bcc1e5de2e68458d20b9c8c3f342de1a
     - [x] https://github.com/dcgtc/dgrants/pull/107/files#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2de
- [x] #134
     - [x] https://github.com/dcgtc/dgrants/blob/richard-ui3/app/src/components/BaseHeader.vue
     - [x] https://github.com/dcgtc/dgrants/blob/richard-ui3/app/src/components/LayoutHeader.vue
- [x] #122
- [x] ActionsNav (https://github.com/dcgtc/dgrants/blob/richard-ui3/app/src/components/ActionsNav.vue)
- [x] BaseFilterNav (implemented by #158)
- [x] BaseHeader (implemented by #158) 
- [x] BaseTabs (implemented by #184)
- [x] BaseTitle (implemented by #158)
[...]

PR #184, opened 2021-09-08
Grant Registry List Page

### LINKED ISSUE #148: UI Conversion: Grant Registry List Page
#### Description

This task involves building up the GrantRegistry Page

**Image:** [HERE](https://user-images.githubusercontent.com/5358146/130941818-90cd6af6-3a79-4849-b818-dd944fca6112.png)
 **Figma:** https://www.figma.com/file/sUQSYyPfooqTcVdajfZm0t/dgrants?node-id=350%3A2695
 
 Below is the checklist of components which needs to be built out to complete this page 
 
- [ ] **component**: Breadcrumbs
- [ ] **component**: BaseHeader
- [ ] **component**: BaseTabs 
  - [ ] **component**: BaseFilterNav
- [ ] **component**: GrantCard (list of grants)

NOTE: Update the Grant Registry list page (i.e. `/dgrants`) to match with [design](https://www.figma.com/file/sUQSYyPfooqTcVdajfZm0t/dgrants?node-id=92%3A80) using [Richard's branch](https://github.com/dcgtc/dgrants/tree/richard-ui3).

TODO: Define which components from #112 will be converted in the process of pulling over this page

carletex:
Fixes #148 

WIP. Marked in #112 the new components that are parts of this PR.

Questions:
 - How do I know the parent round of a given grant?
 - How do I get a list of possible categories (dApp Tech, Community, etc)?
 - Some data is missing from the grants info (or I don't know how to get it): "raised", "image", "address", "date", "categories", and "curation".
 -  Where does the owner name link to?


carletex:
Notes from the meeting:

Screenshot from 2021-09-08 17-48-23

app/src/components/GrantList.vue
```diff
@@ -1,85 +1,44 @@
...
-          <button v-else @click="addToCart(grant.id)" class="my-2 btn btn-primary">Add to Cart</button>
-        </div>
+    <!-- ToDo. Loading... message-->
+    <ul class="text-left grid grid-cols-1 gap-10 py-10 sm:grid-cols-2">
+      <li v-for="grant in grants" :key="grant.id.toString()">
+        <!-- ToDo: Raised data -->
+        <!-- ToDo: Grant image -->
+        <GrantCard
+          :id="grant.id"
+          :name="grantMetadata[grant.metaPtr].name ?? ''"
```
carletex, line 10: Had to this to avoid `error TS2322: Type 'string | undefined' is not assignable to type 'string'.` Not sure if there is a better alternative.

app/src/views/GrantRegistryList.vue
```diff
@@ -1,31 +1,83 @@
...
+        },
+        {
+          label: 'shuffle',
+          action: () => {
+            console.log('shuffle');
+          },
+        },
+      ],
+    },
+  ];
```
carletex, line 61: We don't have the grant date on the metadata yet, so I'm skipping this for now. (I can remove the "Sort" filterNav from the UI)

carletex (commented):
Ready for review. Left out all the things we discussed yesterday and left a few ToDo comments.

app/src/components/GrantCard.vue
```diff
@@ -0,0 +1,65 @@
...
+      </div>
+    </div>
+    <figcaption class="mt-4">
+      <div class="text-grey-500 font-medium truncate">{{ name }}</div>
+      <div class="flex justify-between">
+        <span class="text-grey-400"
+          >by
+          <a
+            class="text-grey-500 underline"
+            :href="`https://etherscan.io/address/${ownerAddress}`"
```
carletex, line 25: This ok? Or should check the `network` we are in?

gdixon: Check out [getEtherscanUrl](https://github.com/dcgtc/dgrants/blob/126cd0ce9faf9d4e0a3c02d5083743166a9ac1b2/app/src/utils/utils.ts#L112) in utils - we could repurpose that fn to accept an additional prop for `tx`/`address`/`token` and that will handle the network part 🙌

gdixon: Check out getEtherscanUrl in utils - we could repurpose that fn to accept an additional arg for tx/address/token and that will handle the network part 🙌

carletex: Oh nice @gdixon Something like this? https://github.com/dcgtc/dgrants/pull/184/commits/275c1c2c25332c7bc2b517ae62f179e6d73e0bdb

I used a third argument (`group`) with a default value to avoid breaking changes, but I could also update the calls to that function.

apbendi (commented):
Hey @carletex, looking awesome! A few things I noticed:

1. The sort menu hovers behind the logo image:

Screen Shot 2021-09-09 at 12 59 32 PM

2. There's no padding on the text in mobile mode:

Screen Shot 2021-09-09 at 1 01 10 PM

3. I think we still want a "Create Grant" button on this page, like the one shown here (taken from the homepage design). It should take the user to `/dgrants/new`

Screen Shot 2021-09-09 at 1 09 50 PM

app/src/components/GrantList.vue
mds1: Are you saying you needed the `?? ''` part of `"grantMetadata[grant.metaPtr].name ?? ''"` to avoid that error? This seems like a good solution to me! Another alternative is changing the prop type of `GrantCard` to `string | undefined` and handling undefined values there, but this works 👌

app/src/components/BaseTabs.vue
```diff
@@ -0,0 +1,31 @@
+<!-- BaseTabs
+a navigation bar that can have various nav items with optional (INT) numbers,
```
mds1, line 2: INT = integer? If so maybe just write "optional integer numbers" for clarity

app/src/components/BaseTabs.vue
```diff
@@ -0,0 +1,31 @@
...
+where one is the "active" by apply "border-b-2" class on it.
+-->
+
+<template>
+  <div class="px-4 md:px-12 py-6 border-b border-grey-100 flex flex-wrap gap-x-10 gap-y-2">
+    <span class="tab active">All (543)</span>
+    <span class="tab">Infra Tech (184)</span>
+    <span class="tab">dApp Tech (184)</span>
```
mds1, line 10: What is the plan for un-hardcoding these numbers?

app/src/components/BaseTabs.vue
```diff
@@ -0,0 +1,31 @@
...
+  </div>
+</template>
+
+<script lang="ts">
+import { defineComponent } from 'vue';
+
+export default defineComponent({
+  name: 'BaseTabs',
+  components: {},
```
mds1, line 19: Let's remove this line since there's no components being imported

app/src/components/GrantCard.vue
```diff
@@ -0,0 +1,65 @@
...
+// --- Methods and Data ---
+import { BigNumber } from 'ethers';
+import { formatAddress, getEtherscanUrl, pushRoute } from 'src/utils/utils';
+// --- Icons ---
+import { Cart2Icon as CartIcon } from '@fusion-icons/vue/interface';
+
+export default defineComponent({
+  name: 'GrantCard',
+  props: {
+    id: { type: BigNumber, required: true },
```
mds1, line 53: Let's use `BigNumberish` as the type and then when using the `id` you can do `BigNumber.from(id).toString()`, etc. In general you shouldn't use `BigNumber` as an input type because it's an ethers specific type, but you can/should:
- use `BigNumberish` as an input type and handle the input type
- use `BigNumber` as a return type (not applicable here, but worth mentioning)

app/src/components/GrantList.vue
```diff
@@ -1,85 +1,44 @@
...
+          :ownerAddress="grant.owner"
+          raised="100"
+          imgurl="https://s3-alpha-sig.figma.com/img/901c/c3ae/ac690f9a43a7634d9d198cf9057c4454?Expires=1632096000&Signature=PnDBcqyesPYCxYzyRTGn-imh2ktN8ImD4GIrtJKoBrpooTX6Uqae5ogcpWdMpn6CyZnfDAnGuA2XYokhIpLTbLTriqd5Ifiiv2zbVHHQlUwQQ8fk4k9gpAj17CTBJGS0Ldqv6pn9MvSTzXKaLa~FjSiX2AAeWgWz1Vy9OJ6KFahsmhfLXlGEmBDr1Dx016AMvWH5DhKRLY4XbpfJz5zJy6zEHbV7GZ6uxop4d-pIk9KvTPt3kjJEsmKRdVxxy8XC8zN2cIFMEQEWJDRe64AekEl5ABzlTN~9o8NYS6~q5ObH3BlwQuvUhPLNvw0Atjwc9DX3yWuo-cAOZISl5sboOg__&Key-Pair-Id=APKAINTVSUGEWH5XD5UA"
```
mds1, line 13: This should be pulled from the GrantMetadata, looks like it's still hardcoded

types/src/app.d.ts
```diff
@@ -37,3 +37,6 @@ export type FilterNavButton = {
   label: string; // label to display in the button
   action?: (payload: MouseEvent) => void; // action to take when we click the button
 };
+
+// type of entities on Etherscan.
+export type EtherscanGroup = 'tx' | 'token' | 'address';
```
mds1, line 42: Per the guidance in https://github.com/dcgtc/dgrants/discussions/114, this type definition should live in the `app` package instead of the `types` package. In your defense @carletex, it looks like we haven't been following that guidance well haha

app/src/components/GrantList.vue
carletex: Cool! Yes, that's what I meant.

app/src/components/GrantList.vue
carletex: The image URL is not coming (yet, I guess) in the GrantMetadata. I left a couple of ToDo comments above the component.

app/src/components/BaseTabs.vue
carletex: I initially added this file as it was since I was showing this hardcoded "categories". But after weds call we decided to remove those categories filter for now. So I've just removed this file for now if that's ok.

app/src/components/GrantCard.vue
carletex: Got it! 

When using it, it seems that `id.toString()` works fine. What is the goal of using `BigNumber.from(id).toString()`? (I changed it anyway)

types/src/app.d.ts
carletex: haha ok! Didn't know about that guide. Thanks!

carletex:
Hey @apbendi ! Thank for the UI review, I missed those, sorry.

> 1. The sort menu hovers behind the logo image
> 2. There's no padding on the text in mobile mode:
> 3. I think we still want a "Create Grant" button on this page, like the one shown here (taken from the homepage design). It should take the user to /dgrants/new

All should be fixed now!

thelostone-mc (approved):

app/src/components/GrantCard.vue
mds1: You are right that numbers, strings, BigInt, and BigNumber all have a `.toString()` method, so `id.toString()` will pretty much always work. Using `BigNumber.from(id).toString()` instead basically forces the value to be an integer, since `BigNumber.from(id)` will throw if you pass in decimals. For this use case either approach is probably ok

app/src/components/GrantList.vue
mds1: Our `GrantMetadata` type has an optional `logoURI` property, and we do resolve grant metadata currently, so we should be able to use the actual grant image now right? And we should just fallback to `placeholder_grant.svg` if the grant metadata has no `logoURI`

carletex: You're completely right. I was just checking the actual test metadata I was getting, but didn't notice the `GrantMetadata.logoURI` property.

Fixed in https://github.com/dcgtc/dgrants/pull/184/commits/7299102e963653726b6b0e9721df09fa71267c6f

Thanks Matt!

mds1: LGTM, thanks!

mds1 (approved):
I haven't verified the fixes for things @apbendi mentioned, but code-wise looks ready 👍️

apbendi (approved):
Looking great! Nice job @carletex

PR #185, opened 2021-09-08
improved grants detail view

scco:
- rewrote that grantdetailrow from scratch for slicker code and especially for bens concerns regarding to big grant-image  
- removed v-slot from GrantDetailsRow for share/edit buttons as i outsourced these to a dedicated navbar in the GrantRegistryGrantDetail.vue for better scalability of more interactions and a simpler codebase.

https://user-images.githubusercontent.com/569641/132556626-74e6717a-86bd-4c05-83a0-5367fb25b65b.mov

### DIFF
--- a/app/src/components/GrantDetailsRow.vue
+++ b/app/src/components/GrantDetailsRow.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="border-b border-grey-100 grid grid-cols-1 md:grid-cols-2 gap-x-8">
     <!--grid:left (img)-->
-    <div class="bg-pink">
+    <div>
       <img class="shadow-light object-cover h-full" :src="logoURI || '/src/assets/placeholder_grant.svg'" />
     </div>
 
--- a/app/src/components/GrantDetailsRow.vue
+++ b/app/src/components/GrantDetailsRow.vue
@@ -1,89 +1,79 @@
 <template>
-  <div class="border-b border-grey-100">
-    <div class="grid grid-cols-4 gap-0">
-      <!-- img -->
-      <div class="col-span-4 md:col-span-2 shadow-light flex">
-        <img class="m-auto" :src="logoURI || '/src/assets/logo.svg'" />
-      </div>
-      <div class="col-span-4 md:col-span-2 grid grid-cols-12 pt-8">
-        <!-- raised, contract, round, matching -->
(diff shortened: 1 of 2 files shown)


thelostone-mc:
@scco is this how you wanted it cause it looks a bit broken in terms of UI when i see it ?

image

mds1:
A small issue, but like @thelostone-mc said it looks odd to have the "Matching" word wrap at that middle-ish width. Wider and narrower both look good. One idea is a small "Match amounts" header above the two rows?

app/src/components/GrantDetailsRow.vue
```diff
@@ -1,89 +1,79 @@
...
+            </span>
+          </div>
+        </div>
+
+        <!--subgrid:right ( donate button, matching example )-->
+        <div class="pr-12">
+          <!-- button -->
+          <div class="text-right xl:flex xl:justify-end">
+            <button v-if="isInCart(grant?.id)" @click="removeFromCart(grant?.id)" class="btn in-cart btn-primary">
+              <CartIcon class="icon-small" />Remove
```
mds1, line 52: The teal/green color for the remove button feels a bit confusing since removing from cart feels like a red or neutral color. IMO "Add" should be teal and "Remove" dark grey or red, or both Add and Remove should be dark grey. But @scco you're the designer here so if you disagree that's fine with me

scco:
thanks for the feedback. 

i agree its option to simplify to a 2 grid layout, it also looks good. ( and thankfully have a slicker codebase again )

please do not forget all this will anyway change over and over within next months as our app grows in features, so thats why i am aiming for a lot whitespace and easy enhanceable areas ! it might look a bit thin for you - but me as a designer likes this a lot - as its beautifull, open and elegant. please trust me on this.

regarding the button class "in-cart" what turns the background of a button to green (green-light it) : 

this button have 2 contexts :  
1. it indicates if something is IN CART ( whats positive = green)
2. it indicates a "REMOVE" function ( what could but do not have to be red )

the first meaning is more important, and i do not wana emphasize on a negative remove thing .

as i use this button and class "in-cart" also in the LISTVIEW ( see my ui-branch ... ) it definitely makes sense and feels good. 
so lets just continue and code the list view with this "in-cart" class ... , then see how it feels in context.

mds1 (approved):
Looks great, nice work @scco!

apbendi (approved):
Looks great, thanks @scco 👍

PR #188, opened 2021-09-09
InputRow Component + Modify BaseInput / BaseTextarea + Rework CreateGrant UI

scco:
- Adding InputRow.vue
- Reworking BaseInput.vue / BaseTextarea.vue
- Apply these Components in /dgrants/new ( GrantRegistryNewGrant.vue )

This is a crucial PR what allow us to build forms as in design planed !
It will work nicely together with all the other components that are still waiting in my ui-branch ... 

An `<InputRow>` "wrapps" a v-slot for what ever input ( e.g. a `<BaseSelect>` or even vanilla `<input>` or what ever other `<div>stuff</div>` combination ) with its related label, description and a delete-icon (for later needed in other interfaces)

to achieve this, i had to remove the description and label from the `<BaseSelect>` & `<BaseTextarea>`
what shrinks down their html/css structure. i redesigned the error messages in these as well. i also removed that warning-icon part, as i think it was not visibible or planed to be in.

With this change we can build forms with this ROWS but also can be flexible and just use basic input fields like `<BaseSelect>` in a stand alone way or other wrappers like i do in in the cart.

To see all this in action i reworked the GrantRegistryNewGrant.vue to give an impression and first test , how all this changes should look. 

My hope is, to get this PR in very quick before its outdated again and then continue develop other inputs  like dragArea , new BaseSelect, checkboxes  later on + obviously build pother potential forms.

To see, switch to this branch, goto GrantRegistryNewGrant.vue  and edit some options like : 

1. see how width="w-full" looks vs "w-1/2" or "w-1/3" 
2. see what happens when you edit :deleteable="false" to :deleteable="true"
3. see what happens when you edit :intended="false" to :intended="true"

Definitely also checkout before and after to see and understand why this a good step into the right direction.

Bildschirmfoto 2021-09-09 um 06 10 25

ps:  for very later : 
- would love to see growing textarea (like facebook)
- delete a row functionality needs to be programmed ( needed for e.g. custom-links, team-members and so on somewhere in the future - for now not needed )

### DIFF
--- a/app/src/components/BaseInput.vue
+++ b/app/src/components/BaseInput.vue
@@ -3,7 +3,7 @@
     <input
       v-model="val"
       @input="onInput"
-      :class="[!isValid ? '' : '', customcss ? customcss : '']"
+      :class="[customcss ? customcss : '']"
       :id="id"
       :name="id"
       :required="required"
--- a/app/src/components/BaseInput.vue
+++ b/app/src/components/BaseInput.vue
@@ -24,7 +24,6 @@ import { computed, defineComponent, ref } from 'vue';
 
 export default defineComponent({
   name: 'BaseInput',
-  components: {},
   props: {
     // --- Required props ---
     modelValue: { type: [String, Number], required: true, default: undefined }, // from v-model, don't pass this directly
--- a/app/src/components/BaseInput.vue
+++ b/app/src/components/BaseInput.vue
@@ -1,42 +1,34 @@
 <template>
   <div :class="width">
-    <label :for="id" class="block text-sm font-medium text-gray-700"> {{ label }} </label>
-    <p :for="id" class="text-gray-500 text-sm">{{ description }}</p>
-    <div>
-      <input
-        v-model="val"
-        @input="onInput"
-        :class="[!isValid ? 'border-pink text-pink' : '', customcss ? customcss : '']"
-        :id="id"
-        :name="id"
-        :required="required"
-        :type="type"
-        :readonly="readonly"
-        :disabled="disabled"
-        :placeholder="placeholder"
-      />
-      <div v-if="!isValid" class="absolute inset-y-0 right-0 pr-3 flex items-center pointer-events-none h-12 w-12">
-        <WarningIcon class="icon stroke-pink" />
-      </div>
-      <p v-if="!isValid" class="text-xs text-red-600" :id="`${id}-error`">{{ errorMsg }}</p>
-      <p :for="id" class="text-grey-400 mt-2 text-sm">{{ description }}</p>
-    </div>
+    <input
+      v-model="val"
+      @input="onInput"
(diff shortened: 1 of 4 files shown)


thelostone-mc (approved):

app/src/components/BaseInput.vue
```diff
@@ -1,42 +1,34 @@
...
 </template>
 
 <script lang="ts">
 import { computed, defineComponent, ref } from 'vue';
-import { WarningIcon } from '@fusion-icons/vue/interface';
 
 export default defineComponent({
   name: 'BaseInput',
-  components: { WarningIcon },
+  components: {},
```
mds1, line 27: Let's delete this line since there are no components

app/src/components/BaseInput.vue
```diff
@@ -1,42 +1,34 @@
...
-      <div v-if="!isValid" class="absolute inset-y-0 right-0 pr-3 flex items-center pointer-events-none h-12 w-12">
-        <WarningIcon class="icon stroke-pink" />
-      </div>
-      <p v-if="!isValid" class="text-xs text-red-600" :id="`${id}-error`">{{ errorMsg }}</p>
-      <p :for="id" class="text-grey-400 mt-2 text-sm">{{ description }}</p>
-    </div>
+    <input
+      v-model="val"
+      @input="onInput"
+      :class="[!isValid ? '' : '', customcss ? customcss : '']"
```
mds1, line 6: Seems like we don't use the `isValid` check here, so can we make this:
```suggestion
      :class="[customcss ? customcss : '']"
```

app/src/components/BaseTextarea.vue
```diff
@@ -1,39 +1,31 @@
 <template>
-  <div class="grid grid-cols-12 py-5 items-center gap-m-8 border-b border-grey-100">
-    <div class="col-span-12 md:col-span-3 mb-3 md:mb-0 grid-rows-3">
-      <label :for="id" class="text-grey-400"> {{ label }}: </label>
-    </div>
+  <div :class="width">
+    <textarea
+      v-model="val"
+      @input="onInput"
+      :class="[!isValid ? '' : '']"
```
mds1, line 6: Similar comment, seems we don't use `isValid` and can delete this whole line

app/src/components/BaseTextarea.vue
```diff
@@ -1,39 +1,31 @@
...
 </template>
 
 <script lang="ts">
 import { computed, defineComponent, ref } from 'vue';
-import { WarningIcon } from '@fusion-icons/vue/interface';
 
 export default defineComponent({
   name: 'BaseTextarea',
-  components: { WarningIcon },
+  components: {},
```
mds1, line 28: Similar comment, can delete this line

scco:
fixed the suggestions right now

mds1 (approved):

PR #189, opened 2021-09-09
chore: GrantRoundUI

thelostone-mc:
#### Description 

Implements UI for GrantRound page

#### Pending

- [ ] Share functionality  (this not sure what to do here)
- [x] Lint fixes (@mds1 could use a hand here)
- [x] Replace BaseInput with Input from @scco PR (waiting on #188 to get in )

#### Blockers

Needs rebase against https://github.com/dcgtc/dgrants/pull/188

#### Testing

image

### DIFF
--- a/contracts/scripts/app.ts
+++ b/contracts/scripts/app.ts
@@ -61,7 +61,7 @@ const fixtureRound = async (registry: Contract, manager: Contract) => {
   let startTime = Math.floor(new Date().getTime() / 1000); // time in seconds
   startTime = await setNextBlockTimestamp(startTime + 200);
 
-  const endTime = startTime + 86400; // 1 day later
+  const endTime = startTime + (86400 * 5); // 5 days later
 
   // GrantRound Argument
   const metadataAdmin = '0x34f4E532a33EB545941e914B25Efe348Aea31f0A';
--- a/contracts/scripts/app.ts
+++ b/contracts/scripts/app.ts
@@ -67,7 +67,7 @@ const fixtureRound = async (registry: Contract, manager: Contract) => {
   const metadataAdmin = '0x34f4E532a33EB545941e914B25Efe348Aea31f0A';
   const payoutAdmin = '0x06c94663E5884BE4cCe85F0869e95C7712d34803';
   const matchingToken = tokens.dai.address;
-  const metaPtr = 'https://ipfs-dev.fleek.co/ipfs/bafybeiejov726khhgtcr4hzjzohoo5x4xzxp7c3nng55z4iv4xj6fpan2a';
+  const metaPtr = 'https://ipfs-dev.fleek.co/ipfs/bafybeib5r75zbrzavbskwywct42alawcwo7d4dlyipimvnjy23hfkqhham';
   const minContribution = ethers.constants.One;
 
   const tx = await manager.createGrantRound(
--- a/app/src/components/GrantDetailsRow.vue
+++ b/app/src/components/GrantDetailsRow.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="border-b border-grey-100 grid grid-cols-1 md:grid-cols-2 gap-x-8">
+  <div class="border-b border-grey-100 grid grid-cols-1 md:grid-cols-2 gap-x-14">
     <!--grid:left (img)-->
     <div>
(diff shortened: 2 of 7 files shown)


apbendi (commented):
Hey @thelostone-mc, looking awesome!! A couple thoughts/questions:

- [x]  Why is this token address shown/editable on the contribution page? The contribution has to be made in the round's `matchingToken`, right? So I think we would just want to display the token name/symbol, maybe with the address and a link to the etherscan token tracker or something. But I don't see why we'd want it to be an editable field.

Screen Shot 2021-09-09 at 1 24 53 PM

Another nit on this form: why is the tip text (like "The number of tokens to contribute") show above and below the field?

- [x] Upon donation to the Round, would it be possible to a show a transaction status screen like on checkout? (If this is a bigger job, we can make an issue and track it separate from this PR).

Screen Shot 2021-09-09 at 1 27 53 PM

- [ ] I also noticed I can still navigate to this old contribution form. This should probably be remove now, right?
Screen Shot 2021-09-09 at 1 27 12 PM

**Aditya:** Oh i've that would be retired as a part when that page is rewritten

The main round info page looks great 👍

apbendi:
One more nit @thelostone-mc, I don't think we should populate any value in the token amount field. Just leave it empty and let the user fill it in, I think.

contracts/scripts/app.ts
```diff
@@ -67,7 +67,7 @@ const fixtureRound = async (registry: Contract, manager: Contract) => {
   const metadataAdmin = '0x34f4E532a33EB545941e914B25Efe348Aea31f0A';
   const payoutAdmin = '0x06c94663E5884BE4cCe85F0869e95C7712d34803';
   const matchingToken = tokens.dai.address;
-  const metaPtr = 'https://ipfs-dev.fleek.co/ipfs/bafybeiejov726khhgtcr4hzjzohoo5x4xzxp7c3nng55z4iv4xj6fpan2a';
+  const metaPtr = 'https://ipfs-dev.fleek.co/ipfs/bafybeib5r75zbrzavbskwywct42alawcwo7d4dlyipimvnjy23hfkqhham';
```
mds1, line 70: Why does this hash not begin with the `Qm` multihash prefix that the other `metaPtr`s in the `fixtureGrants` method do? cc @corydickson

thelostone-mc: I think this is cause I host the metaPtr in my fleek account

scco:
im wondering why we do not use the regular cart system for this. is this technical not possible ?

apbendi:
@scco 

> im wondering why we do not use the regular cart system for this.

I think donating to the matching funds is a separate enough concept from donating to a grant that it's actually better to have a different flow for the two. Having them combined together would lead to confusion, is my guess. It would be unusual for a user to want to donate to a bunch of Rounds at the same time, I think.

apbendi (commented):
Hey @thelostone-mc, thanks for making those tweaks. Looking great. Sorry to continue being annoying, but here's a couple things I noticed:

- [x] The contribution form does not validate your balance, instead it errors silently if you don't have enough of the token and attempt to contribute:

Screen Shot 2021-09-10 at 10 33 25 AM

- [x] This is a nit and I'm sorry to harp on it but I'd really love if we could remove this prepopulated amount 😅 

Screen Shot 2021-09-10 at 10 30 25 AM

mds1:
> The contribution form does not validate your balance, instead it errors silently if you don't have enough of the token and attempt to contribute:

@thelostone-mc In the `checkout()` method in `cart.ts` we handle this with an `assertSufficientBalance` method. You can move that into the utils folder, or just export it from the cart, and use that same logic

thelostone-mc:
Oh noiceee didn't think of that! lol i was testing with a wallet having 5000 DAI on local :P

thanks @apbendi + @mds1

contracts/scripts/app.ts
mds1: Sorry what exactly do you mean? I'm still not too familiar in fleek but curious what you're doing differently than the app. Is it _bad_ in some way that this hash is different? And I thought all IPFS hashes started with Qm currently so trying to understand what's going on here!

apbendi (approved):

contracts/scripts/app.ts
thelostone-mc: I don't think so ! @corydickson any inputs here ?

PR #192, opened 2021-09-09
Contract cleanup

### LINKED ISSUE #133: Final Pre-MVP Contract Review
- [ ] Internal
    - [ ] Test coverage -> 100%
- [ ] 3rd Party

mds1:
Closes #133

### DIFF
--- a/contracts/contracts/GrantRoundManager.sol
+++ b/contracts/contracts/GrantRoundManager.sol
@@ -91,6 +91,7 @@ contract GrantRoundManager is SwapRouter {
     GrantRound _grantRound = new GrantRound(
       _owner,
       _payoutAdmin,
+      registry,
       donationToken,
       _matchingToken,
       _startTime,
--- a/contracts/contracts/GrantRoundManager.sol
+++ b/contracts/contracts/GrantRoundManager.sol
@@ -21,23 +21,24 @@ contract GrantRoundManager is SwapRouter {
   /// @notice Address of the ERC20 token in which donations are made
   IERC20 public immutable donationToken;
 
+  /// @dev Used for saving off swap output amounts for verifying input parameters
   mapping(IERC20 => uint256) internal swapOutputs;
 
+  /// @dev Used for saving off contribution ratios for verifying input parameters
   mapping(IERC20 => uint256) internal donationRatios;
 
   /// @dev Scale factor on percentages when constructing `Donation` objects. One WAD represents 100%
   uint256 internal constant WAD = 1e18;
 
   /// --- Types ---
-  /// @notice Defines the total `amount` of the specified `token` that needs to be swapped to `donationToken`. If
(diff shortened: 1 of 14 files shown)


app/src/views/GrantRound.vue
```diff
@@ -207,7 +207,7 @@ function useGrantRoundDetail() {
   const form = computed<{ token: string; amount: string }>(() => {
     return {
       token: grantRound.value.matchingToken.address,
-      amount: String(grantRound.value.minContribution),
+      amount: '0',
```
mds1, line 210: @gdixon @thelostone-mc Can you verify this change is ok / what we want?

thelostone-mc: Yup this is right

thelostone-mc (approved):

contracts/contracts/GrantRoundManager.sol
```diff
@@ -69,39 +70,32 @@ contract GrantRoundManager is SwapRouter {
...
     uint256 _endTime,
-    string memory _metaPtr,
-    uint256 _minContribution
+    string memory _metaPtr
   ) external {
     require(_matchingToken.totalSupply() > 0, "GrantRoundManager: Invalid matching token");
     GrantRound _grantRound = new GrantRound(
       _owner,
       _payoutAdmin,
-      _registry,
```
apbendi, line 98: Rather than removing this parameter from the `GrantRound` constructor, I think we want to remove it from the `createGrantRound` method parameters, and instead pass the storage variable `registry` in the `GrantRoundManager` through to the `GrantRound`. The Round should have some publicly accessible hold on the registry it refers to, otherwise the `grantRoundIds` it refers to in its metadata would be meaningless.

mds1: Hmm, why does the registry need to be part of the `GrantRound` on chain? It can't hurt, but just want to make sure we're on the same page. Some thoughts:
- If the `grantRoundIds` are defined off-chain in the metadata, you can put the registry there also, and if we want to make verifying that easier we should index the grant round address in the `GrantRoundCreated` event (maybe we should index that anyway)
- We currently have an on-chain link (in the form of events) to go from GrantRoundManager to the GrantRounds it's created
- The GrantRound contract does not need to know anything about the GrantRoundManager on-chain as of now

corydickson (approved):

apbendi (approved):

contracts/contracts/GrantRoundManager.sol
mds1: Decision from today's sync was that we'll keep the registry parameter here as described in @apbendi's original comment. Reimplemented in 41e22348c24784cd2af68404b9b8fe4e9a87f030. Added ~11k gas to the `createGrantRound` cost for reference

PR #193, opened 2021-09-10
Adding base dockerfile for building in Fleek

phutchins:
This is the base docker container that we are using to build in Fleek.

To build this I use:
```
docker build -f ./app/Dockerfile -t phutchins/volta-base:0.0.1 .
```

then to push to docker hub:
```
docker push phutchins/volta-base:0.0.1
```

We should and will create an org for dGTC and push to that at some point instead of my org.

### DIFF
--- a/app/Dockerfile
+++ b/app/Dockerfile
@@ -1,0 +1,9 @@
+FROM node:slim
+
+RUN apt update
+RUN apt install git -y
+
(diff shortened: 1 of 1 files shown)


apbendi (approved):
LGTM!

gdixon (approved):

Issue #176, opened 2021-09-03
Implement Matching Donation Token Amount In Cart


apbendi:
- [x] Estimated amount for total donations in donation token
- [x] Estimated amount for matching tokens in donation token (Depends on #175)
- [x] Use the ratio from uniswap quote as prices in each case `quoteExactInput`

PR #194, opened 2021-09-10
feat: show DAI-equivalent donation amount on cart page

mds1:
Resolves two-thirds of #176

The "17,278.11 DAI" portion of the screenshot below is what this PR adds. Prior to this PR, that just said "TODO"

For now we hardcode DAI but in the future we'll need to show this based on the GrantRoundManager's `donationToken`. I thought about creating an issue for this, but I a meta "Post-POC improvements" issue would be cleaner, to avoid having tons of small issues for each TODO + simplify triaging/prioritizing after the POC. @apbendi thoughts on this? 

image

### DIFF
--- a/app/src/store/cart.ts
+++ b/app/src/store/cart.ts
@@ -13,6 +13,7 @@ import {
   ETH_ADDRESS,
   GRANT_ROUND_MANAGER_ABI,
   GRANT_ROUND_MANAGER_ADDRESS,
+  SUPPORTED_TOKENS,
   SUPPORTED_TOKENS_MAPPING,
   WAD,
   WETH_ADDRESS,
@@ -60,7 +61,8 @@ const toHex = (val: BigNumberish) => BigNumber.from(val).toHexString();
 
 // --- State ---
 const lsCart = ref<CartItemOptions[]>([]); // localStorage cart
-const cart = ref<CartItem[]>([]);
+const cart = ref<CartItem[]>([]); // source of truth for what's in the user's cart
+const quotes = ref<Record<string, number>>({}); // mapping from token address to DAI exchange rate, i.e. multiple token quantity by the exchange rate to get the value in DAI
 
 // --- Composition function for state management ---
 export default function useCartStore() {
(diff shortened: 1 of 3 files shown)


apbendi (approved):
Looks good to me. My only question/feedback is whether we should add something like a "~" to indicate it's an estimate, or maybe something in the copy? Something to discuss, but PR looks good.

apbendi:
> a meta "Post-POC improvements" issue would be cleaner, to avoid having tons of small issues for each TODO + simplify triaging/prioritizing after the POC. @apbendi thoughts on this?

I actually think it's OK to go ahead and just make individual issues for stuff like this @mds1. Might be easier to assign them off to folks that way, and we can always triage the issues with the "POC" tag on for now to avoid noise before the round.

mds1:
> I actually think it's OK to go ahead and just make individual issues for stuff like this @mds1. Might be easier to assign them off to folks that way, and we can always triage the issues with the "POC" tag on for now to avoid noise before the round.

Created https://github.com/dcgtc/dgrants/issues/215 to track this

gdixon (approved):
LGTM!

app/src/store/cart.ts
```diff
@@ -265,6 +267,28 @@ export default function useCartStore() {
...
+   */
+  async function fetchQuotes() {
+    // TODO use multicall for better performance + fewer RPC requests
+    const _quotes = await Promise.all(
+      SUPPORTED_TOKENS.map(async (token) => {
+        if (token.symbol === 'DAI' || token.symbol === 'USDC') return { token, rate: 1 };
+        const path = SWAP_PATHS[<keyof typeof SWAP_PATHS>token.address];
+        const amountIn = parseUnits('1', token.decimals); // for simplicity, use a value of 1 token for getting quotes
+        const amountOut = await quoteExactInput(path, amountIn); // as raw BigNumber
+        return { token, rate: Number(formatUnits(amountOut, token.decimals)) };
```
thelostone-mc, line 284: Lol looks like `formatUnits` wasn't imported :P 
Fixing it up on 10eb0de @mds1

PR #198, opened 2021-09-10
Grant Logo Metadata

### LINKED ISSUE #119: Add Grant Logo To Metadata
This issue is dependent on #86. It expands the metadata stored for a Grant to include a Grant Logo, and displays the grant logo on the site in various contexts where it should be.

Incorporation of the logos should directionally follow the [figma designs](https://www.figma.com/file/sUQSYyPfooqTcVdajfZm0t/dgrants?node-id=92%3A80), but does not need to be precise. A proper re-working of the relevant components will occur with #112.

- [x] When a user creates a Grant, the grant form should have an optional field that allows them to select a logo image, and validates:
    - [x] dimension : 1920x1080 ( 16:9 )
    - [x] formats : png or svg . no jpg as its to big and not a format for logos but for photos .
    - [x] filesize : max 200kb ( thats an edgecase already, regular it should be like 15kb - 80kb )
- [x] Before the grant is created:
    - [x] The logo image should be uploaded to IPFS first
    - [x] The cloudflare IPFS gateway URL using the logo CID should be included in the Grant Metadata using the field `logo`. See the schema example below
    - [x] The rest of grant creation process now continues, i.e. the grant metadata is uploaded to IPFS, then the user submits the transaction to create the Grant, which includes the metadata CID
- [x] Grant logos should be displayed in Grant list contexts, i.e. on the registry page
- [x] The Grant logo should be displayed on the Grant Detail Page
- [x] Fetching of Grant logos should be non-blocking to the UI, and fail gracefully with a placeholder in each instance where it doesn't succeed.
- [x] Our fixture data should be updated to metadata that includes the logo field pointing to valid images on IPFS

#### Updated Grant Metadata Schema

```json
{
    "name": "My Grant",
    "description": "There are many Grants, but this one is mine",
    "logo": "https://cloudflare-ipfs.com/ipfs/QmaKj8z3icu3UFqXK4NXy7Mcch9kCPZJA128gP4pa4fQQM"
}
```

mk1r:
Resolves #119
Screen Shot 2021-09-10 at 10 50 41 AM
Screen Shot 2021-09-10 at 10 53 15 AM

### DIFF
--- a/app/src/utils/utils.ts
+++ b/app/src/utils/utils.ts
@@ -65,6 +65,11 @@ export function isValidUrl(val: string) {
   return (val.includes('http') || val.includes('https')) && val.includes('://') && val.includes('.');
 }
 
+// Returns true if the provided address is valid (TODO support ENS)
+export function isValidAddress(val: string | undefined) {
+  return val && isAddress(val);
+}
+
 // Returns true if the provided website is a valid option
 export function isValidWebsite(val: string | undefined) {
   return val ? isValidUrl(val) : true;
@@ -82,9 +87,21 @@ export function isValidTwitter(val: string | undefined) {
   return /^[a-zA-Z0-9_]{1,15}$/.test(handle);
 }
 
-// Returns true if the provided address is valid (TODO support ENS)
-export function isValidAddress(val: string | undefined) {
-  return val && isAddress(val);
+// Returns true if the provided logo is valid
+export async function isValidLogo(file: File | undefined) {
+  if (!file) return true;
+  if (file.size > 200000) return false;
+  if (!(file.type.includes('image/png') || file.type.includes('image/svg'))) return false;
+
+  const dimensions = await new Promise<{ width: number; height: number }>((resolve) => {
+    const img = new Image();
+    img.src = window.URL.createObjectURL(file);
+    img.onload = () => {
+      resolve({ width: img.width, height: img.height });
+    };
+  });
+
+  return dimensions.width === 1920 && dimensions.height === 1080;
 }
 
 // --- Tokens ---
--- a/app/src/components/BaseImageUpload.vue
+++ b/app/src/components/BaseImageUpload.vue
@@ -1,26 +1,28 @@
 <template>
   <div :class="width">
-    <label :class="['flex', 'flex-col', 'bg-white', 'items-center', 'block', 'px-6', 'py-8', 'border border-grey-400']">
+    <label class="flex flex-col bg-white items-center block px-6 py-8 border border-grey-400 cursor-pointer">
       <span>Choose image file to upload</span>
-      <input type="file" @change="onFile" class="hidden" />
+      <input type="file" @input="onInput" class="hidden" />
     </label>
-    <div v-if="!isValid" class="absolute inset-y-0 right-0 pr-3 flex items-center pointer-events-none h-12 w-12">
-      <WarningIcon class="icon stroke-pink" />
+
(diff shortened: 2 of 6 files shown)


mk1r:
Remaining tasks:
1) figure out isValid, see Discord
2) allow for editing
3) update fixtures + rebase

apbendi:
Hey @mk1r, just tried this out. I clicked the button, selected a random JPEG file and clicked select. The app didn't seem to do anything, neither giving me an error nor showing me that the image had been selected. When I created the grant, it did not have an image.

Additionally, I was able to repeat the same process with a PDF, which probably shouldn't be selectable in the file picker at all.

I'm guessing this is failing because I'm not selecting a valid image, so a couple thoughts:

* The app shouldn't let me select invalid filetypes, e.g. PDF, TXT, etc...
* If I select an image that doesn't meet the requirements for size, I should be notified in some way
* The form should indicate what the requirements are, for example under the "Grant logo (optional)" label or inside the button

app/src/utils/utils.ts
```diff
@@ -59,6 +59,11 @@ export function isValidUrl(val: string) {
   return (val.includes('http') || val.includes('https')) && val.includes('://') && val.includes('.');
 }
 
+// Returns true if the provided address is valid (TODO support ENS)
+export function isValidAddress(val: string | undefined) {
+  return val && isAddress(val);
+}
```
mds1, line 65: Is the purpose of this wrapper around `isAddress` just to allow us to pass in values which may be undefined? If so let's just add in a comment since it seems like an unnecessary wrapper method otherwise

mk1r:
hey @apbendi yeah that's what I'm trying to fix with validation, it is validating images - they have to be under 200 kb, png, and 1920x1080

app/src/utils/utils.ts
mk1r: is this new? I don't think I made a change here, will change it to what it was before

apbendi:
hey @mk1r, I just tested it with an image of the correct size and it worked. However, there was not indication on the form itself that I'd successfully selected an image. Is this also related to the validation fix you are working on?

mk1r:
@apbendi should I specify that image upload went successfully? this is related somewhat to the validation UI stuff, there I want to have an error displayed if it's not successful - if an image is uploaded successfully, then create grant btn will become black, do I need to add a msg that says img / logo was uploaded successfully?

app/src/utils/utils.ts
```diff
@@ -81,9 +86,21 @@ export function isValidTwitter(val: string | undefined) {
   return /^[a-zA-Z0-9_]{1,15}$/.test(handle);
 }
 
-// Returns true if the provided address is valid (TODO support ENS)
-export function isValidAddress(val: string | undefined) {
-  return val && isAddress(val);
+// Returns true if the provided logo is valid
+export async function isValidLogo(file: File | undefined) {
+  if (!file) return true;
```
thelostone-mc, line 91: shouldn't this be false ? Or are we saying no file is true AKA no error is thrown ?

mds1: Yep the latter, we default to true otherwise you'd get an "invalid image" error as soon as the form loads

mk1r: it's true, because logo is optional so if no file is uploaded, logo is undefined

thelostone-mc (approved):

apbendi (approved):

PR #202, opened 2021-09-10
grant detail page tweaks

### LINKED ISSUE #173: Grant Detail Page Tweaks
- [x] Links to a given service should only be shown if there is data populated, otherwise remove them from the UI completely

Screen Shot 2021-09-03 at 10 16 06 AM

- [x] If no links are provided, remove the whole "Links" section

- [x] What should we display before a grant has any contributions? Right now it shows the titles and the page ends.  Show a "No contributions yet" message below.

Screen Shot 2021-09-03 at 6 58 14 AM

mk1r:
Resolves #173 

- links should not show undefined
- link section should be removed if no links
- contribution section should be removed if no contributions

Screen Shot 2021-09-10 at 1 14 44 PM

### DIFF
--- a/app/src/views/GrantRegistryGrantDetail.vue
+++ b/app/src/views/GrantRegistryGrantDetail.vue
@@ -50,60 +50,52 @@
       </div>
 
       <!-- LINKS -->
-      <SectionHeader
-        v-if="
-          isDefined(grantMetadata?.properties?.websiteURI) ||
-          isDefined(grantMetadata?.properties?.githubURI) ||
-          isDefined(grantMetadata?.properties?.twitterURI)
-        "
-        title="Links"
-      />
-      <div
-        v-if="
-          isDefined(grantMetadata?.properties?.websiteURI) ||
-          isDefined(grantMetadata?.properties?.githubURI) ||
-          isDefined(grantMetadata?.properties?.twitterURI)
-        "
-        class="px-4 md:px-12 py-8 border-b border-grey-100 flex flex-col gap-y-4"
-      >
-        <div v-if="isDefined(grantMetadata?.properties?.websiteURI)" class="flex gap-x-4">
-          <span class="text-grey-400">Website:</span>
-          <a :href="grantMetadata?.properties?.websiteURI" target="_blank">{{
-            grantMetadata?.properties?.websiteURI
-          }}</a>
-        </div>
+      <template v-if="areLinksDefined">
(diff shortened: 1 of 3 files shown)


gdixon (approved):
LGTM! Good catch on the typo 😅

mds1:
@mk1r Just a heads up I updated your description to say "Resolves #173" instead of "ref issue #173", since the former will automatically close the issue once this PR is merged. More info: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -43,16 +43,29 @@
...
+          isDefined(grantMetadata?.properties?.githubURI) ||
+          isDefined(grantMetadata?.properties?.twitterURI)
+        "
+        title="Links"
+      />
+      <div
+        v-if="
+          isDefined(grantMetadata?.properties?.websiteURI) ||
+          isDefined(grantMetadata?.properties?.githubURI) ||
+          isDefined(grantMetadata?.properties?.twitterURI)
```
mds1, line 65: I think it'd be cleaner and more readable to pull these `isDefined` checks into a computed property instead of redefining it twice in the template

mds1: Similarly, could wrap the whole LINKS section in one div that way you only need one `v-if` statement

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -74,11 +87,11 @@
       </div>
 
       <!-- CONTRIBUTIONS -->
-      <SectionHeader title="Contributions" />
-      <div>
+      <SectionHeader v-if="grantContributions.length > 0" title="Contributions" />
+      <div v-if="grantContributions.length > 0">
```
mds1, line 91: Wrap in a parent div so you only need one `v-if` statement?

app/src/views/GrantRegistryGrantDetail.vue
thelostone-mc: i'll fix this up

thelostone-mc:
@mk1r i threw in a commit to fix up the comments left by @mds1 ! merging this in now

PR #205, opened 2021-09-10
fix: change from history to hash routing for IPFS support

mds1:
I don't think we have an associated issue, but this changes the router from history to hash mode, where has mode is "Useful for web applications with no host (e.g. file://) or when configuring a server to handle any URL isn't an option"

Relevant docs here: https://next.router.vuejs.org/api/#parameters


apbendi (approved):
lgtm, working fine!

phutchins (approved):

PR #211, opened 2021-09-11
Richard grants round design issue

### LINKED ISSUE #209: GrantsRound Design Issue
Bildschirmfoto 2021-09-11 um 16 44 02

- use the codebase of `GrantDetailsRow.vue` for `GrantRoundDetailsRow.vue`
to ensure same style .

reference screenshot of current` GrantDetailsRow.vue` of  difference : 

Bildschirmfoto 2021-09-11 um 16 55 01

- add funds button : remove green from icon
- share will be "tweet" + tweet icon ( from fusion )
- description :  each <p> should be `<p class="text-intend"> `

scco:
closes #209

- fixed the .intent in index.css  { was somehow not working for me ...  }
- fixing positioning of GrantRoundDetailsRow text elements / button
- removing the green from Add Funds
- add Tweet Button Design
- generating a rudimentary but probably not working tweet-link ( i created an issue #212 ) for this

### DIFF
--- a/app/src/index.css
+++ b/app/src/index.css
@@ -51,7 +51,7 @@
 
 /* Add styles to components */
 @layer components {
-  .intent {
+  .text-indent {
     text-indent: 1em;
   }
 
--- a/app/src/index.css
+++ b/app/src/index.css
@@ -51,7 +51,7 @@
 
 /* Add styles to components */
 @layer components {
-  .indent {
+  .intent {
     text-indent: 1em;
   }
 
--- a/app/src/views/GrantRound.vue
+++ b/app/src/views/GrantRound.vue
@@ -114,21 +114,21 @@
     <div class="px-4 md:px-12 py-12 border-grey-100 flex flex-col gap-y-4">
       <div v-if="isDefined(grantRoundMetadata?.properties?.websiteURI)" class="flex gap-x-4">
         <span class="text-grey-400">Website:</span>
-        <a :href="grantRoundMetadata?.properties?.websiteURI" target="_blank">{{
+        <a :href="grantRoundMetadata?.properties?.websiteURI" target="_blank" class="link">{{
           grantRoundMetadata?.properties?.websiteURI
         }}</a>
       </div>
 
       <div v-if="isDefined(grantRoundMetadata?.properties?.governanceURI)" class="flex gap-x-4">
         <span class="text-grey-400">Governance:</span>
-        <a :href="grantRoundMetadata?.properties?.governanceURI" target="_blank">{{
+        <a :href="grantRoundMetadata?.properties?.governanceURI" target="_blank" class="link">{{
           grantRoundMetadata?.properties?.governanceURI
         }}</a>
       </div>
 
       <div v-if="isDefined(grantRoundMetadata?.properties?.twitterURI)" class="flex gap-x-4">
         <span class="text-grey-400">Twitter:</span>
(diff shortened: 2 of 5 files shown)


mds1 (approved):

app/src/views/GrantRound.vue
```diff
@@ -75,25 +75,37 @@
...
     </div>
 
     <!-- Description -->
     <SectionHeader title="Description" />
+
     <div class="border-b border-grey-100">
-      <p style="white-space: pre-line" class="intent px-4 md:px-12 py-24 mx-auto max-w-6xl">
-        {{ grantRoundMetadata?.description }}
-      </p>
+      <div class="px-4 md:px-12 py-12 mx-auto max-w-6xl">
```
thelostone-mc, line 104: This does mean everything would be not aligned right. 
I'd say keep white-space: preline until you've thrown in a CRLF / ENTER 

Right now line breaks as stored as `/n` and we don't want to store the entire HTML of description onto IPFS

app/src/index.css
```diff
@@ -56,7 +56,7 @@
 
 /* Add styles to components */
 @layer components {
-  .indent {
+  .intent {
```
thelostone-mc, line 59: this seems like a really generic name ! 
could we do a different name ?

app/src/views/GrantRound.vue
```diff
@@ -174,7 +186,7 @@ import {
 import { GrantRound, GrantRoundMetadata, Breadcrumb } from '@dgrants/types';
 
 // --- Icons ---
-import { ArrowToprightIcon as ShareIcon } from '@fusion-icons/vue/interface';
+import { TwitterIcon as TweetIcon } from '@fusion-icons/vue/interface';
```
thelostone-mc, line 189: ^ do we need the rename ? why not use TwitterIcon as is

app/src/views/GrantRound.vue
```diff
@@ -75,25 +75,37 @@
...
+        <!-- Tweet -->
+        <a
+          target="_blank"
+          :href="
+            'https://twitter.com/intent/tweet?text=' +
+            grantRoundMetadata?.name +
+            '&url=https://grants.gtcdao.net' +
+            $route.path
+          "
+          class="flex items-center gap-x-2 cursor-pointer group ml-auto"
```
thelostone-mc, line 87: Also add no `noreferrer noopener`

thelostone-mc (changes requested):
Left a few comments

app/src/views/GrantRound.vue
scco: changed to "whitespace-pre-line"  but later maybe revisit to have it as i want in design.

app/src/index.css
scco: i think i mixed up also   .intent with .indent ...  so "indent" or "indentation" is the right world in english.
but as we wana use this just for <p> with long TEXT  i renamed it 

.text-indent

app/src/views/GrantRound.vue
scco: im ok with keep it TwitterIcon. question of taste.

app/src/views/GrantRound.vue
scco: ok

scco:
added another fix to have links in the links section be underlined.

thelostone-mc (approved):

PR #216, opened 2021-09-13
bug: fix broken edit page

thelostone-mc:
#### Description 

image

### DIFF
--- a/app/src/views/GrantRegistryGrantDetail.vue
+++ b/app/src/views/GrantRegistryGrantDetail.vue
@@ -106,7 +106,7 @@
       <div v-if="selectedEdit == 0" class="text-left">
         <form class="space-y-5" @submit.prevent="saveEdits">
           <!-- Grant name -->
-          <InputRow :deleteable="false" :intended="false" text="">
+          <InputRow>
             <template v-slot:label>Title:</template>
             <template v-slot:input>
               <BaseInput
@@ -121,7 +121,7 @@
           </InputRow>
 
           <!-- Owner address -->
-          <InputRow :deleteable="false" :intended="false" text="has permission to edit the grant">
+          <InputRow text="has permission to edit the grant">
             <template v-slot:label>Owner Adddress:</template>
             <template v-slot:input>
               <BaseInput
@@ -136,7 +136,7 @@
           </InputRow>
 
           <!-- Payee address -->
-          <InputRow :deleteable="false" :intended="false" text="contributions and matching funds are sent to">
+          <InputRow text="address all contributions and matching funds are sent to">
             <template v-slot:label>Payee Adddress:</template>
             <template v-slot:input>
               <BaseInput
@@ -151,7 +151,7 @@
           </InputRow>
 
           <!-- Grant Description -->
-          <InputRow :deleteable="false" :intended="false" text="">
+          <InputRow>
             <template v-slot:label>Description:</template>
             <template v-slot:input>
               <BaseTextarea
@@ -167,7 +167,7 @@
           </InputRow>
 
           <!-- Grant website -->
-          <InputRow :deleteable="false" :intended="false" text="">
+          <InputRow>
             <template v-slot:label>Website:</template>
             <template v-slot:input>
               <BaseInput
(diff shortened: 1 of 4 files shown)


mds1:
@thelostone-mc I missed this when we merged the `InputRow.vue` component, but can you or @scco make some small updates to that component:
1. Update the prop definitions there to include `required` and `default` fields. Right now they only define type, so it's not clear which are required. This PR often uses `<InputRow :deleteable="false" :intended="false" text="">`, and it'd be cleaner if all of those unused attributes are just left out, e.g. `text=""` is odd in particular
2. Add comments describing each prop:
    1. `deletable` adds an X icon, but there's no way to attach a method to the X icon, so `deletable` is useless currently and should either be fixed or removed
    2. The purpose of the `intended` prop is not clear
3. Why is `text` a prop but `label` and `input` slots? Seems they should all be slots

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -105,82 +105,120 @@
...
-          <!-- Grant name -->
-          <BaseInput
-            v-model="form.name"
-            description="Your grant's name"
-            id="grant-name"
-            label="Grant name"
-            :rules="isDefined"
-            errorMsg="Please enter a name"
-          />
+          <InputRow :deleteable="false" :intended="false" text="contributions and matching funds are sent to">
```
mds1, line 139: Current phrasing is missing a word
```suggestion
          <InputRow :deleteable="false" :intended="false" text="address contributions and matching funds are sent to">
```

Or something like "address all received funds are sent to"

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -105,82 +105,120 @@
...
+                :rules="isValidTwitter"
+                errorMsg="Please enter a valid Twitter handle"
+                :required="false"
+              />
+            </template>
+          </InputRow>
 
           <!-- Submit and cancel buttons -->
-          <div class="flex justify-end pt-6">
+          <div class="px-4 md:px-12 py-12">
```
mds1, line 218: The Cancel and Update Grant buttons are not aligned

image

thelostone-mc: fixed ! dankee

thelostone-mc:
@mds1 


- Added default + Did this and updated codebase to remove redundant code
- Agree with removing deletable if it doesn't work (do we wanna do it here / in a follow up PR)
- @scco - > could you comment on what intent is used for 
- @scco  -> Why is text a prop but label and input slots? Seems they should all be slots
- BaseInput already exists. We should rename it to BaseInputRow ? Although the name might be confusing.

mds1:
> BaseInput already exists. We should rename it to BaseInputRow ? Although the name might be confusing.

Oops, yes I meant to write `BaseInputRow.vue` above. I actually it's less confusing, because you're almost always going to be using ``BaseInputRow.vue` with ` `BaseInput.vue`, which follows the style guide [here](https://v3.vuejs.org/style-guide/#tightly-coupled-component-names-strongly-recommended) and [here](https://v3.vuejs.org/style-guide/#base-component-names-strongly-recommended)

> Agree with removing deletable if it doesn't work (do we wanna do it here / in a follow up PR)

I'm ok with doing that in this PR. Alternatively we can keep the prop _if we're certain we'll need it in the future_ and add a TODO comment explaining it

apbendi:
Looking much better @thelostone-mc, thanks for tackling this one. A couple small things I noticed:

* There's no padding at the bottom of the page, so the buttons bump straight into the edge of the browser:

Screen Shot 2021-09-13 at 9 27 37 AM

* In mobile mode, I'm not sure this right hand padding makes sense on the input fields. Wouldn't we want to use the whole space?

Screen Shot 2021-09-13 at 9 27 07 AM

thelostone-mc:
@apbendi / @mds1  looks like @scco is fixing em up in another commit .
thoughts on getting this in and fixing it up

gdixon (approved):
LGTM!

mds1 (approved):

PR #219, opened 2021-09-13
chore: use getEtherscanUrl

### LINKED ISSUE #166: Update all Etherscan links to use a `getEtherscanUrl` helper
Currently `TransactionStatus.vue` uses `getEtherscanUrl` to generate the correct Etherscan link based on chain ID. In other places, we hardcode a mainnet Etherscan URL.

We should:
1. Modify `getEtherscanUrl` to return `https://${networkPrefix}/tx/${value}` or `https://${networkPrefix}/address/${value}` based on the length of the first function input, which here I've called `value`
2. Update all hardcoded Etherscan URLs to use this helper method

thelostone-mc:
#### Description

- Replaces etherscan hardcoding with getEtherscanUrl
- lint fixes

closes #166

### DIFF
--- a/app/src/components/BaseHeader.vue
+++ b/app/src/components/BaseHeader.vue
@@ -11,9 +11,14 @@
     <!--optional by address last seen / last  updated  -->
     <div v-if="formattedOwner">
       <span class="text-grey-400 mr-4">by</span>
-      <a class="link mr-4" :href="`https://etherscan.io/address/${owner}`" target="_blank" rel="noopener noreferrer">{{
-        formattedOwner
-      }}</a>
+      <a
+        v-if="owner"
+        class="link mr-4"
+        :href="getEtherscanUrl(owner, 1, 'address')"
+        target="_blank"
+        rel="noopener noreferrer"
+        >{{ formattedOwner }}</a
+      >
       <span v-if="formattedLastUpdated" class="italic text-grey-400">{{ formattedLastUpdated }}</span>
     </div>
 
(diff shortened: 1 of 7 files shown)


app/src/components/BaseHeader.vue
```diff
@@ -11,9 +11,14 @@
     <!--optional by address last seen / last  updated  -->
     <div v-if="formattedOwner">
       <span class="text-grey-400 mr-4">by</span>
-      <a class="link mr-4" :href="`https://etherscan.io/address/${owner}`" target="_blank" rel="noopener noreferrer">{{
-        formattedOwner
-      }}</a>
+      <a
+        v-if="owner"
+        class="link mr-4"
+        :href="getEtherscanUrl(owner, 1, 'address')"
```
mds1, line 17: Can we pull the chainId from the wallet store instead of hardcoding it? You can add a computed property that gets it and exports it, based off the `network` property. Otherwise we'll have to remember to update the hardcoded chainIds here as part of a potential move to an L2

app/src/components/GrantDetailsRow.vue
```diff
@@ -15,7 +15,7 @@
         <span class="text-grey-400 mr-4">Contract:</span>
         <a
           class="link"
-          :href="`https://etherscan.io/address/${payoutAddress}`"
+          :href="getEtherscanUrl(payoutAddress, 1, 'address')"
```
mds1, line 18: Same comment about hardcoding chainId

app/src/components/BaseHeader.vue
thelostone-mc: done

apbendi (commented):
Looks good @thelostone-mc, thanks for tackling that one! Good to merge after resolving the linting errors.

PR #228, opened 2021-09-13
Include Round Logo in Deployment and Implement Round Metadata Update Script

### LINKED ISSUE #142: Grant Round MetaPtr Update Script
We need a script that can be used to update the GrantRound metadata

#### Generic Requirements

- [ ] Can be run against any network in the hardhat config via command line arg
- [ ] Params are configured with a JSON file on a per-network basis
- [ ] Console logs state as it runs
- [ ] Logs all output to a timestamped file that also includes the network in the filename

#### Steps

- [ ] Check the current account is the [metadataAdmin](https://github.com/dcgtc/dgrants/blob/main/contracts/contracts/GrantRound.sol#L22)
    - [ ] Param: The address of the GrantRound contract to be updated
- [ ] Upload the GrantRound metadata to IPFS (see sample schema below)
    - [ ] Param: name
    - [ ] Param: description
    - [ ] Param: Grant IDs
- [ ] Call [updateMetadataPtr](https://github.com/dcgtc/dgrants/blob/main/contracts/contracts/GrantRound.sol#L116) with cloudflare gateway to CID created in previous step

#### Example GrantRound metadata schema

```json
{
    "name": "My Grant Round", 
    "description": "This is my Grant Round, there are many like it, but this one is mine",
    "grants": [1, 3, 7]
}
```

apbendi:
* Updates the deployment script to include uploading a configurable image to IPFS for use as the Round logo
* Creates a new script to update an existing Round's metadata as long as the signer is the `metadataAdmin` (closes #142)

See the changes to the README for instructions on testing the script locally.


contracts/README.md
```diff
@@ -99,6 +105,38 @@ You must configure various deployment parameters on a per-network basis in the `
...
+
+```sh
+$ yarn update:round --network localhost
+```
+
+As with the deployment logging output described above, you will now have a timestamped file in `./deploy-history` containing the log output of this script's execution.
+
+In the same way, you can configure this update script by editing the `./scripts/update-round.config.ts` file, and configure the metadata that will be published during the update by editing the `./scripts/grant-round-metadata.json`.
+
+To update a round on a "real" network like mainnet or a testnet, replace the network name, as defined in `hardhat.config.ts`.
```
mds1, line 134: Small clarification since this phrasing confused me at first
```suggestion
To update a round on a "real" network like mainnet or a testnet, replace the network name, as defined in `hardhat.config.ts`. For example, `yarn update:round --network mainnet`
```

contracts/scripts/deploy-poc.config.ts
```diff
@@ -13,22 +12,24 @@ type NetworkParams = {
...
-  metadataJson: string;
+  roundLogoPath: string; // this path must be relative to the 'contracts' directory
+  metadataJson: Record<string, any>;
 };
 
 type DeployParams = Record<string, NetworkParams>;
 
 const params: DeployParams = {
-  hardhat: {
+  localhost: {
```
mds1, line 22: Thoughts on defining all these params in JSON files instead of TS files? The `grant-round-metadata.json` is JSON, but `deploy-poc.config.ts` and `update-round.config.ts` are TS, and that inconsistency in configuration params confusing to me. JSON feels like the right format here since these are just static values. Then in the deploy script you can cast them to the right types to type-check the JSON values

contracts/scripts/update-round.ts
```diff
@@ -0,0 +1,72 @@
...
+import params from './update-round.config';
+
+const { ethers } = hre;
+
+const ipfs = create({
+  url: 'https://ipfs-api.dev.fleek.cool',
+  headers: {
+    Authorization: `v2 ${process.env.FLEEK_STORAGE_API_KEY}`,
+  },
+});
```
mds1, line 14: `ipfs` instances are created like this in 3 separate places. Feels like a good candidate to pull out into a utility method in `@dgrants/utils`, with the method's input being the Fleek API key

contracts/scripts/update-round.config.ts
```diff
@@ -0,0 +1,31 @@
+import metadataJson from './grant-round-metadata.json';
+
+type NetworkParams = {
+  roundAddress: string;
+  ipfsRetrievalEndpoint: string;
+  roundLogoPath: string; // this path must be relative to the 'contracts' directory
```
thelostone-mc, line 6: wouldn't we just the uploaded image's hash here ?

contracts/scripts/deploy-poc.config.ts
apbendi: I chose TypeScript to mimic the hardhat config file pattern. I think it has a few advantages:

* We can set and enforce types across the networks
* We can "do work" before exporting setting the params, like including other logic or processing. See [here](https://github.com/dcgtc/dgrants/blob/main/contracts/scripts/deploy-poc.config.ts#L2) for an example, which is removed in this PR, but is still demonstrative.
* We get auto-complete and type safety when importing the config into the deploy script itself.

As for why `grant-round-metadata.json` is then a `json` file instead of also being TypeScript: I chose that because it's less of a config, and more of an exact representation of the file as it will exist once uploaded to IPFS. In other words, that file is a `json` file for the same reason the logo file is a `png` file: that's what we're uploading. Now, this ends up being not *exactly* true, because we have to add the logo param to the metadata after it's been uploaded. I still think it's reasonable to keep it as `json` though, so the user can simply provide the file (nearly) as they want to see it on IPFS.

contracts/scripts/update-round.ts
apbendi: Great call. Done in a60210972147240914729f51cff673a853c043f7

contracts/scripts/update-round.config.ts
apbendi: As I think you figured out, the script automates uploading the file to IPFS as well!

contracts/scripts/deploy-poc.config.ts
mds1: You can still enforce types and get type safety when importing JSON, especially since all these type definitions only use basic types that JSON supports (string, number, objects). "Doing work" is a good counterpoint, though your example link just looks like it imports JSON and doesn't do any work, and I'm not sure where we'd use that functionality for these deploy scripts anyway

I think its fine to keep the TS files if you prefer (even though JSON still feels better IMO, but not a big deal either way), but maybe we create a `config` or `deploy-config` folder and move everything there to make things a bit more clear?

app/src/utils/ipfs.ts
```diff
@@ -1,18 +1,9 @@
...
 import { GrantMetadata } from '@dgrants/types';
 
-const config = {
-  storageEndpoint: 'https://ipfs-api.dev.fleek.cool',
-  storageHeaders: {
-    Authorization: `v2 ${import.meta.env.VITE_FLEEK_STORAGE_API_KEY}`, // or Bearer <JWT> or public <AppKey>
-  },
-  retrievalEndpoint: 'https://ipfs-dev.fleek.co/ipfs',
-};
+const RetrievalEndpoint = 'https://ipfs-dev.fleek.co/ipfs';
```
mds1, line 4: Nit on variable naming
```suggestion
const retrievalEndpoint = 'https://ipfs-dev.fleek.co/ipfs';
```

utils/src/ipfs.ts
```diff
@@ -0,0 +1,12 @@
+import { create } from 'ipfs-http-client';
+
+const IpfsEndpoint = 'https://ipfs-api.dev.fleek.cool';
```
mds1, line 3: Same nit
```suggestion
const opfsEndpoint = 'https://ipfs-api.dev.fleek.cool';
```

app/src/utils/ipfs.ts
apbendi: Fixed! I do this out of habit from Swift conventions. Just for reference, here's what the AirBnB guide [says about constant naming](https://github.com/airbnb/javascript/#naming--uppercase) in JavaScript: use ALL_CAPS_SNAKE_CASE, but only if it's exported.

contracts/scripts/config/update-round.config.ts
```diff
@@ -0,0 +1,31 @@
...
+    // expected address using deploy-poc.ts w/ hardhat default account
+    roundAddress: '0x851248dBF29D723f3A12EA9739A9D1c88d1c7faa',
+    ipfsRetrievalEndpoint: 'https://ipfs-dev.fleek.co/ipfs',
+    roundLogoPath: './scripts/config/assets/dgrants-placeholder.png',
+    metadataJson,
+  },
+  mainnet: {
+    roundAddress: '0x0000000000000000000000000000000000000000',
+    ipfsRetrievalEndpoint: 'https://ipfs-dev.fleek.co/ipfs',
+    roundLogoPath: './scripts/dgrants-placeholder.png',
```
mds1, line 26: I think we need to change these paths now since this relative path is no longer accurate?

thelostone-mc:
```suggestion
    roundLogoPath: '../dgrants-placeholder.png',
```

utils/src/ipfs.ts
```diff
@@ -0,0 +1,12 @@
...
+const ipfsEndpoint = 'https://ipfs-api.dev.fleek.cool';
+
+export const createIpfs = (fleekApiKey: string) => {
+  return create({
+    url: ipfsEndpoint,
+    headers: {
+      Authorization: `v2 ${fleekApiKey}`,
+    },
+  });
+};
```
thelostone-mc, line 12: Hmm do wanna move ipfs.ts into here and use that ?
https://github.com/dcgtc/dgrants/blob/main/app/src/utils/ipfs.ts#L4-L15

thelostone-mc (commented):
codewise lgtm. left one feedback

contracts/scripts/config/update-round.config.ts
apbendi: Actually no, we don't do that. At the time this path is used, it's in the context of the hardhat execution environment. See this comment: https://github.com/dcgtc/dgrants/pull/228/files#diff-4e111680456eadf826b03b37ee14a4b29cbae04d7ba601780d839a8a3f80ffd0R6

utils/src/ipfs.ts
apbendi: I'm confused @thelostone-mc, I did update that file as well: https://github.com/dcgtc/dgrants/pull/228/files#diff-281976158b6ff6d4d2122dab6c7b5ed9119e76f5a9181fc9c999698335b6692aR6

contracts/scripts/config/update-round.config.ts
apbendi: Oh, I see, @mds1 you're right that I missed the mainnet paths, but @thelostone-mc your suggested update is wrong for the reason I explained. Will push a fix.

mds1 (approved):

utils/src/ipfs.ts
thelostone-mc: Oh derp derp ! didn't see this! I messed up :P

thelostone-mc (approved):

Issue #220, opened 2021-09-13
Remove the "Equivalent To" row in the cart conversion if all donations are in Dai


apbendi:
It feels odd to have the "Equivalent To" section in the cart if all donations are already denominated in the token selected. The UI should remove this row in this case.

<img width="1140" alt="Screen Shot 2021-09-13 at 9 53 31 AM" src="https://user-images.githubusercontent.com/1479802/133096624-13ff9f4d-83af-4afc-9599-adf876257b15.png">

phutchins:
Resolved by #229

Issue #175, opened 2021-09-03
Add Predictions to Cart Row


apbendi:
- [ ] Refactor to move predictions up so they're not recalculated on each page
- [ ] Show matching predictions on the cart row for items in the cart

PR #229, opened 2021-09-14
Refactor dataStore and add predictions to cart

gdixon:
This PR:

- moves the contribution and prediction logic into `data.ts`
- performs fewer fetches for the TrustBonus scores (moved to `data.ts` - only fetches new contributors)
- adds a mechanism to store data into localStorage and retrieve it again on page-load with checks for blockNumber/timestamp so that we can set the frequency of updates (*)
- removes distribution calc from `data.ts`
- adds matching predictions to the cart row + total

 (*) - @mds1 - I haven't refactored to use `GrantCreated` or `GrantRoundCreated` in this PR - I will open another issue for it 🙌 
 
--

- [x] Refactor to move predictions up so they're not recalculated on each page
- [x] Show matching predictions on the cart row for items in the cart

--

Closes: #175
Closes: #238
Closes: #220 
Closes: #236


app/src/store/cart.ts
```diff
@@ -95,6 +96,102 @@ export default function useCartStore() {
     }
   }
 
+  // convert the amount to a different token (direction allows for moving to/from the tokenAddress)
+  function getConvertedAmount(amount: number, tokenAddress: string, direction = 1) {
```
mds1, line 100: Can you document with a comment whether `amount` is human-readable amount (e.g. 1 DAI) or raw/parsed number (e.g. 1e18 DAI). If the latter, make sure to change multiplication and division to use the `BigNumber` class

app/src/store/cart.ts
```diff
@@ -95,6 +96,102 @@ export default function useCartStore() {
...
+    const data = ((grantRoundsCLRData && grantRoundsCLRData.value && grantRoundsCLRData.value) || {}) as {
+      [grantRound: string]: GrantRoundCLR;
+    };
+    const roundData = data[round.address] || {};
+
+    return roundData.predictions && roundData.predictions[Number(grantId)];
+  }
+
+  // given the grantId and the contribution amount find all matching in all rounds
+  function getPredictionForAmount(grantId: string, amount: number, token: TokenInfo) {
```
mds1, line 122: Lots of these methods don't feel like cart-specific methods, and feel more like dcurve methods or utils. They also don't rely on the cart state, so I think they should be pulled out to live somewhere else (partly because the cart file is getting pretty big too)

app/src/store/cart.ts
```diff
@@ -95,6 +96,102 @@ export default function useCartStore() {
...
+    return amount * (direction == 1 ? exchangeRate : 1 / exchangeRate);
+  }
+
+  // Linear Interpolation function to find the point on the curve that amount corresponds to
+  function lerp(x_lower: number, x_upper: number, y_lower: number, y_upper: number, x: number) {
+    return y_lower + ((y_upper - y_lower) * (x - x_lower)) / (x_upper - x_lower);
+  }
+
+  // returns the prediction curve for this grant in the given round
+  function getPredictionsForGrantInRound(grantId: string, round: GrantRound) {
```
mds1, line 112: For all the methods with `grantId` as an input, I'd suggest changing the type to `BigNumberish` and converting the input to the required type with `BigNumber.from(grantId).toString()? A bit more robust that way / removes the burden of handling types from the caller

app/src/store/cart.ts
```diff
@@ -95,6 +96,102 @@ export default function useCartStore() {
...
+        const clr_predictions = getPredictionsForGrantInRound(grantId, round);
+        const clr_prediction_curve = clr_predictions?.predictions?.map((prediction, key) => {
+          return key == 0 ? prediction.predictedGrantMatch : prediction.predictionDiff;
+        });
+
+        let predicted_clr = 0;
+
+        // convert amount to the matchingToken (double hop to get into dai then into the matchingToken)
+        amount =
+          token.symbol == 'dai' || token.address == round.matchingToken.address
```
mds1, line 137: Where are the token symbols populated from? The supported tokens list uses all caps here, and `'dai' == 'DAI'` returns false, so I think we want to change these token equality checks to `token.symbol === 'DAI'` or they'll always return false in that branc.

Alternatively, we really only need the address checks anyway, because the symbol and address checks are equivalent and should either both be true or both be false

app/src/store/cart.ts
```diff
@@ -95,6 +96,102 @@ export default function useCartStore() {
...
+            ? amount
+            : getConvertedAmount(amount, token.address, 1);
+        amount =
+          round.matchingToken.symbol == 'dai' || token.address == round.matchingToken.address
```
mds1, line 141: Same comment on symbol

app/src/store/cart.ts
```diff
@@ -324,13 +421,48 @@ export default function useCartStore() {
...
+        item.grantId,
+        item.contributionAmount,
+        item.contributionToken
+      ) as CartPrediction[];
+    });
+
+    return _predictions;
+  });
+
+  const clrPredictionsByToken = computed<{ [key: string]: number }>(() => {
```
mds1, line 440: Can you add a comment describing this property? I think it takes `clrPredictions` and groups/sums the match results by token, similar to what `cartSummary` does to `cart`, but want to confirm

app/src/store/data.ts
```diff
@@ -6,7 +6,7 @@
 import { computed, markRaw, Ref, ref } from 'vue';
 
 // --- Our imports ---
-import { BigNumber, Contract } from 'src/utils/ethers';
+import { BigNumber, Contract, formatUnits } from 'src/utils/ethers';
```
mds1, line 9: This is a big file, and I'm no longer certain exactly what's happening in this file, what order data is being fetched, which things are only fetched once, etc. Any suggestions on how to make this more clear? Just comments? Or maybe can this be cleaned up by extracting helper methods? We don't have to do it in this PR, just wanted to bring attention to it

app/src/store/data.ts
```diff
@@ -37,94 +48,288 @@ const lastBlockTimestamp = ref<number>(0);
...
    * @notice Called each block to poll for data, but can also be called on-demand, e.g. after user submits a transaction
    */
-  async function poll() {
+  async function rawPoll(forceRefresh = false) {
     if (!multicall.value || !registry.value || !roundManager.value) return;
 
     // Define calls to be read using multicall
     const calls = [
       { target: MULTICALL_ADDRESS, callData: multicall.value.interface.encodeFunctionData('getCurrentBlockTimestamp') },
-      { target: GRANT_REGISTRY_ADDRESS, callData: registry.value.interface.encodeFunctionData('getAllGrants') },
```
mds1, line 56: Do we even need this first round of multicalls here? I don't think we use this timestamp for anything and can probably get rid of it. Seems like an artifact of the default dummy `data.ts` code we had when we initialized the project

app/src/store/data.ts
```diff
@@ -37,94 +48,288 @@ const lastBlockTimestamp = ref<number>(0);
...
+
+        return data?.grantRound;
+      })
+    )) as GrantRound[];
+
+    // Get latest set of contributions
+    const contributions = await fetchContributions(grantsDict, grantRoundsList[0].donationToken, forceRefresh);
+
+    // Pull all trust scores from gitcoin api
+    const trustBonusScores = await fetchTrustBonusScores(contributions.contributions, forceRefresh);
```
mds1, line 120: Looks like the async calls for `contributions` and `trustBonusScores` can be batched with `Promise.all` to speed it up a bit

app/src/store/data.ts
```diff
@@ -37,94 +48,288 @@ const lastBlockTimestamp = ref<number>(0);
...
+    grantRounds.value = grantRoundsList as GrantRound[];
+    grantRoundsCLRData.value = grantDataByRound as { [key: string]: GrantRoundCLR };
+    grantContributions.value = contributions.contributions as { [hash: string]: Contribution };
+  }
+
+  async function syncWithLocalStorage(
+    key: string,
+    callback: (localStorageData?: LocalStorageData) => Promise<LocalStorageAnyObj>,
+    validityCheck?: (data: LocalStorageAnyObj) => boolean
+  ) {
```
mds1, line 164: Can you add comments to all the methods here and document the function's purpose / input parameters / return data? For example it's not too clear what I should be doing with the callback and validity checks, and why this method even needs those, instead of syncing the full store each time

app/src/views/Cart.vue
```diff
@@ -86,8 +86,15 @@
...
                 <div class="text-grey-400 text-left lg:text-right">
                   <!-- TODO use real match estimates -->
-                  <p v-if="true">not in an active round</p>
-                  <p v-else>USD estimated matching</p>
+                  <p v-if="clrPredictions[item.grantId]">
+                    <span v-for="(clr, index) in clrPredictions[item.grantId]" :key="index">
+                      {{ formatNumber(clr.matching, 2) }} {{ clr.matchingToken.symbol }}
+                      {{ index !== clrPredictions[item.grantId].length - 1 ? '+' : '' }}
+                      <br />
+                      estimated matching
```
mds1, line 94: Why do some of these have match amounts of zero? 

image

mds1:
Also looks like this branch needs a rebase to resolve conflicts

app/src/store/cart.ts
gdixon: Ahh, good catch on the casing!! I've refactored and fixed up the naming a little bit to show intent - let me know if theres anything here we need to change

app/src/store/cart.ts
gdixon: These should always be human readable - I've added comments, thank you!

app/src/store/cart.ts
gdixon: I've abstracted some of the less relevant ones over to `dcurve`, but kept anything that depends on `quotes`

app/src/store/data.ts
gdixon: All fetching methods have been moved into a new directory (`./app/src/utils/data/*`) - The dataStore should be more manageable now 👍

app/src/store/data.ts
gdixon: This has been removed 🙌

app/src/store/data.ts
gdixon: `TrustBonusScores` depends on the `contributions` to get the `contributor` list

app/src/store/data.ts
gdixon: I've tidied this up a little bit -- if we think this might add friction then I'm happy to refactor to use a setter/getter approach, I think this just read a little bit better?

gdixon: I've tidied this up a little bit -- if we think this might add friction then I'm happy to refactor to use a getter/setter approach, I think this just read a little bit better?

app/src/views/Cart.vue
gdixon: Hmmm, I'm not sure, could you try it with the latest changes?

utils/.eslintcache
```diff
@@ -0,0 +1 @@
+[{"/Users/graham/dgrants/utils/src/trustBonus.ts":"1"},{"size":1477,"mtime":1631756544883,"results":"2","hashOfConfig":"3"},{"filePath":"4","messages":"5","errorCount":0,"fatalErrorCount":0,"warningCount":0,"fixableErrorCount":0,"fixableWarningCount":0},"12uxfxl","/Users/graham/dgrants/utils/src/trustBonus.ts",[]]
```
thelostone-mc, line 1: is this needed?

thelostone-mc (approved):
LGTM! left one minor comment

mds1 (approved):
Only skimmed through the changes this morning but in general LGTM, thanks @gdixon! As long as someone has tested/verified the functionality I'd say it's ready to merge

thelostone-mc:
LOL @mds1 go offline :P It's your day off

phutchins (approved):

PR #230, opened 2021-09-14
fix: Cart input/select alignment issue

gdixon:
This PR puts in a fix for the alignment issue we have on the cart rows mentioned here by @scco 🙌 

Fixes: https://github.com/dcgtc/dgrants/pull/177#issue-727004965

### DIFF
--- a/app/src/components/BaseInput.vue
+++ b/app/src/components/BaseInput.vue
@@ -1,9 +1,9 @@
 <template>
-  <div :class="width">
+  <div :class="containerClass">
     <input
       v-model="val"
       @input="onInput"
-      :class="[customcss ? customcss : '']"
+      :class="[inputClass ? inputClass : '']"
       :id="id"
       :name="id"
       :required="required"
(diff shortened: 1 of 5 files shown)


thelostone-mc:
Oh we might have to more changes once https://github.com/dcgtc/dgrants/pull/216 gets in

scco:
dont know how this fixed it - as the BaseInput would need that fixed height of "h-14" too - or not ?

regarding this  "h-14" class you added :

i try to prevent fixed heights in inputs for a lot of reasons 
e.g. ( regarding browser compatibility, vertical alignment within inputs and so on. )

i would prefer let the dimensions be defined by the font-size + padding  automatically
what would be then consistent heights on ALL input elements automatically.

thats why i think the right way of fixing it, is get rid of the fake-input + headless  ui dependency
and bring back the native default <select> for consistency.

mds1 (approved):

scco:
this is my not-ready-yet approach for <BaseSelect> to replace
the non-native Listbox, ListboxOptions ... for a simple native <select>
what would lead to 

-better mobile usability
-better consistency
-easyer to combine with other <inputs> without ui issues
-get rid of a dependency ( @headlessui/vue )
-less complex code

im not 100% sure that <BaseSelect> in my version  would be 100% reusable because of this 
`@update:model-value="$emit('update:modelValue', $event)"` 

so here is what i came up with , maybe someone can inspect and see if this might be it ? 

https://github.com/dcgtc/dgrants/tree/richard-bring-back-native-select

gdixon:
Closing this PR in favour of https://github.com/dcgtc/dgrants/pull/240

Issue #140, opened 2021-08-24
app: use smarter pathfinding algorithm for checkout


mds1:
In #110 we hardcoded swap paths for known liquid token pairs. This is not ideal/robust/scalable, so a smarter pathfinding algorithm should be implemented. 

The easiest way to do this is likely to use [this method](https://github.com/Uniswap/uniswap-interface/blob/1304acd8c7cac97578dab9f64b686ccca6e421a1/src/hooks/useBestV3Trade.ts#L26-L113) from the Uniswap interface
- It's written for React so we'll need to make some modifications, e.g. remove `useMemo` and add `chainId` as an input
- The interface is GPL-3.0 so we can use that code freely

mds1:
Uniswap recently released the [Auto Router](https://uniswap.org/blog/auto-router/) for Uniswap V3. This is now preferred over the method linked in the above comment

If we also support V2 swaps (https://github.com/dcgtc/dgrants/discussions/263), we should use Uniswap's V2 routing algorithm in that case

PR #235, opened 2021-09-14
Multi-chain config

mds1:
Sets up the app to easily work with dGrants deployment on any chain. The focus of this PR was rinkeby. The `VITE_DGRANTS_CHAIN_ID` in `app/.env` defines the chain ID the app should load data from, regardless of the chain the user's wallet is connected to (currently this functionality is broken and you need to be the same chain ID to read the data, cc @gdixon since we talked about this) 

The `chains.ts` file is based on Uniswap's implementation, and is the core file for modifying chain settings. The very bottom of the file contains the chain-specific exports that should be used by other files throughout the app


app/src/store/cart.ts
```diff
@@ -52,10 +31,50 @@ const SWAP_PATHS = {
...
+    // ETH to USDC through the 0.3% pool
+    '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE': '0x82aF49447D8a07e3bd95BD0d56f35241523fBab1000bb8ff970a61a04b1ca14834a43f5de4533ebddb5cc8', // prettier-ignore
+    // USDC "swap path" is just its token address for our router (no DAI on Arbitrum yet, so using USDC instead)
+    '0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8': '0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8'.toLowerCase(),
+  },
+  [SupportedChainId.ARBITRUM_RINKEBY]: {
+    // Seems Uniswap V3 has no liquidity here (this swap path is not correct)
+    '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE': '0x82aF49447D8a07e3bd95BD0d56f35241523fBab1000bb86b175474e89094c44da98b954eedeac495271d0f', // prettier-ignore
+    // TODO choose donation token since there's no DAI -- USDC?
+  },
```
thelostone-mc, line 73: ^ would make sense to push this to it's file after MVP launch

thelostone-mc (approved):

app/README.md
```diff
@@ -24,6 +24,10 @@ This configuration is required to ensure your account is funded with tokens for
 When you rebuild the app, you'll likely need to reset MetaMask so the nonces match what the local network expects.
 You can do this in MetaMask by clicking the circle in the top right > Settings > Advanced > Reset Account.
 
+Set `DGRANTS_CHAIN_ID` to the chain ID containing the set of dGrants contracts to use.
+This chain ID will be the only supported network after building the app.
+This architecture is used to ensure that regardless of the network a user's wallet is connected to when visiting the dGrants site, they are shown the correct set of grants and can populate their cart.
```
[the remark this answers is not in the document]
mds1: That's the goal, but currently that's not working. Me, you, or @gdixon need to investigate. I'm guessing it's something I configured wrong in `wallet.ts` or `data.ts` using the wrong provider

app/src/store/cart.ts
mds1: After MVP we'll replace this with a better solution: #140

PR #251, opened 2021-09-17
Round update for rinkeby

phutchins:
Deployed new contracts with a grant round that we can update. Updated the round to contain the first 6 grants.


thelostone-mc:
@phutchins do we want the deploy history here as well ? I think we can remove those

apbendi (approved):
looks good to me

thelostone-mc (approved):

mk1r (commented):
looking good, I would just second what was mentioned about deploy-history files - only include the successful one and disregard the failed ones

contracts/hardhat.config.ts
```diff
@@ -43,6 +43,7 @@ function createTestnetConfig(network: keyof typeof chainIds): NetworkUserConfig
   return {
     accounts: [(process.env.DEPLOY_PRIVATE_KEY as string) || dummyPrivateKey],
     chainId: chainIds[network],
+    allowUnlimitedContractSize: true,
```
mds1, line 46: @thelostone-mc @apbendi @phutchins why did we need to add this flag?

contracts/scripts/deploy-poc.ts
```diff
@@ -71,7 +71,8 @@ const ipfs = createIpfs(process.env.FLEEK_STORAGE_API_KEY!);
       networkParams.matchingToken,
       networkParams.roundStartTime,
       networkParams.roundEndTime,
-      metadataPtr
+      metadataPtr,
+      { gasLimit: 1000000 }
```
mds1, line 75: @thelostone-mc @apbendi @phutchins and why did we need to hardcode the gas limit? We should prefer having the node estimate this. I'm guessing this was because deploys were failing at gas estimation, so we tried hardcoding a gas limit to bypass that? If so, this change should be undone

contracts/hardhat.config.ts
mk1r: this was because we thought that the contract size was too large - the issue was with startTime (this is deleted in code cleanup commit)

contracts/scripts/deploy-poc.ts
mk1r: yes exactly, it couldn't estimate gas - probably better to delete this now
btw, is there a better solution to solving gas estimation error?

Issue #245, opened 2021-09-16
bug: local storage get stuck and newly added grants are not visible


phutchins:
For Grants/GrantRounds, our data mechanism is reading every entry from the contract every time we update, because of this we have set it to only update every 10 minutes. This isn't frequent enough.

We need to:

- Update the data reqs to read from events
- Alter the localSorage mechanism to check for updates every block - wherever this isn't possible we should shorten the caching period.

PR #252, opened 2021-09-17
Updates data helpers to use events to get new state

gdixon:
This PR:

- [x] Updates the data reqs to read from events
- [x] Alters the localSorage mechanism to check for updates every block - wherever this isn't possible we shorten the caching period (`getGrantRound` - because we need to update every time a new Round Contribution is added)

Closes: #245

### DIFF
--- a/app/src/utils/data/grantRounds.ts
+++ b/app/src/utils/data/grantRounds.ts
@@ -41,17 +41,24 @@ export async function getAllGrantRounds(blockNumber: number, forceRefresh = fals
     },
     async (localStorageData?: LocalStorageData | undefined, save?: () => void) => {
       const { grantRoundManager } = useWalletStore();
+      // use the ls_blockNumber to decide if we need to update the roundAddresses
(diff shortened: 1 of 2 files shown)


apbendi (approved):
Working for me *after* clearing my local storage first. 👍

thelostone-mc (approved):
tested locally and works fine

app/src/utils/data/grantRounds.ts
```diff
@@ -87,18 +94,18 @@ export async function getGrantRound(blockNumber: number, grantRoundAddress: stri
       const { provider, multicall } = useWalletStore();
       const data = localStorageData?.data || {};
 
-      // read from registry on first load and every 10 mins
+      // read from registry on first load and every 30 secs after (we need to pick up any changes)
```
mds1, line 97: @gdixon Why do we special case this to run every 30 seconds instead of just running each block, which is every ~13 seconds?

PR #258, opened 2021-09-17
bug: changing name of grant var to avoid colission

phutchins:


mk1r (approved):
nice fix - tested it by clearing local storage again, going to the correct grant now

Issue #265, opened 2021-09-20
Post-POC tech debt cleanup


mds1:
Below is a list of tech debt-related things I've noticed. When applicable, I've included the PR number where the item was introduced for traceability, in case any of these are intentional for a reason I've missed, and because in some of them I left comments on the specific line being referring to

- [ ] [PR 248](https://github.com/dcgtc/dgrants/pull/248): `sortByStartTime` method duplicated in two places, should be moved into utils (introduced in 248)
- [ ] [PR 248](https://github.com/dcgtc/dgrants/pull/248): `useGrantRegistryList` composition function mostly duplicated in two files. Seems there are slight differences so maybe leave this? (cc @mk1r)
- [ ] [PR 249](https://github.com/dcgtc/dgrants/pull/249): added another hardcoded reference to a specific grant round, so we just need to keep track of these somewhere. The other reference is in data.ts [here](https://github.com/dcgtc/dgrants/blob/ef47c9cac0db13ca76f467708d800ee9e05da675/app/src/components/GrantCard.vue#L59), and there may be others? (cc @gdixon)
- [ ] Type-related issues:
  - [ ] [PR 249](https://github.com/dcgtc/dgrants/pull/249): GrantCard.vue: `getTotalRaised` input type should be `BigNumberish`, not `any`, and the input should be handled/casted as appropriate
  - [ ] General Grant ID inconsistency throughout app: sometimes grant IDs are numbers (e.g. `GrantRegistryGrantDetail.vue`), sometimes strings (e.g. `GrantRegistryNewGrant.vue`), and sometimes `any` (e.g. throughout `GrantCard.vue`). Grant ID inputs should always be of type `BigNumberish` and then cast to a number with `BigNumber.from(x).toNumber()`, and when a grant ID is used somewhere it should always be number
  - [ ] `grantContributions` ref in `data.ts` should default to empty array instead of undefined? If yes, this removes the need for `|| []` clause [here](https://github.com/dcgtc/dgrants/blob/ef47c9cac0db13ca76f467708d800ee9e05da675/app/src/components/GrantCard.vue#L55) (cc @gdixon) 
- [ ] Docs questions
[...]

Issue #253, opened 2021-09-17
Implement top level grant filter


phutchins:
As a stop gap prior to having a grant curation system in place, we should implement a hard coded list of grants that are white-listed for being displayed on the dGrants system.

Requirements
- List should be hosted on IPFS OR provide an ENV variable to configure white-listed grant ids
- Should be able to add grants by their ID to an array I.E. [0,1,2,3,5,10]
- Should be able to change the setting to show all grants - Empty array should signify show all grants

apbendi:
For MVP, probably fine to go with a JSON file with lists by chainId.

PR #277, opened 2021-09-22
Sorting grants

### LINKED ISSUE #262: Implement Grant Sorting on All Grant List Pages
Should enable the user to sort by grant creation time, using grant id as a proxy for the order in which the grants were created (lower id means older grant).

- [ ] Newest (lower ids first)
- [ ] Oldest (higher ids first)
- [ ] Shuffle (show them in a totally random order)

carletex:
Fixes #262 

I created a new `GrantListWithFilter` component to be able to reuse the sorting functionality across different views. Also added a `bigNumberComparator` util.

@apbendi from the issue you created:

> Should enable the user to sort by grant creation time, using grant id as a proxy for the order in which the grants were created (**lower id means older grant**).
> - Newest (lower ids first)
> - Oldest (higher ids first)
> - Shuffle (show them in a totally random order)

I think it's the other way around, right? If lower id means older grant... then if we sort by "Newest" we would want the higher id first.


carletex:
As we discussed in the stand up, marking this as draft until I implement the "multiple shuffle" working

carletex:
https://github.com/dcgtc/dgrants/pull/277/commits/04bc59e617b35f393b7cfe0f1291588607b78f6a Implemented a shuffle sorting "nonce" to force the `sortedGrants` to recalculate when clicking Shuffle consecutively. 

Not sure if this is best way to go. Maybe Vue has some re-render magic I'm not aware of.

app/src/components/GrantListWithFilter.vue
```diff
@@ -0,0 +1,102 @@
...
+    grantMetadata: { type: Object as PropType<Record<string, GrantMetadataResolution>>, required: true },
+    button: { type: Object as PropType<FilterNavButton>, required: false },
+  },
+  setup(props) {
+    const { addToCart, isInCart, removeFromCart } = useCartStore();
+    const sortedGrants = computed(() => {
+      // Force to recalculate when clicking "Shuffle" consecutively.
+      shuffleNonce.value;
+      return [...props.grants].sort((a, b) => bigNumberComparator(a.id, b.id, grantsSortingMode.value));
+    });
```
mds1, line 87: I think a cleaner way to do this might be to remove the computed property and create a new ref from the `grants` prop which stores the sorted refs. Then have the actions in `grantRegistryListNav` call a method called `sortGrants` which does:

```typescript
const sortedGrants = ref(props.grants);
function sortGrants(order: 'ascending' | 'descending' | 'shuffle') {
   if (order === 'ascending') sortedGrants.value = props.grants.sort((a,b) => a.id > b.id ? 1 : -1)
   else if (order === 'descending') sortedGrants.value = props.grants.sort((a,b) => a.id < b.id ? 1 : -1)
   else sortedGrants.value = props.grants.sort((a,b,) => Math.random() < 0.5 ? 1 ? -1 // shuffle is default
}
```

(I may have the orders wrong but you get the idea). Then you don't need the `bigNumberComparator` and you can remove the shuffle nonce

mds1: With this change you can also move the `SortingMode` type definition into this file since it's no longer needed anywhere else

app/src/components/GrantListWithFilter.vue
```diff
@@ -0,0 +1,102 @@
+<template>
```
mds1, line 1: Just curious, why did you decide to use a new file here here instead of putting the logic directly in `GrantRegistryList.vue`?

app/src/components/GrantListWithFilter.vue
```diff
@@ -0,0 +1,102 @@
...
+import { bigNumberComparator } from 'src/utils/sorting';
+
+const defaultSortingMode = 'descending';
+const grantsSortingMode = ref<SortingMode>(defaultSortingMode);
+const shuffleNonce = ref(1);
+
+const filterNavTag = computed(() => {
+  switch (grantsSortingMode.value) {
+    case 'descending': {
+      return 'newest';
```
mds1, line 27: Why do we alias `descending` to `newest`, `random` to `shuffle`, etc? Can we just use the user-facing words as the sort mode names? Seems cleaner that way

app/src/views/GrantRoundGrants.vue
```diff
@@ -1,9 +1,8 @@
 <template>
-  <template v-if="hasLoadedHeader">
+  <template v-if="hasLoadedHeader && hasLoadedGrants">
```
mds1, line 2: Why did we move `hasLoadedGrants` from the `GrantList` component to the `template` element?

mk1r:
I agree with @mds1 on his suggestions. I'm also not sure if we need a new file (i.e `GrantListWithFilter.vue`), seems the sorting functionality could be added inside `GrantRegistryList.vue` or better yet inside`GrantList.vue`. Sorting method implementation can be inside a different file (perhaps in `utils.ts`). 

We can also discuss this again when discussing the Home.vue duplicates in #265.

I have some ideas on specifics as this PR is related to #253.

mk1r:
I agree with @mds1 on his suggestions. I'm also not sure if we need a new file (i.e GrantListWithFilter.vue), seems the sorting functionality could be added inside GrantRegistryList.vue or better yet insideGrantList.vue. Sorting method implementation can be inside a different file (perhaps in utils.ts).

We can also discuss this again when discussing the Home.vue duplicates in #265.

I have some ideas on specifics as this PR relates to #253.

mk1r:
I agree with @mds1 on his suggestions. I'm also not sure if we need a new file (i.e `GrantListWithFilter.vue`), seems the sorting functionality could be added inside `GrantRegistryList.vue` or better yet inside `GrantList.vue`. Sorting method implementation can be inside a different file (perhaps in `utils.ts`).

We can also discuss this again when discussing the `Home.vue` duplicates in #265.

I have some ideas on specifics as this PR relates to #253.

carletex:
@mds1 @mk1r I think we have different approaches and wanted to discuss it before making further changes. I don't want to fall into a review loop or waste people's time, but I think this issue is a good chance to discuss it that can save us some time in the future.

I always try to write declarative & functional (probably biased from React), instead of being imperative. I'll try to explain my approach using the comments raised in this PR.

### Sorting

I created a `bigNumberComparator` util and a `SortingMode` type to decouple sorting between different types of concepts and moving the sorting logic outside the «domain» layer. 

We might want to sort by newest/oldest or amount raised, but we are just actually sorting BigNumbers behind the scenes. That's why we are using `ascending | descending | shuffe` instead of `newest | oldest` (tied to a specific property). Otherwise we'd have to re-implement BigNumber sorting for every use case.

This is the code you suggested for the sorting logic
```js
const sortedGrants = ref(props.grants);
function sortGrants(order: 'ascending' | 'descending' | 'shuffle') {
   if (order === 'ascending') sortedGrants.value = props.grants.sort((a,b) => a.id > b.id ? 1 : -1)
   else if (order === 'descending') sortedGrants.value = props.grants.sort((a,b) => a.id < b.id ? 1 : -1)
   else sortedGrants.value = props.grants.sort((a,b,) => Math.random() < 0.5 ? 1 ? -1 // shuffle is default
}
```
For me the following looks cleaner, easier to understand, and we are not storing derived state, relying on a single source of truth:

```js
const sortedGrants = computed(() => {
  // Force to recalculate when clicking "Shuffle" consecutively.
  shuffleNonce.value;
  return [...props.grants].sort((a, b) => bigNumberComparator(a.id, b.id, grantsSortingMode.value));
});
```

This has the problem with the shuffle "nonce" hack. But we just have `grantsSortingMode` as state and everything is derived from it on updates.

### Component structure
> Just curious, why did you decide to use a new file here here instead of putting the logic directly in GrantRegistryList.vue?
> I'm also not sure if we need a new file (i.e GrantListWithFilter.vue), seems the sorting functionality could be added inside GrantRegistryList.vue or better yet inside GrantList.vue

We are using the Filtered Grant List in 3 places (so far): home, /dgrants and /dgrants/rounds/:address/grants. So putting the logic inside `GrantRegistryList.vue` would mean to duplicate the logic in every view (or maybe use a Vue composable pattern).

Putting the login on `GrantList` is an option, but I decided to create a new component. That makes `GrantList` easier to reuse on its own (without the filter). I also considered to extend `GrantList` with a new prop to hide/show the filters, but I decided to go with two simpler components, rather than a single more complex component. I don't feel too strongly about it though.

## Wrap up

Having said all of that (sorry for this wall of text :angel:), I'm more than happy to implement all the changes you requested, just wanted to show my approach so you can understand my reasoning.

Thanks!

app/src/views/GrantRoundGrants.vue
carletex: You're right, there is no need. Just put it back https://github.com/dcgtc/dgrants/pull/277/commits/0cad7951dcdf9f12c4d569e5cd898228da7ead1d

mds1:
Thanks @carletex, appreciate the detailed response!

> I created a bigNumberComparator util and a SortingMode type to decouple sorting between different types of concepts and moving the sorting logic outside the «domain» layer.
> 
> We might want to sort by newest/oldest or amount raised, but we are just actually sorting BigNumbers behind the scenes. That's why we are using ascending | descending | shuffe instead of newest | oldest (tied to a specific property). Otherwise we'd have to re-implement BigNumber sorting for every use case.

This is totally reasonable and in general I agree

My counter here would be that in this case it's premature optimization / "premature modularity" with no real benefit, since there's nothing else we use it for. In this case, yes we are just sorting numbers, but post-POC when we have fancier sort methods (grant amount raised, round match amounts, etc.), it becomes more complex and we likely won't be able to use the same sorting logic or same `SortingMode` type for both grant-sorting and rounds-sorting anyway

> This has the problem with the shuffle "nonce" hack. But we just have grantsSortingMode as state and everything is derived from it on updates.

This is the "code smell" of the current approach—whenever you need a hack like this to trigger Vue's reactivity, it's a sign you are doing something suboptimally/not the "Vue way". One solution is what I suggested, another is to add a setter method on your computed property (see syntax [here](https://stackoverflow.com/a/64281689)). Either approach is ok with me if you prefer the computed setter. 

I totally understand the preference for functional approaches and generally like that approach, but functional approaches don't work well for SPAs like Vue apps where you have shared, mutable state

> We are using the Filtered Grant List in 3 places (so far): home, /dgrants and /dgrants/rounds/:address/grants. So putting the logic inside GrantRegistryList.vue would mean to duplicate the logic in every view (or maybe use a Vue composable pattern).
> 
> Putting the login on GrantList is an option, but I decided to create a new component. That makes GrantList easier to reuse on its own (without the filter). I also considered to extend GrantList with a new prop to hide/show the filters, but I decided to go with two simpler components, rather than a single more complex component. I don't feel too strongly about it though.

Sorry, I'm not super familiar with the internals of each component here. Which logic would be duplicated in each of those components? Either way though this seems ok to me, as long as it also makes sense to @gdixon or @thelostone-mc who are probably more familiar with this part of the codebase than I am

mk1r:
> Sorry, I'm not super familiar with the internals of each component here. Which logic would be duplicated in each of those components?

@mds1 `Home.vue`, `GrantRegistryList.vue`, and `GrantRoundGrants.vue` these are the pages that call the `GrantList.vue` component. To @carletex's answer to

> Just curious, why did you decide to use a new file here here instead of putting the logic directly in `GrantRegistryList.vue`?

the reason he didn't include sorting inside `GrantRegistryList.vue` is because he would then need to duplicate the code in multiple pages, hopefully this clarifies a little more.

@carletex I still think it would be better to create a sorting method in `utils.ts` and perhaps import that function into `GrantList.vue` and we should still be able to reuse it without the filter (i.e no sort). I don't have a strong opinion either necessarily, but it feels like it would be a cleaner approach.

carletex:
Hey @mds1 thanks for the detailed response. I definitely have some counter arguments to your arguments, but let's save it for another occasion :P 

Made the following changes:
- Removed `GrantListWithFilters` and embed the logic inside the existing `GrantList`
- Inlined the sorting logic into the component.

Let me know if you spot any other tweaks to make.

@mk1r 

> @carletex I still think it would be better to create a sorting method in utils.ts and perhaps import that function into GrantList.vue and we should still be able to reuse it without the filter (i.e no sort). I don't have a strong opinion either necessarily, but it feels like it would be a cleaner approach.

That's how it was done before (extracting the sorting logic using a BigNumberComparator) but after Matt's feedback I inlined it into the component.

carletex:
I think the CI error is unrelated to the PR.

image

Just clicked re-run.

mds1 (approved):

mds1:
LGTM, just approved! Thanks @carletex 

> I definitely have some counter arguments to your counter arguments, but let's save it for another occasion :P

Excited to hear them! If you want to discuss further, feel free to start a discussion in this repo with some Post-POC architecture changes/suggestions. We are tracking tech debt items in #265, so depending on the outcome of the discussion we can always add more items there and triage/prioritize them later

> I think the CI error is unrelated to the PR.

Yea it is unrelated, in #281 I bumped the test timeout, but looks like re-running was sufficient here anyway

mk1r (approved):

PR #278, opened 2021-09-22
equivalent amount zero

### LINKED ISSUE #233: Equivalent amount shows as zero until it loads sometimes requiring a reload to update
When loading the cart with a grant in it, the Equivalent amount is displayed as 0 for some period of time. It did load for me after a few seconds but it was enough time to make me stare at it and wonder why it displayed 0.

The amount should display a - or something similar instead while its loading.

Additionally, after coming back to the page after a while, the Equivalent amount showed 0 and required a page refresh to get it to update.

Screen Shot 2021-09-14 at 9 51 11 AM

mk1r:
- attempting to resolve #233 
- needs a site review as I'm not seeing the error
- small refactor for show / hide logic

### DIFF
--- a/app/src/views/Cart.vue
+++ b/app/src/views/Cart.vue
@@ -134,21 +134,18 @@
           <span class="text-grey-400">Equivalent to:</span>
           <span>~{{ formatNumber(equivalentContributionAmount, 2) }} DAI</span>
         </div>
-
         <LoadingSpinner v-else />
       </div>
 
       <div class="py-8 border-b border-grey-100">
-        <div class="flex gap-x-4 justify-end">
+        <div v-if="Object.keys(clrPredictionsByToken).length" class="flex gap-x-4 justify-end">
           <span class="text-grey-400">Estimated matching value:</span>
-          <span v-if="Object.keys(clrPredictionsByToken).length">
-            <span v-for="(symbol, index) in Object.keys(clrPredictionsByToken)" :key="index">
-              {{ formatNumber(clrPredictionsByToken[symbol], 2) }} {{ symbol }}
-              {{ index !== Object.keys(clrPredictionsByToken).length - 1 ? '+' : '' }}
-            </span>
+          <span v-for="(symbol, index) in Object.keys(clrPredictionsByToken)" :key="index">
+            {{ formatNumber(clrPredictionsByToken[symbol], 2) }} {{ symbol }}
+            {{ index !== Object.keys(clrPredictionsByToken).length - 1 ? '+' : '' }}
           </span>
-          <span v-else> 0.0 DAI </span>
         </div>
+        <LoadingSpinner v-else />
       </div>
 
       <div class="mt-12 mb-12">
@@ -236,7 +233,7 @@ function useCart() {
     }
     // hide when cart is all denominated in a single stablecoin
     hideEquivalentContributionAmount.value =
-      sum == 0 || (isStableCoin == 1 && isStableCoin === Object.keys(cartSummary.value).length);
+      isStableCoin == 1 && isStableCoin === Object.keys(cartSummary.value).length;
     return sum;
   });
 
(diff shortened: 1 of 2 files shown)


mds1:
Reminder to self: do not merge until we hear from @gdixon about whether his PR resolves this

app/src/views/Cart.vue
```diff
@@ -129,11 +129,13 @@
         </div>
       </div>
 
-      <div class="py-8 border-b border-grey-100" :class="{ hidden: !showEquivalentContributionAmount }">
-        <div class="flex gap-x-4 justify-end">
+      <div class="py-8 border-b border-grey-100" :class="{ hidden: hideEquivalentContributionAmount }">
+        <div v-if="equivalentContributionAmount" class="flex gap-x-4 justify-end">
```
gdixon, line 133: If `equivalentContributionAmount` was `0` then `hideEquivalentContributionAmount` would be `true` so `equivalentContributionAmount` here is always going to hold a value and I don't think that the spinner path would ever be followed?

gdixon (changes requested):
I think that the error was resolved here (sorry @mk1r - I missed this issue!): https://github.com/dcgtc/dgrants/pull/229/files#diff-3ee3d8fbcd5124bc277e2f180b8b89cc259db8aac2de24ebcd02c5c2de90244aR204 

The rest of this PR looks a lot better though, I prefer `hideEquivalentContributionAmount` 👍

app/src/views/Cart.vue
mk1r: Ah cool, so this was an attempt to fix the loading state that was occurring in #233 - I personally never actually saw it, but @mds1 and @phutchins were seeing it. This was more of a head start to get it fixed. The ideas was that if it wasn't loaded yet, it would show the loading spinner. (see line 38 here for reference: https://github.com/dcgtc/dgrants/blob/main/app/src/views/Home.vue)

If #233 is resolved by #229 already, then we can discard `2ad124d` and keep `b6b892f` (code refactor).

mk1r: @mds1 since you experience #233, can you verify if it's resolved in main (or this PR)?

mds1: Sure, I can try tomorrow! If your internet is too fast for you to see it (I'm jealous), you could always simulate slower speeds with the browser dev tools or just modify the component state to force the app into the loading state

mds1: (also, if you don't put back ticks around the commit hashes, GitHub will link to them and format it with the code style anyway, which is handy, e.g. `b6b892f` vs b6b892f)

mds1: So I think @gdixon is right here that we never hit that branch. I don't see this resolved in #229 or here, because it's not really an error but more of a UX thing. Here's a slowed down gif showing what happens. You'll notice that "Estimated matching amount" starts at zero until it's loaded, but should show a spinner instead. Similarly, "Equivalent to" is hidden until loaded, but should always be present and also show a spinner

image

mk1r: thanks @mds1 - I'll take a look and simulate slower speeds to test

mk1r:
pushed a cleaner approach and resolved both Equivalent to: / LoadingSpinner & Estimated matching value: / LoadingSpinner
don't need `hideEquivalentContributionAmount` or `showEquivalentContributionAmount` - works by checking `equivalentContributionAmount` alone

mk1r:
but actually need to handle hiding for stablecoins though - don't approve / merge yet, will push another commit soon

app/src/views/Cart.vue
mds1: Sorry if this was unclear @mk1r—two thoughts:
1. I think the loading spinner should be in place of the numbers, not the whole div, otherwise you get a jumpy page from different div sizes. See image below, where loading spinner is taking up a large portion of the page, and the "Equivalent to" text is gone, so you can't tell what's loading
2. Somewhere along the way, the top-level Cart.vue if-else logic broke, as the page-level loading spinner never shows. The whole page is blank while data is loading

image

mds1:
@mk1r Could you also reword the last commit to a more meaningful commit message? Thanks! 🙂

mds1 (approved):

thelostone-mc (approved):

PR #281, opened 2021-09-22
Polygon support

### LINKED ISSUE #273: Polygon Deployment Prep
- [x] ~~Rename `GrantRoundManager.sol` to `GrantRoundManagerUniV3.sol`. This version will be compatible with chains that have a Uniswap V3 deployment~~ — I think this rename is unnecessary and just introduces a larger diff with minimal benefit, so am leaving the original contract name as-is
- [x] Copy/paste `GrantRoundManagerUniV3.sol`, rename to `GrantRoundManagerUniV2.sol`, modify it, and add tests. This version will be compatible with chains that do not have have a Uniswap V3 deployment and only have Uniswap V2 or a forks of it
- [x] Remove all networks except Mainnet, Hardhat, Rinkeby, and Polygon from `chains.ts`
- [x] ~~Update the `checkout` method to use the appropriate method/syntax based on `DGRANTS_CHAIN_ID`~~ this was not required, we just needed to changing how the quote method behaved and the rest of the checkout code should now work as-is
- [x] Update `quoteExactInput` to use the appropriate method/syntax based on `DGRANTS_CHAIN_ID`
- [x] Update #265 with any resulting tech debt resulting from this issue (e.g. contract architecture is now suboptimal)

mds1:
Closes #273 

## Changes

- Adds `GrantRoundManagerUniV2.sol` and associated tests. This is a copy/paste of `GrantRoundManager.sol` with modifications to call to the SushiSwap router on Polygon (using SushiSwap + Polygon is hardcoded for now because IMO this seems like what we'll want to do)
    - Diff of the new `GrantRoundManagerUniV2.sol` with the original  `GrantRoundManager.sol` can be found here: https://www.diffchecker.com/GdoPeEz7
- Updates `chains.ts` to only support Mainnet, Hardhat, Rinkeby, and Polygon Mainnet. Polygon is so cheap that it's not worth the hassle of using their Mumbai testnet. According to https://www.polygongasstation.com/, a token swap is $0.005, so the cost of mainnet testing will be significantly less than the cost of hours spent supporting Mumbai 
- Updates. the `quoteExactInput` method of `cart.ts`, which was the only change needed for checkout code to compile

## Things not changed
- We did not yet deploy a set of contracts to polygon
- We have not tested that polygon checkout code works (because dependent on deployment)

Both of those will be handled in the next PR

## What to review

- Review all changes above
- Look for any tech debt introduced that's not documented in #265 
- Make sure Rinkeby still works and that there have been no regressions


contracts/test/__snapshots__/gas.test.ts.snap
```diff
@@ -12,14 +12,14 @@ exports[`dGrants gas tests GrantRegistry updateGrantPayee 1`] = `41527`;
 
 exports[`dGrants gas tests GrantRoundManager createGrantRound 1`] = `968360`;
 
-exports[`dGrants gas tests GrantRoundManager five donations input token DAI, output token DAI 1`] = `241373`;
+exports[`dGrants gas tests GrantRoundManager five donations input token DAI, output token DAI 1`] = `224273`;
```
mds1, line 15: Even though we didn't touch the GrantRoundManager or its tests, the new test commands result in `GrantRoundManager` tests being run in a different order. I believe that new ordering is why the snapshots changed. Previously the gas snapshots ran first, but now they run last, so there are likely some state changes that persisted between test files as a result of us often not using fixtures (fixtures would revert the state after each test)

This is also why we made the change about how to compute `startTime` in `gas.test.ts`

contracts/contracts/GrantRoundManagerUniV2.sol
```diff
@@ -0,0 +1,261 @@
...
+  constructor(GrantRegistry _registry, IERC20 _donationToken) {
+    // Validation
+    require(_registry.grantCount() >= 0, "GrantRoundManager: Invalid registry");
+    require(_donationToken.totalSupply() > 0, "GrantRoundManager: Invalid token");
+
+    // Set state
+    registry = _registry;
+    donationToken = _donationToken;
+
+    // Token approvals of common tokens
```
apbendi, line 81: Still testing things out and working on a full review, but quick question: why not just extend from `UniswapV2Router02`, the same way we did for v3, to avoid these kinds of issues? Is it a significant lift technically? Or is there some other downside to doing so?

mds1: Mainly because UniswapV2 was written with Solidity 0.6.6 😕

apbendi: Got it. I'd be curious what the upgrade would take. But we can hold off for now.

In the meantime, I wonder if instead of pre-approving select tokens in the constructor, we check for existing approval and conditionally approve inside the donate method? This adds some gas (especially for the first donator of a given token) but given we're hardcoding token addresses for Polygon as it is, and Polygon is a very low fee environment, it seems a reasonable tradeoff to make. It also gives us more flexibility token choice wise.

contracts/package.json
```diff
@@ -65,10 +66,14 @@
...
-    "test": "hardhat test",
+    "test:mainnet": "yarn set-fork:mainnet && hardhat test test/*.test.ts",
+    "test:polygon": "yarn set-fork:polygon && hardhat test test/polygon/*.test.ts; yarn set-fork:mainnet",
+    "test": "yarn test:polygon && yarn test:mainnet",
     "typechain": "hardhat typechain",
     "app:node": "ALCHEMY_API_KEY=$(grep ALCHEMY_API_KEY .env | cut -d '=' -f2) && yarn hardhat node --fork \"https://eth-mainnet.alchemyapi.io/v2/$ALCHEMY_API_KEY\"",
-    "app:setup": "hardhat run scripts/app.ts --network localhost"
+    "app:setup": "hardhat run scripts/app.ts --network localhost",
+    "set-fork:polygon": "sed -i -e 's/eth-mainnet.alchemyapi.io/polygon-mainnet.g.alchemy.com/g' ./hardhat.config.ts && rm -f ./hardhat.config.ts-e",
+    "set-fork:mainnet": "sed -i -e 's/polygon-mainnet.g.alchemy.com/eth-mainnet.alchemyapi.io/g' ./hardhat.config.ts && rm -f ./hardhat.config.ts-e"
```
apbendi, line 76: I had an idea for an alternative way of managing this, but wasn't sure it would work, so I coded it up to see: https://github.com/dcgtc/dgrants/commit/256728654705270f775090401b71e6746cdc5d03

This approach uses ENV vars which are defined inline with the package.json commands to drive the fork URL. It feels cleaner to me. What do you think? Feel free to cherry-pick the commit onto this branch if you like the approach.

apbendi: You could also define alternative blockheights in the same if clause if you wanted, though use the same one seems to work fine.

apbendi:
Left some comments, but overall first pass review is looking good. Was not able to find any regressions on Rinkeby.

I agree after merging, we should do a Polygon deployment to test with (even if the "test" is on polygon mainnet, with low gas fees it seems reasonable to go straight to that). To do that we'll need to update the deploy script to deploy the correct round manager based on the network, right? I can help out with this tomorrow, either as a commit on this PR or a new one.

contracts/contracts/GrantRoundManagerUniV2.sol
mds1: > Got it. I'd be curious what the upgrade would take. But we can hold off for now.

Yea I'm also curious, don't recall enough of the breaking changes between the two versions offhand so would have to investigate, probably not worth the time investigating right now until we have a more clear long-term path forward, since like you said the additional cost doesn't matter much on Polygon.

Added in 94446ea542315410bae41811bcae4dc4e8b96bc7

contracts/package.json
mds1: There's no native `sed` on Windows so the env var approach is definitely better for cross-platform support. Cherry-picked your commit and modified/rebased it in 98d8330d811363cadd174ac5a26a5c7018bddc11 to set block number also, since I think using a more recent block is probably ideal. 

FWIW, recently had to use the `sed` trick that we do here to adjust a hardcoded bytecode hash in a solidity contract, which is why I defaulted to it without thinking much about it 😅

mds1:
> Was not able to find any regressions on Rinkeby.

Awesome!

> I agree after merging, we should do a Polygon deployment to test with (even if the "test" is on polygon mainnet, with low gas fees it seems reasonable to go straight to that). To do that we'll need to update the deploy script to deploy the correct round manager based on the network, right? I can help out with this tomorrow, either as a commit on this PR or a new one.

So I actually already did the modifications and and deployed to mainnet polygon. Just opened up this draft PR (not targeting `main` for diff purposes) if you wanted to start checking it out: https://github.com/dcgtc/dgrants/pull/284

mds1:
Except I just realized I forgot to change which artifact is used in the deploy script. So that PR needs some more fixing haha, whoops 🙃

mds1:
Ok, fixed that part of #284 so deploy aspect of that PR is at least ready

app/src/utils/chains.ts
```diff
@@ -145,129 +121,61 @@ const ALL_SUPPORTED_TOKENS_MAPPING: { readonly [chainId: number]: Record<string,
...
-    nativeCurrency: {
-      name: 'Ethereum',
-      symbol: 'ETH',
-      decimals: 18,
-    },
-    grantRegistry: '0xd0F350b13465B5251bb03E4bbf9Fa1DbC4a378F3',
-    grantRoundManager: '0xB40a90fdB0163cA5C82D1959dB7e56B50A0dC016',
-    multicall: '0x5BA1e12693Dc8F9c48aAD8770482f4739bEeD696',
-    rpcUrl: `https://eth-mainnet.alchemyapi.io/v2/${ALCHEMY_API_KEY}`,
+  [SupportedChainId.POLYGON]: {
```
mds1, line 150: This object has lots of wrong data, but there's a fix for it in #284

apbendi (approved):
For what we're trying to do in this PR, I think we're good to merge 👍

thelostone-mc (approved):

mk1r:
looks good to me as well - didn't find any tech debt, but want to give it a second pass in my editor
tests are working and Rinkeby still works

Issue #241, opened 2021-09-15
feat: add clear grant logo + preload image on edit form


thelostone-mc:
#### Description 

- [x] Ability to remove image on edit/create grant page
- [x] Ability to preview/preload the image on the edit form when a user edits the grant

These changes would need to made around:
- Grant Creation
- Grant Edit
- BaseImageUpload

scco:
i spend a bit time to checkout the edit a grant in current main branch and helped out, get my planed UI regarding image-upload / display uploaded images  into  the form to edit a grant.

<img width="1689" alt="Bildschirmfoto 2021-09-16 um 16 10 11" src="https://user-images.githubusercontent.com/569641/133629128-2af495d4-ceca-4099-9810-d102fe0731bb.png">

### Component : ImageRow
- tryed to do the simple work on this already,
you just have to make the delete (X-Icon) click in this component 
is doing outside of the component the "delete"
- when there is no image uploaded, this component can hide.

### Component : BaseUpload
- on click : regular "file explorer have to open" to select an image. 
this needs probably some work to make it work.
- on drag images on it : yea the hard part i guess ... 
but you have the UI - hurray :) 

### Little changes on `GrantRegistryGrantDetail.vue`
- I have noticed some "flex ... css" in line `116` what is not needed. i removed these classes, as this form should take up full width without any left right paddings ... 
- removed on line `120` the text-left class.  now as in App.vue is no extra css code from old times, its not needed anymore.

Happy to help as best as i can on questions - hope this helps a bit - Checkout here :

https://github.com/dcgtc/dgrants/tree/richard-imageupload-ui

mk1r:
[ ] ask a user to provide a large enough image (with different ratios)
[ ] let the user choose the content to crop out of it to fit 16:9 ratio
[ ] resize image on the client side as long as we have source image with high enough quality

PR #282, opened 2021-09-23
redo image upload

thelostone-mc:
#### Description

- Uploads logo to IPFS on input
- Previews IPFS hash of uploaded logo before form submit
- Adds an inline loader to ensure user is aware that image is being uploaded to IPFS
- Ensures image is preloaded on edit screen 
- Has the ability to clear the logo

Untitled

refs: https://github.com/dcgtc/dgrants/issues/241


app/src/components/BaseImageUpload.vue
```diff
@@ -1,16 +1,67 @@
...
+    </div>
+
+    <div v-if="logoURI && !isValid" class="bg-pink p-4 text-white" :id="`${id}-error`">
+      {{ errorMsg }}
+    </div>
+  </template>
+
+  <template v-else-if="isUploading">
+    <span>Uploading to IPFS...</span>
+  </template>
```
mds1, line 38: Seems this div needs to be before the `v-if="!logoURI || !isValid"`, since currently this is never shown if you upload a valid image—`isValid` starts as true by default, so if I upload a valid image, we still land in the above div while the image is uploading, then jump to the last div once uploaded

So it should be

```
<template v-if="isUploading">
...
<template v-else-if="!logoURI || !isValid">
...
<div v-else-if="logoURI" >
```

Also, is it possible to have a case where none of those three conditions are true and nothing ends up being displayed?

app/src/components/BaseImageUpload.vue
```diff
@@ -1,16 +1,67 @@
...
+
+  <!-- Uploaded Image Preview -->
+  <div v-else-if="logoURI" class="py-8">
```
mds1, line 41: How did you decided to use `<template>` for the first two `if` conditions, and `<div>` for the third? Curious if there's a best practice here we're following. The only difference I'm aware of is that `<template>` doesn't render in the DOM, so in this component they seem interchangeable

mk1r:
- tried two images that met the criteria - first one uploaded and displayed, second one uploaded okay but didn't display on the detail page for some reason
- editing the image doesn't seem to work
- when replacing the image and canceling, it doesn't show the preview of the original image

app/src/components/BaseImageUpload.vue
thelostone-mc: - so the first one has 2 divs and it seemed redundant enclosing a div and hence used a template
- the second one which has Uploading to IPFS doesn't need a div and hence used a template (the argument could be made i just apply the if on the span tag
- the last one, there is an enclosing div so i ended up adding it to that instead of wrapping a template.

I'm not sure what the cleanest way here is but open to suggestions

mds1: I don't feel strongly about it, was just curious! I think the guidance is "do you need a div or does a div make sense, if yes `div`, if no `template`"

app/src/components/BaseImageUpload.vue
thelostone-mc: Done

thelostone-mc:
@mds1 + @mk1r I pushed in the changes.
Could you guys give this a review?
The edit metaPtr still doesn't work and I'm not sure what changed here

mk1r:
first issue is fixed now (can upload both)
aside from editing, take a look at the last point: 
> when replacing the image and canceling, it doesn't show the preview of the original image
if it's easy enough, I would show the preview of the original image again so that if someone attempts to change the image, then changes their mind (and cancels), their original image still shows up

mds1 (approved):

mk1r (approved):

PR #284, opened 2021-09-24
Polygon deploys and checkout

mds1:
Changes:
- Deploys contracts to Polygon mainnet
- Updates app to work with Polygon
    - Configure `chains.ts` with Polygon info
    - Change `startBlock` used when scanning for events from 0 to a value specified in `chains.ts`

Notes:
- Creating a new grant redirects to the newly created grant detail page, but the page is empty for a few seconds until the data loads, but with no UI indication that data is loading. Is this a known issue or should we create an issue to track/resolve it?

To test:
- Use the information [here](https://docs.polygon.technology/docs/develop/network-details/network) to add Polygon to MetaMask
- Set `VITE_DGRANTS_CHAIN_ID=137` in `app/.env`
- Clear site data by going to the browser console > Application > Storage > Clear site data (might be a little different for Firefox). This is required because the app is only configured to work with one canonical Chain ID
- Obtain MATIC (the native token on Polygon, instead of ETH) by one of the following approaches:
    1. Use one or more of the various MATIC faucets to obtain small amounts of MATIC (enough for some transactions). Faucets: [1](https://faucet.firebird.finance/), [2](https://macncheese.finance/matic-polygon-mainnet-faucet.php), [3](https://matic.supply/) (they're a bit flaky, so try them all)
    1. Buy MATIC on L1 and [bridge](https://wallet.polygon.technology/bridge) it to Polygon 
    1. [Bridge](https://wallet.polygon.technology/bridge) ETH from L1 to MATIC, use a faucet from (i) to get MATIC, use SushiSwap to swap your bridged WETH to WMATIC, use Etherscan to unwrap WMATIC for MATIC. This is a hassle, but cheaper than the previous option if you already have L1 ETH and no L1 MATIC
    1. DM me your address and I'll send you $5 of MATIC which should last a long time (contract deploys are a few cents, everything else is typically less than a cent)
- Use SushiSwap to swap some of your MATIC fo DAI, USDC, USDT, which is the list of currently supported tokens for checkout
- When sending transactions, bump the suggested gas price a bit so it gets mined faster

TODO (things that can arguably be done in a separate PR)
- Fix estimated amount in screenshot below. There's also `Quote Update Failed` in the console, which is likely related

image


contracts/deploy-history/poc-polygon-2021-09-23T23:52:54.769Z.json
```diff
@@ -0,0 +1,30 @@
...
+  "contracts": {
+    "GrantRegistry": "0x0D39B9BB82d41F33F894FCeb85FbE4777377A092",
+    "GrantRoundManager": "0x6C9C48aE6A74DA5A98CFAD6617cF0B860E112071"
+  },
+  "actions": {
+    "PublishGrantRoundLogo": "QmYMXFkv1LU2YzQPe8LEDR7NrtL1rgNXV5un9Fn9hH765F",
+    "PublishGrantRoundMetadata": "QmPsAXpuwddVQ2w5jfGFZSSW1oTNQCoSJPhpAfhLXyZjUN",
+    "CreateGrantGround": "0xd277a2d9197cf4ddc9985430928f2a611c1fc4ab29c1c8bbc7fa728c83247d7b"
+  }
+}
```
thelostone-mc, line 30: ^ remove this ?

mds1: Done in 04468b29d8299256b55fd1c608f882bda66974f0

app/src/utils/chains.ts
```diff
@@ -101,7 +104,7 @@ const POLYGON_TOKENS = [
   { ...DAI_TOKEN, chainId: SupportedChainId.POLYGON, address: '0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063' },
   { ...USDC_TOKEN, chainId: SupportedChainId.POLYGON, address: '0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174' },
   { ...USDT_TOKEN, chainId: SupportedChainId.POLYGON, address: '0xc2132D05D31c914a87C6611C10748AEb04B58e8F' },
-  { ...WBTC_TOKEN, chainId: SupportedChainId.POLYGON, address: '0x1BFD67037B42Cf73acF2047067bd4F2C47D9BfD6' },
+  // { ...WBTC_TOKEN, chainId: SupportedChainId.POLYGON, address: '0x1BFD67037B42Cf73acF2047067bd4F2C47D9BfD6' },
```
thelostone-mc, line 107: do we want to remove this ?

app/src/utils/chains.ts
```diff
@@ -144,22 +149,26 @@ const ALL_CHAIN_INFO: ChainInfo = {
...
     multicall: '0x5BA1e12693Dc8F9c48aAD8770482f4739bEeD696',
     rpcUrl: `https://eth-mainnet.alchemyapi.io/v2/${ALCHEMY_API_KEY}`,
+    startBlock: 13285176,
   },
   [SupportedChainId.POLYGON]: {
-    bridge: 'https://bridge.arbitrum.io/',
-    explorer: 'https://arbiscan.io/',
-    label: 'Arbitrum',
-    logoUrl: '/arbitrum_logo.svg',
+    bridge: 'https://wallet.polygon.technology/bridge/',
```
thelostone-mc, line 158: do we remove the trailing `/` ?

thelostone-mc (approved):
lgtm ! left a few minor comment

app/src/utils/chains.ts
mds1: Yea I commented it out for simplicity / speed for now so we can get this merged. Details in this commit message: https://github.com/dcgtc/dgrants/pull/284/commits/3eb709c55cd273fbf913e47656085e7d0063c3eb

app/src/utils/chains.ts
mds1: Not necessary since, unlike the etherscan URLs, we don't modify this (we don't even use it currently)

apbendi (approved):
Looking and working well! Nice :)

Here's one thing I noticed, though I suspect it's not specifically related to the Polygon changes (correct me if I'm wrong):

When I created a new grant, then clicked "Continue" on the tx screen, it said "No grant selected" until I refreshed:

Screen Shot 2021-09-24 at 12 53 57 PM

mds1:
Yea I was wondering about that also. For me it populated the data after waiting long enough. Maybe @gdixon knows, here's how I described it above:

> Creating a new grant redirects to the newly created grant detail page, but the page is empty for a few seconds until the data loads, but with no UI indication that data is loading. Is this a known issue or should we create an issue to track/resolve it?

contracts/scripts/deploy-poc.ts
```diff
@@ -31,20 +33,27 @@ const ipfs = createIpfs(process.env.FLEEK_STORAGE_API_KEY!);
 
     // Deploy the GrantRegistry
     const GrantRegistry = await ethers.getContractFactory('GrantRegistry', deployer);
-    const registry = await GrantRegistry.deploy();
+    const registry = await GrantRegistry.deploy(overrides);
```
mk1r, line 36: why do you need `overrides` here?

contracts/scripts/config/deploy-poc.config.ts
```diff
@@ -45,6 +45,19 @@ const params: DeployParams = {
...
     metadataJson,
   },
+  polygon: {
+    donationToken: '0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063', // DAI
+    uniswapFactory: '0xc35DADB65012eC5796536bD9864eD8773aBc74C4',
+    weth: '0x0d500B1d8E8eF31E21C99d1Db9A6444d3ADf1270',
+    metadataAdmin: '0xD2553382a60F121d9b1e35cFC9EBF4870FbCC96F',
+    payoutAdmin: '0xD2553382a60F121d9b1e35cFC9EBF4870FbCC96F',
+    matchingToken: '0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063', // DAI
+    roundStartTime: 1632460000, // Friday, September 24, 2021 05:06:40 AM GMT
```
mk1r, line 55: we can maybe add a comment here that roundStartTime must be a time in the future (might experience issues deploying if start time has passed already), but maybe this can be mentioned in deploy docs instead

contracts/scripts/deploy-poc.ts
```diff
@@ -31,20 +33,27 @@ const ipfs = createIpfs(process.env.FLEEK_STORAGE_API_KEY!);
...
+      roundManager = await GrantRoundManager.deploy(registry.address, networkParams.donationToken, overrides);
+    } else {
+      roundManager = await GrantRoundManager.deploy(
+        registry.address,
+        networkParams.donationToken,
+        networkParams.uniswapFactory,
+        networkParams.weth,
+        overrides
+      );
+    }
```
mk1r, line 54: does this always need to be this way? i.e does polygon not accept uniswapFactory as an input param

contracts/scripts/deploy-poc.ts
mds1: A few lines up in the script you'll notice we hardcode a gas price for polygon, but not any other networks. This is because polygon gas prices are pretty predictable so this can be easier than relying on the node's gas price estimate, whereas for other networks _not_ hardcoding a gas price means we'll use the one returned from the node

contracts/scripts/config/deploy-poc.config.ts
mds1: Yea I agree. More generally, I think we should have a validation step at the beginning of the deploy script. I accidentally did a run of the deploy script that failed on the third contract deploy because `roundStartTime` was in the past, and it'd be nice to know that before doing the other deploys + IPFS uploads. cc @apbendi if you want to tackle this / create an issue for it

contracts/scripts/deploy-poc.ts
mds1: Correct, right now `GrantRoundManagerUniV2` hardcodes the Polygon SushiSwap router address. 

`GrantRoundManager` takes the factory address as an input because it inherits from `SwapRouter` which requires it, but for `GrantRoundManagerUniV2` we do not inherit from the router so this isn't needed. As for why not, see here: https://github.com/dcgtc/dgrants/pull/281#discussion_r715175157

contracts/scripts/config/deploy-poc.config.ts
mk1r: yeah and I think Phillip ran into an issue like this as well - cool, will create an issue for it

mk1r (approved):

Issue #285, opened 2021-09-24
Data fetching improvements


gdixon:
### Description

GrantRound data fetching todos:

- [ ] Watch for `MetadataUpdated` and update the `metaPtr` for the `GrantRound`
- [ ] Watch for Transfers to the `GrantRound` on the `MatchingTokens` contract and update the `GrantRounds` fund balance
- [ ] Only update the `GrantRound` (predictions etc) if an event tells us something has changed (alter `getGrantRound` calls to only happen when they need to)

Issue #267, opened 2021-09-20
CLR not active for DAI grants


frankchen07:
**Describe the bug**
When I add different currencies to the cart, it seems the matching doesn't take for DAI.

**To Reproduce**
1. Go to checkout with at least two grants with both DAI and ETH.

![Screen Shot 2021-09-20 at 16 17 18](https://user-images.githubusercontent.com/7516920/134088914-91135ffc-203e-478f-8e2c-a5ddd9fccb6b.png)

**Expected behavior**
Contributed amount should have a match estimate regardless of currency.

frankchen07:
Additional issue

I believe the Coffee grant and the Fusion Grant doesn't seem to be in the round, while "Octavio Leaving" is, which doesn't make sense why there would be matching for each of those, and vice versa.

PR #286, opened 2021-09-24
data: Fixes data gathering and localstorage

gdixon:
This PR:

- Stores `Grants` as a key-value obj in localStorage to prevent overwriting on update
- Sources `GrantRound`  data from `MetadataUpdated` and `Token.Transfer` events
- Ensures we only calculate and display matching for `Grants` in a `round`

--

Closes: #285
Closes: #267


app/src/utils/utils.ts
```diff
@@ -161,3 +161,77 @@ export async function assertSufficientBalance(tokenAddress: string, requiredAmou
...
+  calls.forEach((call) =>
+    flatMap.push(
+      ...call.fns.map((fn) => {
+        const args = typeof fn === 'string' ? [fn] : [fn.fn, fn.args];
+        return {
+          target: call.target,
+          callData: call.contract.interface.encodeFunctionData(
+            args[0] as string,
+            args[1] as (string | BigNumberish | boolean)[]
+          ),
```
mds1, line 190: Does `getMulticallData` only calls that have up to two args of these specified types?

app/src/views/Cart.vue
```diff
@@ -101,10 +101,12 @@
                   <!-- TODO use real match estimates -->
```
mds1, line 101: Is this TODO comment out of date?

app/src/views/Cart.vue
```diff
@@ -101,10 +101,12 @@
...
                     <span v-for="(clr, index) in clrPredictions[item.grantId]" :key="index">
-                      {{ formatNumber(clr.matching, 2) }} {{ clr.matchingToken.symbol }}
-                      {{ index !== clrPredictions[item.grantId].length - 1 ? '+' : '' }}
-                      <br />
-                      estimated matching
+                      <span v-if="typeof clr.matching === 'number'">
+                        {{ formatNumber(clr.matching, 2) }} {{ clr.matchingToken.symbol }}
+                        {{ index !== clrPredictions[item.grantId].length - 1 ? '+' : '' }}
+                        <br />
+                        estimated matching
```
mds1, line 108: I cleared local storage, built using Rinkeby, then added three items to my cart at 5 DAI each. Two of the three grants still have 0 DAI matching though. Is that because they aren't in a round, or is that a bug? If it's because they aren't in a round, perhaps we should make that more clear (though I guess that's not necessary for the POC)

image

gdixon: We missed adding the logic to check if a grant should be part of a CLR based on `metadata.grants` - this should be good now 🙌

app/src/components/GrantCard.vue
```diff
@@ -51,14 +51,14 @@ import { formatNumber, formatAddress, getEtherscanUrl, pushRoute } from 'src/uti
 // --- Icons ---
 import { Cart2Icon as CartIcon } from '@fusion-icons/vue/interface';
 
-function getTotalRaised(grantId: any) {
+function getTotalRaised(grantId: BigNumberish) {
   const { grantRounds, grantContributions } = useDataStore();
-  const contributions = filterContributionsByGrantId(grantId.value.toString(), grantContributions?.value || []);
+  const contributions = filterContributionsByGrantId(grantId.toString(), grantContributions?.value || []);
```
mds1, line 56: [`BigNumberish`](https://github.com/ethers-io/ethers.js/blob/4166b2753d0d6b230a3a22b5a2b6c4e1906a7af7/packages/bignumber/src.ts/bignumber.ts#L25) can be of type `Bytes` and I'm not sure if that has a `toString()` method. Might want to use this to be safe:

```suggestion
  const contributions = filterContributionsByGrantId(BigNumber.from(grantId).toString(), grantContributions?.value || []);
```

app/src/components/GrantCard.vue
```diff
@@ -71,8 +71,10 @@ export default defineComponent({
   },
   components: { CartIcon },
   setup(props) {
-    const grantId = ref<any>(props.id);
+    const grantId = ref<BigNumberish>(props.id);
```
mds1, line 74: `grantId` type should be a number here (and everywhere else, though I'm not sure if we've been consistent). Typically `BigNumberish` type should only be for function inputs, and variables/returned values should have a stricter type, e.g. we wouldn't want `grantId` to be represented as `Bytes` here just to cast to a number in the HTML template

app/src/components/GrantList.vue
```diff
@@ -89,12 +88,31 @@ export default defineComponent({
   setup(props) {
     const { addToCart, isInCart, removeFromCart } = useCartStore();
 
+    const sortedGrants = ref<Grant[]>([]);
```
mds1, line 91: Having this logic within setup which also relies on the `useSortedGrants` composition function is a bit unconventional/confusing IMO. It's easier to read and reason about when composition functions are standalone methods that can be directly returned in the setup function, e.g. `setup() { return { ...useSortedGrants() } }`.

My preference would be to move the relevant portions of code from the `setup` method into the existed `useSortedGrants` composition function

app/src/store/cart.ts
```diff
@@ -252,17 +252,18 @@ export default function useCartStore() {
...
       // Execute donation
       // The donate function has two different signatures depending on chainId (e.g. GrantRoundManager
-      // vs GrantRoundManagerUniV2 contracts), and the static TS type inference can't tell the difference, so we
-      // ignore the TS error
-      // @ts-ignore
-      return <ContractTransaction>await manager.donate(swaps, deadline, donations, { value });
+      // vs GrantRoundManagerUniV2 contracts)
+      return <ContractTransaction>(
+        await manager.donate(<(SwapSummary & SwapSummaryUniV2)[]>swaps, deadline, donations, { value })
+      );
```
mds1, line 266: Nice TS cleanup here and throughout this file 👌

app/src/store/cart.ts
```diff
@@ -386,26 +388,31 @@ export default function useCartStore() {
...
-          clr_predictions,
-          amount // pass in the donationToken denominated amount
-        );
+        let matching: number | boolean = false;
+        const metadata = grantRoundMetadata.value[round.metaPtr];
+        if (metadata && metadata.grants?.includes(parseInt(item.grantId))) {
+          // all calculations are denominated in the rounds donationToken
+          const roundToken = round.donationToken;
+          // get the predictions for this grant in this round
+          const clr_predictions = getPredictionsForGrantInRound(item.grantId, grantRoundsCLRData.value[round.address]);
```
mds1, line 397: Nit, but would be nice rename to `clrPredictions` for variable name consistency
```suggestion
          const clrPredictions = getPredictionsForGrantInRound(item.grantId, grantRoundsCLRData.value[round.address]);
```

app/src/store/cart.ts
```diff
@@ -386,26 +388,31 @@ export default function useCartStore() {
...
+          // if contribution/donationToken token is DAI we can skip that step of the conversion
+          const contributionIsDai = token.symbol === 'DAI';
+          const roundTokenIsDai = roundToken.symbol === 'DAI';
+          // take the initial contributionAmount and convert it to be denominated in donationToken
+          let amount = item.contributionAmount;
+          // convert amount to the donationToken (double hop to get into dai then into the donationToken)
+          amount =
+            contributionIsDai || contributionIsRoundToken ? amount : getConvertedAmount(amount, token.address, 1);
+          amount =
+            roundTokenIsDai || contributionIsRoundToken ? amount : getConvertedAmount(amount, roundToken.address, -1);
```
mds1, line 409: I don't follow the logic in this section. Why do we convert the amount in the `1` direction then back in the `-1` direction? I think the comment is saying regardless of token, we normalize the amount by first converting the price to DAI, then converting the DAI amount to be denominated in `donationToken`. But why go through DAI in the first place?

app/src/store/data.ts
```diff
@@ -46,22 +49,27 @@ export default function useDataStore() {
     // Get blockdata
     lastBlockNumber.value = BigNumber.from(await provider.value.getBlockNumber()).toNumber();
 
+    // Get computed on grantRoundManager
+    const grantRoundManager = computed(() => _grantRoundManager.value);
```
mds1, line 53: Can we just use the imported `grantRoundManager` from the wallet store directly? It's returned as a computed property to begin with, so it seems here we just wrap it in another computed property

app/src/store/data.ts
```diff
@@ -85,100 +100,92 @@ export default function useDataStore() {
...
-      timeout = setTimeout(function () {
-        timeout = undefined;
-      }, 1000);
+      try {
+        await rawPoll(forceRefresh);
+        timeout = setTimeout(function () {
+          timeout = undefined;
+        }, 1000);
+      } catch (e) {
+        console.log('dGrants: Data fetch error - please check your network -- ', e);
```
mds1, line 164: Do we want to throw the error after catching it/logging it, that way it surfaces to the user as an alert in the top right? Or maybe we don't want to show this, depending on when you saw this and what prompted you to add it. Also worth mentioning another option is to add the error message to the `messagesToIgnore` array

```suggestion
        console.log('dGrants: Data fetch error - please check your network -- ', e);
        throw e;
```

app/src/store/data.ts
```diff
@@ -85,100 +100,92 @@ export default function useDataStore() {
...
+  async function startPolling() {
     provider.value.removeAllListeners(); // remove all existing listeners to avoid duplicate polling
     provider.value.on('block', (/* block: number */) => void poll());
+    // record the network value to detect changes
+    const networkValue = (await getStorageKey('network'))?.data || network.value;
+    // watch the network
+    watch(
+      () => [network.value],
+      async () => {
+        // clear storage and poll again for all network changes after first load
```
mds1, line 181: Could you clarify what exactly we are clearing here on network changes? Want to make sure we aren't clearing the cart data, because a user might be not connected / connected to a testnet while adding things to their cart, switch to mainnet to checkout, and we don't want them to lose their cart in that case

app/src/store/wallet.ts
```diff
@@ -177,7 +177,7 @@ export default function useWalletStore() {
     network.value = markRaw(_network);
 
     // Start polling for data
-    startPolling();
+    poll();
```
mds1, line 180: Why `poll` instead of `startPolling`? If a user changes networks, will this remove old listeners? I think it still will, but want to confirm

app/src/types/index.ts
```diff
@@ -20,17 +20,23 @@ export type CartItem =
...
+};
+
 // Generic localStorage type to store key->any
-export type LocalStorageAnyObj = {
+export type LocalForageAnyObj = {
   [key: string]: any; // eslint-disable-line @typescript-eslint/no-explicit-any
 };
-export type LocalStorageData = {
+export type LocalForageData = {
   ts?: BigNumberish;
```
mds1, line 39: What is `ts`? I'd suggest a more descriptive variable name or a comment

app/src/utils/constants.ts
```diff
@@ -34,12 +36,18 @@ export const ERC20_ABI = [
 
 export const LOREM_IPSOM_TEXT = 'A brief description of your grant goes here.\n\nLorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua.'; // prettier-ignore
 
-// LocalStorage keys
+// LocalForage config
+export const localForageConfig: LocalForageConfig = {
+  name: 'dGrants',
+  version: 1,
```
mds1, line 42: Should this correspond to the `package.json` version number, in which case we can import that like we do in the `About` component, or is this separate?

app/src/utils/data/contributions.ts
```diff
@@ -20,30 +20,29 @@ const { grantRoundManager } = useWalletStore();
...
     {
       blockNumber: blockNumber,
-      timeStamp: Date.now() / 1000,
     },
-    async (localStorageData?: LocalStorageAnyObj | undefined, save?: (saveData?: LocalStorageAnyObj) => void) => {
-      const rawData = localStorageData?.data || {};
-      const contributions: Record<string, Contribution> = rawData?.contributions || {};
+    async (LocalForageData?: LocalForageAnyObj | undefined, save?: (saveData?: LocalForageAnyObj) => void) => {
+      // pick up contributions from the localStorage obj
+      const ls_contributions: Record<string, Contribution> = LocalForageData?.data?.contributions || {};
```
mds1, line 40: Similar comment to above on preferring `lsContributions` vs `ls_contributions`

app/src/utils/data/contributions.ts
```diff
@@ -120,55 +110,56 @@ export async function getTrustBonusScores(
...
       if (
         forceRefresh ||
-        !localStorageData ||
-        (localStorageData && (localStorageData.blockNumber || START_BLOCK) < blockNumber)
+        !LocalForageData ||
+        (LocalForageData && (LocalForageData.blockNumber || START_BLOCK) < blockNumber)
       ) {
         // set score for null user to prevent linear from fetching (if there are no other scores matched)
-        trustBonusData['0x0'] = 0.5;
+        ls_trustBonus['0x0'] = 0.5;
```
mds1, line 127: Naming convention again (note that I'm open to keeping these underscores in variable names if there's a convention I'm missing)

app/src/utils/data/grantRounds.ts
```diff
@@ -36,20 +36,19 @@ export async function getAllGrantRounds(blockNumber: number, forceRefresh = fals
...
     },
-    async (localStorageData?: LocalStorageData | undefined, save?: () => void) => {
+    async (LocalForageData?: LocalForageData | undefined, save?: () => void) => {
       const { grantRoundManager } = useWalletStore();
       // use the ls_blockNumber to decide if we need to update the roundAddresses
-      const ls_blockNumber = localStorageData?.blockNumber || START_BLOCK;
+      const ls_blockNumber = LocalForageData?.blockNumber || START_BLOCK;
       // only update roundAddress if new ones are added...
-      let roundAddresses = localStorageData?.data?.roundAddresses || [];
+      const ls_roundAddresses = LocalForageData?.data?.roundAddresses || [];
```
mds1, line 46: Same thing on naming convention

mds1 (commented):
Haven't finished reviewing everything yet, but @gdixon here's my first set of comments!

app/src/utils/data/grantRounds.ts
```diff
@@ -87,120 +86,136 @@ export async function getGrantRound(blockNumber: number, grantRoundAddress: stri
...
-        // collect the donationToken before promise.all'ing everything
-        const donationTokenAddress = data?.donationToken?.address || (await roundContract.donationToken());
-        const matchingTokenAddress = data?.matchingToken?.address || (await roundContract.matchingToken());
-        // use matchingTokenContract to get balance
-        const matchingTokenContract = new Contract(matchingTokenAddress, ERC20_ABI, provider.value);
-
+    async (LocalForageData?: LocalForageData | undefined, save?: () => void) => {
+      const { provider } = useWalletStore();
+      // use the ls_blockNumber to decide if we need to update the rounds data
+      const ls_blockNumber = LocalForageData?.blockNumber || 0;
```
mds1, line 94: We should never default to zero as the start block, since it's bad for performance. Instead let's use the `START_BLOCK` exported from `chains.ts`

app/src/utils/data/grantRounds.ts
```diff
@@ -231,98 +246,112 @@ export async function getGrantRoundGrantData(
...
       blockNumber: blockNumber,
     },
-    async (localStorageData?: LocalStorageData | undefined, save?: () => void) => {
-      const { provider } = useWalletStore();
-      const roundGrantData = localStorageData?.data?.grantRoundCLR || {};
+    async (LocalForageData?: LocalForageData | undefined, save?: () => void) => {
+      // unpack round state
+      const roundAddress = grantRound.address;
+      const matchingTokenDecimals = grantRound.matchingToken.decimals;
+      const totalPot = BigNumber.from(grantRound.funds).toNumber();
```
mds1, line 257: Sanity check since this is converting to a number: `grantRound.funds` is a human-readable number, e.g. 1 DAI, not a parsed number like 1e18 DAI?

mds1 (commented):
Ok, finished with my review!

app/src/store/cart.ts
gdixon: This is for covering the situation where we need to convert the actual donation token in to the GrantRounds `donationToken` so that we can estimate with an approximation of the output of the swap (say we're contributing `UNI` but the rounds `donationToken` is set to `GTC`). 

I couldn't see another way to get into the desired `donationToken` other than adding every combination of path, is there an easier way that I'm missing?

contracts/test/utils.ts
```diff
@@ -10,7 +10,7 @@ import { Log } from '@ethersproject/providers';
 import { abi as UNISWAP_POOL_ABI } from '@uniswap/v3-core/artifacts/contracts/interfaces/IUniswapV3Pool.sol/IUniswapV3Pool.json';
 import { FeeAmount, Pool, Route, computePoolAddress, encodeRouteToPath } from '@uniswap/v3-sdk';
 import { Token } from '@uniswap/sdk-core';
-import { Grant } from '@dgrants/types';
+import { GrantEthers } from '@dgrants/types';
```
thelostone-mc, line 13: why are using GrantEthers ? not super clear on this change

```
export type GrantEthers = GrantArray & GrantObject;
export type Grant = GrantObject | GrantEthers;
```

contracts/scripts/update-round.ts
```diff
@@ -6,7 +6,7 @@ import params from './config/update-round.config';
 
 const { ethers } = hre;
 
-const ipfs = createIpfs(process.env.FLEEK_STORAGE_API_KEY!);
+const ipfs = createIpfs(process.env.FLEEK_STORAGE_API_KEY!); // eslint-disable-line @typescript-eslint/no-non-null-assertion
```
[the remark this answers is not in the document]
gdixon: This was choking when I ran `yarn lint` from root - not required, but nice to have 🤷

contracts/test/utils.ts
gdixon: Because `Grant` (that we use internally) now uses `id: number`, but we still need to _type_ the same data when we pull it from chain before we convert `GrantEthers` to `Grant`

app/src/utils/data/grantRounds.ts
gdixon: Yes, `totalPot`/`funds` should always be human-readable!

app/src/utils/data/grantRounds.ts
gdixon: This makes sense! 🙌

app/src/utils/constants.ts
gdixon: This version is used in `indexedDb` to version the db structure and allow for a migration phase, I think in `localForage` we can still use it like that and force cache-busting on the end-users if we bump the version.

I've added some comments inline to explain this thinking 👍  _(to be tested!)_

app/src/types/index.ts
gdixon: I've removed `ts` -- we was using it to throttle refreshes on the cache (by time) but everything uses the blockNumber now 👍

app/src/store/wallet.ts
gdixon: I've restored the original.

I was seeing two `poll`s happening concurrently and I was trying to cut out the second one - now that we've removed the `.on('block')`, we need to investigate if this is still happening.

It could be that its just `poll` without a wallet connected and one with the wallet - but its going to result in an extra set of network requests we don't need to be making (especially if we're going to be using the graph)!

app/src/store/data.ts
gdixon: I've removed this try/catch! 🙌

app/src/components/GrantCard.vue
gdixon: I've changed grantId to be a number everywhere 👍

app/src/components/GrantCard.vue
gdixon: I've made this change! 🙌

app/src/components/GrantList.vue
gdixon: I've reverted this change 👍

app/src/store/data.ts
gdixon: Yes 🙌 😄

app/src/store/data.ts
gdixon: We're clearing anything that goes into localForage, so all cart data should be retained 🙌

app/src/utils/data/contributions.ts
gdixon: I've updated to follow the convention you've set in `cart.ts` (for localStorage versions of data) 👍

app/src/utils/data/grantRounds.ts
gdixon: I've added a comment for this and changed its type to `number`

app/src/store/wallet.ts
mds1: In `startPolling`, I think if we just import `provider` and `defaultProvider` from the wallet store and use the below snippet that should remove the duplicate polling

```typescript
defaultProvider.removeAllListeners(); // provider used when no wallet is connected
provider.value.removeAllListeners(); // user's new provider, probably don't need this line but cant hurt
```

Looking at our old implementation, I think I configured the remove listeners call wrong so it was never removed from the default provider

contracts/scripts/update-round.ts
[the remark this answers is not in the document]
mds1: This works because we polyfill `process.env`, but the Vite way to do this would be 
```suggestion
const ipfs = createIpfs(import.meta.env.FLEEK_STORAGE_API_KEY);
```

This is how we import env vars elsewhere in the app, and I think would remove the need for this TS change. https://vitejs.dev/guide/env-and-mode.html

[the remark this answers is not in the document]
mds1: Oh wait nevermind we're in the `contracts` dir here, sorry!

mds1 (approved):

Issue #292, opened 2021-09-27
In `chains.ts`, the `rpcUrl` added for Polygon should not use our API key


mds1:
Instead, it should use the default of `https://polygon-rpc.com/` suggested by https://docs.polygon.technology/docs/develop/network-details/network/

PR #294, opened 2021-09-27
RPC URL change + big number fix + cart UX improvements

### LINKED ISSUE #234: Cart: when checking out, there is no UI indication of pending approval transactions
Steps to reproduce:
1. Add items to your cart
2. Select a token as the currency
3. Checkout
4. Send approval transaction and notice that the app shows nothing to indicate there is a pending transaction
5. Once approval transaction is mined, cart checkout transaction comes up next

This is much more evident on Rinkeby than localhost since transaction take longer to mine

### LINKED ISSUE #291: bug: cart page crashes when input amount is too large
**Describe the bug**

Screen crashes when user enters number too large on the checkout page.
This is due to big number conversion.

### LINKED ISSUE #293: Batch Log Requests To Avoid "Block Range Too Large" Errors
Batch requests to 100,000K any time we fetch events.....

mds1:
Closes #234, closes #291, closes #293

Changes:
- Cart now shows approval transaction info when checking out (#234). The loading spinner is currently static and does not rotate, as @scco said he can add the rotation in / provide a snippet
- Changes the RPC URL used by Polygon, that way when users add Polygon to MetaMask they aren't using our API key (#292)
- For numbers larger than 1e21, JS formats the number as 1e+21 instead of 1000...000, which was causing the crash described in #291. This crash was because the call to `commify` failed since it doesn't support scientific notation. In those cases, we now catch the error and convert the number to a formatted string using `.toLocaleString()`. In reality, we don't want users checking out with numbers above `Number.MAX_SAFE_INTEGER` (2^53-1) anyway due to precision loss, and in reality this will not happen because no one owns that many tokens. So the fix implemented is mainly a UX thing to avoid the app crashing, and should have no practical impact

### DIFF
--- a/app/src/store/cart.ts
+++ b/app/src/store/cart.ts
@@ -68,6 +68,7 @@ const toHex = (val: BigNumberish) => BigNumber.from(val).toHexString();
 const lsCart = ref<CartItemOptions[]>([]); // localStorage cart
 const cart = ref<CartItem[]>([]); // source of truth for what's in the user's cart
 const quotes = ref<Record<string, number>>({}); // mapping from token address to DAI exchange rate, i.e. multiple token quantity by the exchange rate to get the value in DAI
+const cartStatus = ref<string>('');
 
 // --- Composition function for state management ---
 export default function useCartStore() {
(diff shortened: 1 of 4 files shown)


mk1r:
Should it continue showing pending when user exits Metamask confirmation (i.e denies transaction)?

Screen Shot 2021-09-27 at 5 47 02 PM

mds1:
@mk1r Good catch, fixed in adf6236

thelostone-mc (approved):

thelostone-mc (approved):

PR #295, opened 2021-09-27
button spinner animation

scco:
- use the `rotate-icon` class to spin the icon inside a button
- use the `disable` class to style a disabled button
- use whatever icon inside the button as we do already

`<button class="btn rotate-icon disabled"><SpinnerIcon/></button>
`

example using the "Spinner1Icon as SpinnerIcon" icon from fusion interface
`import { Spinner1Icon as SpinnerIcon } from '@fusion-icons/vue/interface';`


mds1:
Cherry-picked this commit into #294

PR #296, opened 2021-09-27
Custom Top LVL Grant Filter

### LINKED ISSUE #253: Implement top level grant filter

mk1r:
closes #253 

to test:

- open app/src/CustomGrantIndexList.json and add custom index list I.E. [0, 2, 3, 5]
- check the app to see that these are now the only grants shown (and can be sorted)
- when list is empty, all grants are shown

Review:

- which folder to have CustomGrantIndexList.json in (current location: app/src/)
- consider whether this is fine or shall we store in IPFS or ENV variable


thelostone-mc:
As opposed to having the Customlist as a file !
@mk1r might i suggest using fleek + IPNS instead?
That way if we wouldn't have to do a deploy 
- if we want to not include a grant the first time
- and following that any new valid grant is created which we want on the UI

You could just fetch the output from 
https://storageapi.fleek.co/thelostone-mc-team-bucket/whitelist-grants.json

cc @mds1 + @phutchins 


Untitled

app/src/components/GrantList.vue
```diff
@@ -23,12 +23,20 @@ import GrantCard from 'src/components/GrantCard.vue';
 import useCartStore from 'src/store/cart';
 // --- Types ---
 import { FilterNavButton, FilterNavItem, Grant, GrantMetadataResolution } from '@dgrants/types';
+// hard-coded grant ID List
+import grantIDList from 'src/CustomGrantIndexList.json';
```
mds1, line 27: I believe we said we'd fetch this file from a URL to avoid having to redeploy the app every time we want to change the list. So it'd be hosted on e.g. github in a repo (TBD what repo) and do something like this. We could probably just keep it in this repo for now

```typescript
const jsonFetch = (url: string) => fetch(url).then((res: any) => res.json());
const url = 'https://raw.githubusercontent.com/dcgtc/dgrants/top-level-grant-filter/app/src/CustomGrantIndexList.json'
const grantIdList = await jsonFetch(url);
```

mds1:
> As opposed to having the Customlist as a file !
> @mk1r might i suggest using fleek + IPNS instead?
> That way if we wouldn't have to do a deploy [when things change]

IIRC, in a standup / in #253, we decided to not use fleek and/or IPNS because it'd take more dev work to setup and doesn't matter much for the MVP. However, I'm not that familiar with Fleek + IPNS and @thelostone-mc if you already know how to do this easily (seems you do), then maybe we just switch to that approach. 

My suggestion would be the one mentioned above (https://github.com/dcgtc/dgrants/pull/296#discussion_r717544591), which gives the benefit of not needing to redeploy to change the ID list without the overhead of IPNS. 

Aside from this change, everything else in this PR LGTM and works as expected

app/src/components/GrantList.vue
```diff
@@ -88,13 +95,22 @@ export default defineComponent({
   },
   setup(props) {
     const { addToCart, isInCart, removeFromCart } = useCartStore();
+    const grantList = computed(() =>
+      grantIdList.value.length > 0 ? createCustomGrantList(props.grants) : props.grants
+    );
+
+    onMounted(async () => {
+      const url =
+        'https://raw.githubusercontent.com/dcgtc/dgrants/top-level-grant-filter/app/src/CustomGrantIndexList.json';
```
mds1, line 104: @phutchins @apbendi Any idea where we want to host these JSON files? (plural, since we'll need one per network)

mk1r: also, quick note to self: if we choose to do it this way, need to remember to change branch in url from 'top-level-grant-filter' to 'main'

mds1: We could also format the JSON file by chain ID so we only need one file, e.g.

```json
{
  "4": [0,2,7],
  "137": [1,5,6, 9],
}
```

and the app would use `DGRANTS_CHAIN_ID` to find the correct array

mk1r: +1 to that idea - should we wait for input? maybe bring this up in tomorrow's standup

phutchins:
We should upload the file to Fleek. @phutchins will set up the URL and add the right people to our Fleek account. We should use the format above that allows for all networks in a single JSON file.

thelostone-mc:
@mk1r I've updated https://storageapi.fleek.co/thelostone-mc-team-bucket/whitelist-grants.json with the new format

app/src/CustomGrantIndexList.json
```diff
@@ -0,0 +1 @@
+[]
```
thelostone-mc, line 1: Do we need this file ?

thelostone-mc (approved):

phutchins (approved):

PR #300, opened 2021-09-29
wallet modal style update

### LINKED ISSUE #288: reskin / style the wallet connect modal
Bildschirmfoto 2021-09-24 um 18 38 41

haven't looked in the codebase for this modal, 
but i would love to try to reskin it to fit our minimal layout.

as i have basicaly all the icons for common web3 providers, would love to use these too.

scco:
closes #288

the style-ability for the Onboard Lib is sadly not good enough - same for their outdated documentation of css classes to throw styles on - what in most cases not works as expected.

- there are hardcoded styles i can not modify
- some svelte compiled extra classes i can not target at all as they change names on each build
- theres no way to adjust general behaviours / templates / components to play with grids, other dom structures that would me allow to add custom icons . 

i disagree on @mds1 that users would be confused on other icons, and think that this love to detail would give a more appreciated experience. its no usability issue but a loss for a high quality experience.

so i just could tweak some basics, but would love to revisit this topic when we have time for this level of detail.

Bildschirmfoto 2021-09-29 um 15 22 06

see here how these icons would look in a modal i did for web3modal

https://schumanncombo.com/web3modal/

### DIFF
--- a/app/src/index.css
+++ b/app/src/index.css
@@ -19,6 +19,12 @@
     scroll-behavior: smooth;
   }
 
+  h1 {
+    @apply text-headline;
+  }
+
+  /********* NOTIFICATION STYLES */
+
   .bn-notify-custom {
     margin-left: 0px;
   }
@@ -59,9 +65,7 @@
     display: none;
   }
 
-  h1 {
-    @apply text-headline;
-  }
+  /********* INPUT STYLES  */
 
   [type="text"],
   [type="email"],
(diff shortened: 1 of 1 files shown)


mds1 (approved):
This looks really good to me, could even be the final styling of the modal IMO! But if you wanted to style it more and the current classes don't give you enough flexibility, I'd recommend creating an issue in the onboard repo explaining what you need. They're pretty responsive to ideas/suggestions: https://github.com/blocknative/onboard

thelostone-mc (approved):

PR #308, opened 2021-09-29
feat: add walletconnect support

### LINKED ISSUE #275: Change wallet model should give the ability for a user to select different providers
**Describe the bug**
After connecting a wallet via metamask, when a user clicks `Change Wallet` they are shown a modal that does not have an option to reselect another wallet from the current provider. For usability we may want to add a button that allows the user to disconnect from their current metamask account in the modal.

**To Reproduce**
Connect a wallet using metamask. In the account dropdown select `Change Wallet`, no other options are given for a user to reselect metamask and choose a different account via another provider.

mds1:
Closes #275

To test, use connect with WalletConnect (WC) on Rinkeby and create a grant and/or checkout your cart. My current attempts at WC testing:
- Rainbow wallet's Rinkeby connection didn't seem to be working (the chainId is logged as 1)
- MetaMask connects to chainId but throws `TypeError: Failed to fetch` when I try to send a transaction. (I tried MetaMask + WC + Rinkeby on Uniswap and it throws with `Could not deposit invalid opcode: opcode 0xfe not defined`, so I don't think the problem is on our side)

 Any recommendations on other WC wallets that support Rinkeby are welcome!

### DIFF
--- a/app/src/store/wallet.ts
+++ b/app/src/store/wallet.ts
@@ -23,7 +23,7 @@
 import { computed, ref, markRaw } from 'vue';
 import useDataStore from 'src/store/data';
 import useSettingsStore from 'src/store/settings';
-import { CHAIN_INFO, DGRANTS_CHAIN_ID, GRANT_REGISTRY_ADDRESS, GRANT_ROUND_MANAGER_ADDRESS, MULTICALL_ADDRESS, RPC_URL, SupportedChainId } from 'src/utils/chains'; // prettier-ignore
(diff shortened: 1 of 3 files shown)


app/src/store/wallet.ts
```diff
@@ -166,6 +172,7 @@ export default function useWalletStore() {
 
     // Get ENS name if user is connected to mainnet
     const chainId = _provider.network.chainId; // must be done after the .getNetwork() call
+    console.log('connected on chainId: ', chainId);
```
mds1, line 175: This is temporary just to help you verify the network you're connected to with walletconnect, and should be removed before merging

mk1r (approved):
connecting with Rainbow wallet works fine

thelostone-mc (approved):

PR #335, opened 2021-10-06
BaseFilterNav optional prop : title  closes #329

### LINKED ISSUE #329: It is not clear which section is Grants and which is Grant Rounds on the Home Page
When visiting the home page in dGrants, there are two sections, one with Grants and one with GrantRounds. In our UI designs we have filter presets on the top left of each of the sections (like Active Rounds (2)    All Rounds (3)) indicating what that section holds. We have not implemented this yet however and even with these preset filter links in place it still may not be extremely clear what each of the section holds. 

We should add a visual indicator clarifying that the first section is Grant Rounds, and the second is Grants.

scco:
### issue
as we not 100% going with the original design it became unclear on /home
what is a round list and what is a grant list . 

### solution
was already build but got lost / wrongly used - by
adding an optional property `title` to the component `BaseFilterNav`
that conditional show the title to the front of the BaseFilterNav when there is a title property. 

###  bugfix
- the component BaseFilterNav does not need a title property for `BaseFilterNav` in `GrantList.vue` 

closes #329

### DIFF
--- a/app/src/components/BaseFilterNav.vue
+++ b/app/src/components/BaseFilterNav.vue
@@ -5,6 +5,9 @@
     <div class="flex flex-col gap-y-4 md:flex-row gap-x-8">
       <!-- nav items wrapper -->
       <div class="flex flex-wrap whitespace-nowrap content-center gap-x-8 gap-y-4">
+        <!--optional title -->
+        <div v-if="title" class="uppercase font-medium">{{ title }}</div>
+
         <!-- nav item loop -->
         <ul v-for="(item, index) in items" :key="item.label">
           <!-- nav item -->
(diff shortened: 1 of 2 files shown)


mk1r (commented):
based on #329, to resolve it, don't we need to actually add the titles to each section: Grant Rounds and Grants?

app/src/components/BaseFilterNav.vue
```diff
@@ -63,6 +66,7 @@ export default defineComponent({
   name: 'BaseFilterNav',
   components: {},
   props: {
+    title: { type: String, required: false },
```
corydickson, line 69: I feel like the title should be required as to ensure the delineation between the two sections. Is there a use case where we'd rather not have it appear?

thelostone-mc: agree with this

thelostone-mc: @corydickson looks like there are scenrios where this would be used with BaseHeader and a title isn't required

Issue #336, opened 2021-10-06
Handle Speed Up Transactions On Transaction Status Screen


apbendi:
When the user is on the Transaction Status Screen and speeds up their transaction, we don't detect this. As a result, the transaction status screen will hang forever, even after the replacement transaction is successfully mined and executed.

We should look into detecting replacement txs and updating the status screen appropriately, either via ethers support (if present) or with a tool like Blocknative.

PR #343, opened 2021-10-07
feat: adds watchTransaction util to watch for replacement txs

gdixon:
This PR implements a `watchTransaction` mechanism to watch for replacement transactions.

---

Closes: #336

### DIFF
--- a/app/src/components/TransactionStatus.vue
+++ b/app/src/components/TransactionStatus.vue
@@ -77,7 +77,7 @@ function useTransactionStatus(hash: Ref, context: SetupContext<'onReceipt'[]>) {
   // Etherscan URL helpers
   const etherscanUrl = computed(() => getEtherscanUrl(hash.value));
 
-  // On mount, fetch receipt and wait for it to be mined, and emit event with receipt status once mined
+  // watch for new tx hashes then fetch receipt and wait for it to be mined, finally emit event with receipt status once mined
   const { provider } = useWalletStore();
   // if the props.hash changes then we need to await the new tx
   watch(
--- a/app/src/utils/utils.ts
+++ b/app/src/utils/utils.ts
@@ -4,14 +4,13 @@
 import router from 'src/router/index';
 import { RouteLocationRaw } from 'vue-router';
 import { Event } from 'ethers';
-import { BigNumber, BigNumberish, commify, Contract, ContractTransaction, isAddress } from 'src/utils/ethers';
+import { BigNumber, BigNumberish, commify, Contract, ContractTransaction, isAddress, Logger } from 'src/utils/ethers';
 import { BatchFilterQuery, EtherscanGroup } from 'src/types';
 import useWalletStore from 'src/store/wallet';
 import { GrantRound } from '@dgrants/types';
 import { formatUnits } from 'src/utils/ethers';
 import { ETH_ADDRESS } from 'src/utils/constants';
 import { ETHERSCAN_BASE_URL, FILTER_BLOCK_LIMIT, SUPPORTED_TOKENS_MAPPING, WETH_ADDRESS } from 'src/utils/chains';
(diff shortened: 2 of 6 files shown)


app/src/components/TransactionStatus.vue
```diff
@@ -75,15 +75,20 @@ function useTransactionStatus(hash: string, context: SetupContext<'onReceipt'[]>
   );
 
   // Etherscan URL helpers
-  const etherscanUrl = computed(() => getEtherscanUrl(hash));
+  const etherscanUrl = computed(() => getEtherscanUrl(hash.value));
 
   // On mount, fetch receipt and wait for it to be mined, and emit event with receipt status once mined
```
mds1, line 80: Seems this comment is out of date

app/src/utils/utils.ts
```diff
@@ -179,6 +181,43 @@ export async function assertSufficientBalance(tokenAddress: string, requiredAmou
...
+  // tx is picked up from the provided call and updated if replaced
+  let tx: ContractTransaction;
+  // do the call inside a try catch so that we see the error first
+  try {
+    // get the ContractTransaction from the call
+    tx = await call();
+    // provisionally set the tx.hash to move onto next page
+    stateRef.value = tx.hash;
+    // wait for the transaction to be mined
+    void (await tx.wait());
```
mds1, line 203: Why `void` here?

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -552,23 +552,20 @@ function useGrantDetail() {
...
-      tx = await registry.updateGrantPayee(g.id, payee);
+      txCall = () => registry.updateGrantPayee(g.id, payee);
     } else if (owner === g.owner && payee === g.payee && isMetaPtrUpdated) {
       // update Grant MetaPtr
-      tx = await registry.updateGrantMetaPtr(g.id, metaPtr);
+      txCall = () => registry.updateGrantMetaPtr(g.id, metaPtr);
     } else {
       // update all
-      tx = await registry.updateGrant(g.id, owner, payee, metaPtr);
+      txCall = () => registry.updateGrant(g.id, owner, payee, metaPtr);
```
mds1, line 564: Could you document somewhere—even if just with comments where we define `watchTransaction`—what example usage should look like with some pseudocode/code snippets in the comments? Just since this is a different flow than we had before and I want to make sure it's easy to understand without having to find actual usage of the method

app/src/utils/utils.ts
```diff
@@ -179,6 +181,43 @@ export async function assertSufficientBalance(tokenAddress: string, requiredAmou
...
+  } catch (error) {
+    if (error.code === Logger.errors.TRANSACTION_REPLACED) {
```
corydickson, line 205: To address above something similar to this

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -552,23 +552,20 @@ function useGrantDetail() {
...
     }
 
-    txHash.value = tx.hash;
-
-    // After tx mines, poll so the store has the latest data, then navigate to the grant page
-    await tx.wait();
-    // await poll();
+    // watch the transaction to check for any replacements/cancellations and update txHash accordingly
+    void (await watchTransaction(txCall, txHash));
```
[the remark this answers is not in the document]
corydickson: maybe because we aren't doing anything with this return value @gdixon ?

mds1 (approved):

app/src/utils/utils.ts
gdixon: This is being drawn from ethers (https://docs.ethers.io/v5/api/utils/logger/#errors--transaction-replaced), I think its wallet agnostic, @mds1 any ideas here?

Issue #344, opened 2021-10-07
Bug: Custom wallet stylings are lost in production build


gdixon:
**Describe the bug**

Custom styles defined in `index.css` base `@layer` are being purged from the generated `yarn build` files.

**To Reproduce**

Build the app locally and search for the expected classes (`bn-onboard-custom` or `bn-notify-custom`) in the generated `dist` dir

**Expected behavior**

The custom classes should be applied in the build

PR #345, opened 2021-10-07
fix: adds custom classes to safelist to avoid purge

gdixon:
This PR adds custom classes defined in the base @layer of index.css to the purge safelist.

--

Closes: #344


mds1 (approved):

mk1r (approved):
nice

PR #360, opened 2021-10-11
add validation step for roundStartTime when deploying

### LINKED ISSUE #297: move startTime to a time in the future and write this in the docs

metafraction:
closes #297 added deployment note & validation step for roundStartTime

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dgrants/app",
-  "version": "0.0.6-f9d93d6.0",
+  "version": "0.0.7-96aa3a7.0",
   "private": true,
   "scripts": {
     "clean": "rimraf dist",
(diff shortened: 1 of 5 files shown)


contracts/scripts/deploy-poc.ts
```diff
@@ -22,6 +22,12 @@ const ipfs = createIpfs(process.env.FLEEK_STORAGE_API_KEY!); // eslint-disable-l
     if (!networkParams) {
       throw new Error(`Invalid network ${network}`);
     }
+
+    const nowInSeconds = Math.ceil(Date.now() / 1000);
```
mds1, line 26: My only thought would be maybe make the minimum time `nowInSeconds + 60` or something, to add margin since other contracts are deployed first (i.e. setting `roundStartTime` to current time won't work in practice because we'll be after `roundStarTime` by the time the deploy tx executes). But in reality this current logic will likely protect us in the majority of cases, so I'm also ok leaving this as is

thelostone-mc (approved):

PR #390, opened 2021-10-18
first draft of image editor with some changes for test in image uploader

Ajand:
Image Editor is a component that will be used when an image in the Image Uploader doesn't fit in the selected constraint. It will give the user the ability to crop the image into the desired ratio and it will automatically change the image size to fit into the constraint.

specs:
- [x] select ratio
- [x] move image
- [x] zoom image
- [ ] return the selected part of image
- [ ] svg support

### DIFF
--- a/app/src/components/BaseImageUpload.vue
+++ b/app/src/components/BaseImageUpload.vue
@@ -1,19 +1,36 @@
 <template>
-  <div :class="width">
-    <label class="flex flex-col bg-white items-center block px-6 py-8 border border-grey-400 cursor-pointer">
-      <span>Choose image file to upload</span>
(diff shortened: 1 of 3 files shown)


mk1r:
what's the status here?

Ajand:
@mk1r almost finished. Will be done this week I guess.

Ajand:
The component is finished. I created a new screen for testing and styling it at /image-uploader and after that, I will integrate it since the integration will depend on #506 .
@mds1 Is there a way to merge this into the main repo as an image-uploader-editor branch so @scco could style it easier without needing to clone my fork?

scco:
added ui for all of this , but need a little ping-pong session with ajand to finalize

Issue #366, opened 2021-10-12
Change a grant round "funds" to "matching funds" & default "matching" to "linear"


frankchen07:
In some sense, even I got some of them mixed up:

* funds - user was unsure
* payout admin - I accidentally said that it was the payout address, but really it's the payout address of the admin who will distribute the matching round funds
* matching - user was unsure

the biggest ambiguity was the difference between round funds and round matching - wouldn't those be the same?

https://user-images.githubusercontent.com/7516920/137011293-b2aee66a-9fb7-446a-8bea-2d89fb803459.mp4

PR #399, opened 2021-10-19
chore: update metadata on GrantRound UI

thelostone-mc:
- [x] Change "Matching" to "Matching Algorithm"
- [x] If empty, default to "Linear"
- [x] Change "Funds" to "Matching Funds"

image

Fixes https://github.com/dcgtc/dgrants/issues/366


apbendi (commented):
Looks good @thelostone-mc. One nit. It looks like there's an extra space in front of the matching algo name:

Screen Shot 2021-10-19 at 10 28 02 AM

Issue #423, opened 2021-10-21
Rename and refactor grant list filtering logic


phutchins:
- Filtering all grants down to the whitelisted grants should be done in the data store (utilizing the filterGrantsByID method from utils mentioned below)
  - Data store would pull all grants and populate allGrants in local data store
  - Data store would pull the global whitelist array from IPFS, filter **allGrants** based on the whitelist resulting in **whitelistGrants** stored in local data

- Filtering (filterGrantsByID) (based on **whitelistGrants**, or **allGrants**) would be in a function in utils
  - Function arguments should be filterGrantsByID(initialGrantList []grant, filterByList []int)
  - This function should return []grant

- Filtering by other parameters (filterGrants(initialGrantList []grant, filterByList []grant)) would be in a function in utils
  - To filter by one or more parameters you would only set those params on each of the grants in the filterByList
  - We would need to determine a way to filter with ranges, wild cards, and comparison operators
  - This function should return []grant

- Filtering of grants for a GrantRounds grant list should be done in the view passing the resulting list into the component

- Naming Changes and Additions
  - **allGrants** - Locally stored list of all grants on the GrantRegistry
  - **whitelistedGrants** - Locally stored list of grants on the whitelist that exist in allGrants

PR #427, opened 2021-10-21
arrow navigation whitelist fix

### LINKED ISSUE #409: Forward/Back Arrows on Grant Detail Page Should Browse Through Whitelist IDs
Rather than simply incrementing the ID, clicking the arrow should advance to the next or previous ID in the whitelist.

mk1r:
closes #409 

ready for review, introduces some tech-debt but will get fixed when #423 is resolved


phutchins (approved):
Tested and works as expected.

Issue #451, opened 2021-10-25
Pages should only load content for the grants that they need to display


phutchins:
Currently loading is slow which is in part caused by the fact that we load all grants (all grants on chain, not just the whitelisted grants) on the home and grant view pages. Likely on the GrantRound GrantList page as well.

Issue #429, opened 2021-10-21
List of all grants is briefly displayed before whitelist filter applied


phutchins:
When loading a page with a grant list, all grants are briefly displayed before the whitelist is applied.

- The loading indicator should be shown until the filtered list of grants can be displayed

Issue #407, opened 2021-10-20
Cache whitelist and move it to the data store


apbendi:
If the whitelist doesn't load, we currently show all grants. Instead, cache the whitelist, degrade to the cached version if not available, update when user changes pages. Move whitelist to datastore for global access.

PR #455, opened 2021-10-27
chore: move whitelist to data.ts

thelostone-mc:
## Description

- reduces calls to whitelist API
- ensure relevant grants are shown on the grants landing page
- moves the filter function to getAllGrants

Testing 

- ✅ Homepage shows filtered grants
- ✅Grants list shows filtered grants
- ✅ Next / Prev also shows filtered grants
- ✅ cart shows approved grants
- ✅ ~Access unapproved grants via URL  http://localhost:3000/#/dgrants/1?unverified=true~

closes https://github.com/dcgtc/dgrants/issues/451
closes https://github.com/dcgtc/dgrants/issues/407
closes https://github.com/dcgtc/dgrants/issues/429

~Q: Another way to go about this is to have a separate function in data.ts which just fetches the whitelist and have the frontend do the filtering
Q: As opposed to having data return grants. Would it make sense to return whitelistGrants and allGrants ?~
~What's missing ? Since whitelist is done at grant data layer you cannot access the grant URL directly.
Need to sync up with team if it makes sense a new route URL which would allow viewing unapproved grants by URL~

### DIFF
--- a/app/src/store/data.ts
+++ b/app/src/store/data.ts
@@ -33,6 +33,7 @@ import {
   GRANT_ROUND_MANAGER_ADDRESS,
   SUPPORTED_TOKENS_MAPPING,
   DEFAULT_PROVIDER,
+  DGRANTS_CHAIN_ID,
 } from 'src/utils/chains';
 import { DefaultStorage, getStorageKey, setStorageKey } from 'src/utils/data/utils';
 import { GRANT_ROUND_ABI, ERC20_ABI } from 'src/utils/constants';
@@ -46,6 +47,7 @@ const lastBlockNumber = ref<number>(0);
 
 const grants = ref<Grant[]>();
 const grantContributions = ref<Contribution[]>();
+const approvedGrants = ref<Grant[]>();
 const grantMetadata = ref<Record<string, GrantMetadataResolution>>({});
 
 const grantRounds = ref<GrantRound[]>();
@@ -111,6 +113,9 @@ export default function useDataStore() {
 
       return grants;
     }, {});
+
+
+    const approvedGrantsList = await getApprovedGrants(grantsData.grants);
 
     // Pull state from each GrantRound
     const roundAddresses = grantRoundData.roundAddresses || [];
@@ -190,6 +195,7 @@ export default function useDataStore() {
 
     // Save off data
     grants.value = grantsList as Grant[];
+    approvedGrants.value = approvedGrantsList as Grant[];
     grantContributions.value = contributions as Contribution[];
     grantRounds.value = grantRoundsList as GrantRound[];
     grantRoundsDonationToken.value = SUPPORTED_TOKENS_MAPPING[grantRoundDonationTokenAddress] as TokenInfo;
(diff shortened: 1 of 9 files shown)


app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -324,10 +322,13 @@ function useGrantDetail() {
...
   const route = useRoute();
 
+  // --- show unverified grants if unverified = true
+  const grants = route.query.unverified ? allGrants : approvedGrants;
+
   // --- expose grant data ---
   const grant = computed(() => {
     if (!grants.value) return null; // array type unsupported
-    return grants.value[grantId.value];
+    return grants.value.filter((grant) => grant.id.toString() == route.params.id)[0];
```
mds1, line 331: Why was this changed? The original approach was O(1) and now this is O(n), so it might be slow if there's a lot of grants (though realistically the speed change will probably be unnoticeable)

app/src/components/BaseHeader.vue
```diff
@@ -28,15 +28,15 @@
     <span>
       <!-- optional navigation to jump to next item in a list -->
       <router-link
-        v-if="nextPath"
+        v-if="nextPath && !isUnverified"
```
mds1, line 31: If I build for polygon and directly navigate to `http://localhost:3000/#/dgrants/19` (an unverified grant) I see a page that just says "No grant selected". If I navigate to `http://localhost:3000/#/dgrants/19?unverified=true`, then the page loads. 

It seems a bit odd that you'd need to add that flag just to view a grant, so just want to confirm that's the desired behavior

mk1r: I think this is not desired - tested this is well with grant 9. On main, if we navigate to a non-whitelisted grant like 9 or 19 for example, the grant shows is still viewable which is what we want.

app/src/views/GrantRegistryGrantDetail.vue
mk1r: just looked into it, and I believe it's because `grantId.value` is the index value for non-whitelisted grants, but since we're fetching the specific grant from whitelisted grants, there is an index mismatch and it needs to pull the id instead so it's filtering for that

there probably is a more efficient way to handle this though, and this might be related to the other comment about making non-whitelisted grants viewable / unviewable

mds1: Ah, that makes total sense, thanks!

app/src/components/BaseHeader.vue
thelostone-mc: @mk1r  + @mds1 pushed in a fix. could you recheck this now
Simplified it a bit

mk1r (approved):
looking good! nice changes

mk1r:
oh but actually, don't we want to show non-whitelisted grants as well? @phutchins can also provide some clarity

thought that was changed but double checked and perhaps not?

app/src/store/data.ts
```diff
@@ -112,6 +114,9 @@ export default function useDataStore() {
       return grants;
     }, {});
 
+
+    const approvedGrantsList = await getApprovedGrants(grantsData.grants);
```
paffur, line 118: https://github.com/dcgtc/dgrants.wiki.git

paffur (approved):

mds1 (approved):

PR #460, opened 2021-10-29
Contribution Count Shows Different Numbers  - fixed #443

### LINKED ISSUE #443: Contribution Count Shows Different Numbers
Across different devices and users, the "Contribution Count" on the homepage is showing different numbers.

Ajand:
The root of the bug was the "first" parameter in the GraphQL query which has a default number of 100 and was not handled:
`{
                      grantDonations(where: {lastUpdatedBlockNumber_gte: ${fromBlock}, lastUpdatedBlockNumber_lte: ${blockNumber}}) {
                        grantId
                        tokenIn
                        donationAmount
                        from
                        hash
                        rounds
                        lastUpdatedBlockNumber
                      }
                    }`

so whenever the length of the contributions would go more than 100, the system would act inconsistently.  I created a recursive function that will continue to fetch contributions until there is no more left.

_It only works if the cached data is properly set before "fromBlock" which is not unless there are less than 100 contributions_ **So do the hard refresh and remove the cached data**

### DIFF
--- a/app/src/utils/constants.ts
+++ b/app/src/utils/constants.ts
@@ -41,7 +41,7 @@ export const DefaultForageConfig: LocalForageConfig = {
   // this store should be used for any on-chain/off-chain data but never user data (as we might clear it without notice)
   name: 'dGrants',
   // we can bump this version number to bust the users cache
-  version: 1,
+  version: 3,
 };
 
 // LocalForage keys
--- a/app/src/utils/constants.ts
+++ b/app/src/utils/constants.ts
@@ -41,7 +41,7 @@ export const DefaultForageConfig: LocalForageConfig = {
   // this store should be used for any on-chain/off-chain data but never user data (as we might clear it without notice)
   name: 'dGrants',
   // we can bump this version number to bust the users cache
-  version: 3,
+  version: 2,
 };
 
 // LocalForage keys
--- a/app/src/utils/data/contributions.ts
+++ b/app/src/utils/data/contributions.ts
@@ -56,7 +56,11 @@ export async function getContributions(
 
               const limit = 100;
 
-              const fetchUntilAll = async (SUBGRAPH_URL: string, before: any[] = [], skip = 0): Promise<any[]> => {
+              const fetchUntilAll = async (
+                SUBGRAPH_URL: string,
+                before: ContributionSubgraph[] = [],
+                skip = 0
+              ): Promise<any[]> => {
                 const res = await fetch(SUBGRAPH_URL, {
                   method: 'POST',
                   headers: { 'Content-Type': 'application/json' },
--- a/app/src/utils/data/contributions.ts
+++ b/app/src/utils/data/contributions.ts
@@ -285,6 +285,10 @@ export function grantDonationListener(
   },
   refs: Record<string, Ref>
 ) {
+  let timoutManager = 0;
+  const eachSaveTime = 400;
+  const processingSet = new Set();
+
   const listener = async (
     grantId: BigNumberish,
     tokenIn: string,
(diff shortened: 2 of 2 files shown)


app/src/utils/constants.ts
```diff
@@ -41,7 +41,7 @@ export const DefaultForageConfig: LocalForageConfig = {
   // this store should be used for any on-chain/off-chain data but never user data (as we might clear it without notice)
   name: 'dGrants',
   // we can bump this version number to bust the users cache
-  version: 1,
+  version: 3,
```
mds1, line 44: Why 3 and not 2?

app/src/utils/data/contributions.ts
```diff
@@ -56,7 +56,7 @@ export async function getContributions(
 
               const limit = 100;
 
-              const fetchUntilAll = async (SUBGRAPH_URL: string, before = [], skip = 0): Promise<any[]> => {
+              const fetchUntilAll = async (SUBGRAPH_URL: string, before: any[] = [], skip = 0): Promise<any[]> => {
```
mds1, line 59: Ideally we should have stronger typing than `any`, but we can save that for the PR that handles https://github.com/dcgtc/dgrants/issues/463

app/src/utils/constants.ts
Ajand: I wasn't sure what should I put, thought maybe some other PR is going to use 2.

mds1:
> I contribute to 6 grants on my own wallet and for other users everything was fine, but the contributor user only fetched 2 contributions. This PR is still an improvement and will fix the problem for most users, but for fixing #443 completely I need to find the root of this problem as well. Maybe we should create a new issue for it.

@Ajand code-wise this looks good, but I'm not sure what exactly you mean here? Which users is this working for and which is it only fetching 2 contributions for?

Ajand:
> > I contribute to 6 grants on my own wallet and for other users everything was fine, but the contributor user only fetched 2 contributions. This PR is still an improvement and will fix the problem for most users, but for fixing #443 completely I need to find the root of this problem as well. Maybe we should create a new issue for it.
> 
> @Ajand code-wise this looks good, but I'm not sure what exactly you mean here? Which users is this working for and which is it only fetching 2 contributions for?

app/src/utils/constants.ts
mds1: There are no other open PRs changing this, so we should use 2 🙂
```suggestion
  version: 2,
```

Ajand:
> > I contribute to 6 grants on my own wallet and for other users everything was fine, but the contributor user only fetched 2 contributions. This PR is still an improvement and will fix the problem for most users, but for fixing #443 completely I need to find the root of this problem as well. Maybe we should create a new issue for it.
> 
> @Ajand code-wise this looks good, but I'm not sure what exactly you mean here? Which users is this working for and which is it only fetching 2 contributions for?

Let's assume there are 3 users. user 1  will contribute to 10 grants. for user2 and user3, everything is fine, but for user1 we gonna get inconsistent results.

mk1r:
agree with @mds1 comments, so once resolved and approved, I will approve this
good work!

Ajand:
Wrong commit message. There is nothing here for #469. I meant #460.

app/src/utils/data/contributions.ts
Ajand: Now we use type ContributionSubgraph instead of any.

app/src/utils/constants.ts
Ajand: fixed

Ajand:
There was two problems with #470:
1. The grant donation function would start a race with itself and would override the data it wanted to save so if in a block there would be more than one contribution some of them would not be saved even though they were fetched. I solved that issue by creating an async function call `wait(timeout)` and using a closure called `timoutManager` to put a timeout so the next call would wait so the last call could save its data.

2.  For contributor users for some reason the listener processor would be called 2 times simultaneously so even the in-memory closure wouldn't be able to stop the race. So I added a random timeout in addition to a normal timeout and used a set to prevent processing the same event multiple times.


This approach works fine for now but if we're going to scale ( for example if we have 50 contributions in each block ) it doesn't work. We need to create a queue while processing the events and saving all contributions at once after processing.

Also, it would be nice to find the root of double processing for contributor users so could stop it. Since I'm using a random number there is still a slim chance for some contributors to have some issues.

phutchins:
Whats left here @Ajand ? Is this read for final review? @mk1r thoughts?

Ajand:
Hi @phutchins . 
I think it's finished. It's not the ultimate best solution but it's a pretty good solution for now and I believe it works unless we get to Facebook level scale. I guess we just need @gdixon approval as @mds1 requested.

mk1r:
@phutchins yes looks good to me, are all of @mds1 concerns resolved? @Ajand

mk1r (approved):
good job!

Issue #467, opened 2021-10-31
Bug: App is unable to read/find GrantRounds metadata


gdixon:
**Describe the bug**

App is unable to read the `GrantRounds` metadata which is blocking the site from loading.

**To Reproduce**

1. visit: https://grants.gtcdao.net/#/
2. clear application data/indexedDB data
2. refresh

**Expected behavior**

The site should load correctly.

**Screenshots**

<img width="1919" alt="Screenshot 2021-10-31 at 04 28 09" src="https://user-images.githubusercontent.com/5897836/139567637-ed27eb4e-9f0a-4df1-a54c-46462623aaa0.png">

**Additional context**

Version: 0.0.31-1cce972.0

PR #468, opened 2021-10-31
fix: moves metaPtr update to getGrantRound

gdixon:
This PR will move the metaPtr str replacement into the utility function which syncs the data with storage (this corrects the ptr currently stored in indexedDB without the need for cache busting).

--

Fixes: #467


thelostone-mc (approved):

thelostone-mc (approved):

mk1r (approved):
nice find (and fix)

PR #479, opened 2021-11-03
Breadcrumb now works as expected in #464

### LINKED ISSUE #464: Breadcrumb Navigation Bugs
Breadcrumb navigation is messed up

* View a round and the breadcrumbs are DGRANTS > ROUNDS > #0x123. This should either (1) have the round name instead of address, or (2) remove the pound sign in front of the address
* From that page, click "See Grants" and now breadcrumbs are DGRANTS > DGRANTS PROOF-OF-CONCEPT >, but it should be DGRANTS > ROUNDS > DGRANTS PROOF-OF-CONCEPT >
* Then click on the "Octavio leaving" grant, and breadcrumbs are DGRANTS > REGISTRY > #5, but (1) ideally the grant number is replaced with the name, and (2) the breadcrumbs changed entirely and no longer show round information
* Breadcrumbs will also get long once fixed, so need to make sure they wrap nicely or truncate nicely on mobile

Ajand:
fixed #464 

- [x] View a round and the breadcrumbs are DGRANTS > ROUNDS > #0x123. This should either (1) have the round name instead of address, or (2) remove the pound sign in front of the address
- [x] From that page, click "See Grants" and now breadcrumbs are DGRANTS > DGRANTS PROOF-OF-CONCEPT >, but it should be DGRANTS > ROUNDS > DGRANTS PROOF-OF-CONCEPT >
- [x] Then click on the "Octavio leaving" grant, and breadcrumbs are DGRANTS > REGISTRY > Re-import Original Prototype Contracts #5, but (1) ideally the grant number is replaced with the name, and (2) the breadcrumbs changed entirely and no longer show round information

=====
For the last one, I think it's pretty nice already but @scco  should decide:

Screen Shot 1400-08-12 at 16 43 11

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dgrants/app",
-  "version": "0.0.40-303466f.0",
+  "version": "0.0.41-de1b667.0",
   "private": true,
   "scripts": {
     "clean": "rimraf dist",
@@ -14,9 +14,9 @@
     "prettier": "prettier --write ."
   },
   "dependencies": {
-    "@dgrants/contracts": "^0.0.26-303466f.0",
-    "@dgrants/dcurve": "^0.0.26-303466f.0",
-    "@dgrants/types": "^0.0.15-2183c99.0",
+    "@dgrants/contracts": "^0.0.27-de1b667.0",
+    "@dgrants/dcurve": "^0.0.27-de1b667.0",
+    "@dgrants/types": "^0.0.16-de1b667.0",
     "@fusion-icons/vue": "^0.0.0",
     "@headlessui/vue": "^1.2.0",
     "@tailwindcss/aspect-ratio": "^0.2.1",
--- a/app/src/components/GrantCard.vue
+++ b/app/src/components/GrantCard.vue
@@ -1,7 +1,16 @@
 <!-- GrantCard -->
 
 <template>
-  <figure class="group cursor-pointer" @click="pushRoute({ name: 'dgrants-id', params: { id: id.toString() } })">
+  <figure
+    class="group cursor-pointer"
+    @click="
+      pushRoute({
+        name: 'dgrants-id',
+        params: { id: id.toString() },
+        query: roundAddress && roundName ? { roundAddress: roundAddress, roundName: roundName } : {},
+      })
+    "
+  >
     <!--img-->
     <div class="relative">
       <!--img-->
@@ -70,6 +79,8 @@ export default defineComponent({
     name: { type: String, required: true },
     imgurl: { type: String, required: true },
     ownerAddress: { type: String, required: true },
+    roundAddress: { type: String, default: '' },
+    roundName: { type: String, default: '' },
   },
   components: { CartIcon },
   setup(props) {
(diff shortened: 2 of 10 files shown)


mk1r:
you don't need to fork the repo, feel free to create a branch in this repo for your PRs :)

first two are looking good - will review the last two once the last one is complete, actually I don't see the "Octavio leaving" grant

Ajand:
> you don't need to fork the repo, feel free to create a branch in this repo for your PRs :)
> 
> first two are looking good - will review the last two once the last one is complete, actually I don't see the "Octavio leaving" grant

Oh I was following this for contribution: [CONTRIBUTING.md](https://github.com/dcgtc/dgrants/blob/main/CONTRIBUTING.md)

For the Octavio leaving are you in the rinkeby chain? By the way, it doesn't make any difference just check any grants at it should work. I'm not sure that @scco wants to change the last one, since it's already looking pretty good imo.

mk1r:
I think the forking is for non-team members (i.e open source contributions) but still good to follow most of it - forking makes it more difficult to review so probably better to create a branch in this repo

Ah that makes sense. I'm on Polygon, looking good. Maybe just create a new issue for the last one (and remove it from here) and then I'm good to approve 👍

phutchins:
I haven't run this branch myself as it's in a remote but it all looks good from a glance. I think the last item on this page looks fine and we can add an additional issue if it is decided that the mobile verison of breadcrumb needs work.

Lets get it reviewed and merged.

mk1r (approved):
looking good, nice work

thelostone-mc:
codewise LGTM!  @Ajand  + @apbendi 

I'm not super clear on the 3rd requirement here
image

It still shows
- Grant Number
- Grant Round Name

Is that what is expected ?

apbendi:
> It still shows

> Grant Number
> Grant Round Name
> Is that what is expected ?

I actually don't think this is right. Octavio's grant was NOT in that round, so it definitely shouldn't show. Regardless, Grant's can be in multiple rounds, so I don't think rounds should show in breadcrumbs should show at all.

mds1:
> Regardless, Grant's can be in multiple rounds, so I don't think rounds should show in breadcrumbs should show at all.

@apbendi My thinking was that they would show the round, if that's how you navigated to that grant detail. My understanding of breadcrumbs is that they show the path you took to the current page, so:
- If I accessed grant 5 by viewing all grants from the dGrants POC round, then it would have the round name in the breadcrumbs
- If I accessed grant 5 by viewing all grants in the registry, then it would not have the round name in the breadcrumbs

thelostone-mc:
Hmm the only confusion here is @mds1 the breadcrumb right now says 
**Dgrants** > **Rounds** > **Round-Name** > **GrantId** 

Grant ID is the registry ID which is why i found the path a lil confusing

Ajand:
@thelostone-mc @mds1 @apbendi I made it in a way that it will show it like this (if we go directly to the grant page it shows a different breadcrumb) if we go to a grant page from a specific round. Like it's more about the navigation path, not the destination. 

@thelostone-mc Do you have some alternative approach in mind instead of using registry ID?

apbendi:
> I made it in a way that it will show it like this (if we go directly to the grant page it shows a different breadcrumb) if we go to a grant page from a specific round. Like it's more about the navigation path, not the destination.

Got it, this seems totally reasonable and makes sense. If this is how it's implemented, I'm fine with it. It still doesn't explain to me why Octavio's test grant, which (unless I'm mistaken) was not in the round is shown in the screenshot with the round in the breadcrumb. So just want to make sure we don't have a bug there.

Ajand:
@apbendi It exists on the round page.
Screen Shot 1400-08-19 at 19 47 00

PR #494, opened 2021-11-10
feat: my grants

thelostone-mc:
#### Description

- [x] route from landing to my grants
- [x] shows lift of grants owned by users
- [x] shows if grants is approved / pending state
- [x] clicking on title / picture opens grants
- [x] clicking on edit button opens grant in edit view  
- [x] @scco roping you in for UI / Css fixes 

Check comment below for pending actions items/questions

Untitled


thelostone-mc:
Couple of call outs here:

- synced up with @gdixon and looks like we might need to make changes to subgraph to ensure the data we get always has the latest owner. Although looking at the code , i think we fetch the latest info https://github.com/dcgtc/dgrants/blob/main/app/src/utils/data/grants.ts#L51-L58
@gdixon wouldn't this always return the latest owner ?


**Missing Data**

- Submitted  // creation date (not sure where we could fetch this from)
- Last Updated (we do have the last block number from subgraph)
- @scco is Matching Fund the matching estimate ? How do we handle scenarios where we have diff round ?


**Issues I need to fix**

- [ ] see why https://github.com/dcgtc/dgrants/blob/my-grants/app/src/components/GrantOwnerCard.vue#L106 doesn't populate
- [ ]

app/src/components/GrantOwnerCard.vue
```diff
@@ -0,0 +1,119 @@
...
+  </section>
+</template>
+
+<script lang="ts">
+import { computed, defineComponent, ref } from 'vue';
+// --- Methods and Data ---
+import { filterContributionsByGrantId } from 'src/utils/data/contributions';
+
+// --- Utils/helper ---
+import { formatNumber, pushRoute } from 'src/utils/utils';
```
mds1, line 71: Nit, but probably can put these under the same comment header

```suggestion
// --- Methods and data ---
import { filterContributionsByGrantId } from 'src/utils/data/contributions';
import { formatNumber, pushRoute } from 'src/utils/utils';
```

app/src/components/GrantOwnerCard.vue
```diff
@@ -0,0 +1,119 @@
...
+
+// --- Components/icons ---
+import useDataStore from 'src/store/data';
```
mds1, line 74: This would also be under the "methods and data" comment

app/src/components/GrantOwnerCard.vue
```diff
@@ -0,0 +1,119 @@
...
+      contributions.reduce((total, contribution) => contribution?.amount + total, 0),
+      2
+    )} ${grantRounds.value && grantRounds.value[0] && grantRounds.value[0].donationToken.symbol}`;
+
+    return { raised: raised, contributionCount: contributionCount };
+  }
+
+  const grantInfo = computed(() => getGrantInfo(grantId.value));
+
+  return { status: status, grantInfo };
```
mds1, line 103: Nit
```suggestion
  return { status, grantInfo };
```

app/src/router/index.ts
```diff
@@ -16,6 +16,7 @@ const routes: Array<RouteRecordRaw> = [
   { path: '/dgrants', name: 'dgrants', component: () => import('../views/GrantRegistryList.vue') },
   { path: '/review', name: 'review', component: () => import('../views/GrantReview.vue') },
   { path: '/dgrants/new', name: 'dgrants-new', component: () => import('../views/GrantRegistryNewGrant.vue') },
+  { path: '/dgrants/mygrants', name: 'dgrants-my-grants', component: () => import('../views/MyGrants.vue') },
```
mds1, line 19: Nit on path slug, this would be my personal preference 
```suggestion
  { path: '/dgrants/my-grants', name: 'dgrants-my-grants', component: () => import('../views/MyGrants.vue') },
```

app/src/store/data.ts
```diff
@@ -295,18 +298,24 @@ export default function useDataStore() {
    * @notice helper function to filter approved grants
    *
    * @param grants Grant[]
+   * @param approvedList []
```
mds1, line 301: The `approvedList` entry should be deleted and `grants` is missing the description string

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -647,6 +647,13 @@ function useGrantDetail() {
     form.value.twitter = cleanTwitterUrl(grantMetadata.value?.properties?.twitterURI) || '';
   }
 
+  /**
+   * @notice auto open if URL has query edits
+   */
+  if (route.query && route.query.edit && isOwner) {
+    enableEdit();
+  }
+
```
mds1, line 656: Should this be within an `onMounted` hook? It looks like it just lives directly in the composition function, so I wonder if there could be a race condition where this doesn't always execute properly if it's ran before mounting finishes? I'm not sure how it works when directly within the composition function like this

mds1 (commented):
Looking good! One other comment—if I try creating a new grant on localhost with `VITE_DGRANTS_CHAIN_ID=31337`, I get a 500 from fleek and can't create the grant

app/src/views/GrantRegistryGrantDetail.vue
thelostone-mc: moved it to onMount

mk1r (approved):
tested it on Rinkeby as opposed to hardhat network, and looking good!

app/src/components/GrantOwnerCard.vue
```diff
@@ -0,0 +1,116 @@
...
+  return { status, grantInfo };
+}
+
+export default defineComponent({
+  name: 'GrantOwnerCard',
+  components: { EditIcon },
+  props: {
+    id: { type: Number, required: true },
+    name: { type: String, required: true },
+    imgurl: { type: String, required: true },
```
mds1, line 109: This should take `logoPtr` as type `MetaPtr` as a prop, and resolve it to an image URL. We always want to pass around the full pointer object and resolve it as late as possible. I think this PR needs to be rebased against `main` because you'll see we removed all `imgurl` props from components and now pass in the logo prop like this: https://github.com/dcgtc/dgrants/blob/a752c90bc1dbc44a901d22e3efe05557f0726361/app/src/components/GrantCard.vue#L84

And resolve it as late as possible like this: https://github.com/dcgtc/dgrants/blob/a752c90bc1dbc44a901d22e3efe05557f0726361/app/src/components/GrantCard.vue#L18-L21

thelostone-mc: Hmm the parent was actually taking care of that but i see what you're saying. this is cleaner

apbendi (commented):
Looks good @thelostone-mc! Here's one issue I noticed: an off by one error for the month when displaying the created/updated dates.

Screen Shot 2021-11-23 at 9 32 51 AM

Also wondering how hard it would be to localize the date format, but that's not critical.

thelostone-mc:
Hmm this is weird !
@gdixon any idea if the subgraph value is wonky ? cause Ive just do a .getMonth here

app/src/utils/utils.ts
```diff
@@ -62,6 +62,12 @@ export function daysAgo(val = 0) {
   return formatter.format(Math.round(deltaDays), 'days');
 }
 
+// Convert unix ts to DD-MM-YYYY
+export function formatDate(dateStr: BigNumber): string {
+  const date = new Date(BigNumber.from(dateStr).toNumber() * 1000);
+  return `${date.getDate()}-${date.getMonth()}-${date.getFullYear()}`;
```
thelostone-mc, line 68: @apbendi + @gdixon this is what i was talking about

mds1: `getMonth()` is zero indexed and returns a value between 0 through 11

mds1: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth

thelostone-mc: UGH FML

thelostone-mc:
```suggestion
  return `${date.getDate()}-${date.getMonth()+1}-${date.getFullYear()}`;
```

mds1: 😜😂

gdixon: Months range from 0-11 in javascript (a quirk that was copied from java)! So we just need to +1 on the month (with an exception for when adding 1 will incr the year too)!

PR #496, opened 2021-11-10
update metaPtr type

### LINKED ISSUE #481: Change MetaPtrs to Include Discreet Storage Type
Currently, metaPtrs on Grant Rounds and Grants are stored as full URIs, which are then fetched by the client to obtain the data after the pointer itself is read from the chain. While we currently use IPFS for this data, via Fleek, the URI could in theory be any http endpoint.

A number of issues with this have become clear over time.

1. Trust/CORS/performance issues that come from fetching data from arbitrary endpoints
2. Flexibility to change how data from certain protocols is fetched. Importantly, when we attempted to switch endpoint for the IPFS proxy, we quickly realized all existing Grant metadata would have to be updated to do this correctly.

One naive solution would be to change the spec to instead store only the IPFS CID, and allow the client to fetch this data through any source it preferred. The obvious downside here is the lock in to IPFS— what if we want to support other protocols (Filecoin, Arweave, IC, etc...) in the future for metadata storage?

In order to avoid these issues, while also being flexible for the future, we'll update the definition of a metaPtr. We can store two values as a Tuple: `(uint kind, bytes identifier)`.

`kind` will be a de facto Enumerated type, defined externally as part of the protocol spec, defining the kind of decentralized storage that is being used. For now, the only valid `kind` value will be `1` for IPFS. (0 should be reserved as a none/invalid metaPtr to avoid mistaking an unset metaPtr as some protocol).

`identifier` will be whatever data is needed for the protocol defined by this code, the definition of which will also be considered part of the dGrants spec. For the IPFS case, it will be defined as the CID of the metadata to be fetched.

This change will require changes to the contracts and to the frontend. It needs to be implemented for Grants, Grant Rounds, AND the sub-metaPtrs to image data stored in the metadata itself. In the latter case, the same Tuple structure should be used, just as an element in the metadata JSON:

```json
{
    "kind": 1,
    "identifier": "QmNzQmytmt5vU7XsKyLcYgkD7sdkXzgR2rSiTJkDhdcRxP",
}
```

When data is fetched in the frontend, it should simply:

1. Validate the `kind` is 1, and fail if not
2. Fetch the data using the CID via our IPFS gateway of choice

Additional note:

Open to alternative name suggestions for both `kind` and `identifier`

mds1:
Closes #481 

This description first contains details about the fact that this PR is a breaking change, then details on the changes are given. Lastly, there are some open questions we should answer before merging this PR

## WARNING

**This PR is a MAJOR breaking change because it changes the keys used to cache data locally**. As a result:
1. You will need to clear all site data
2. Once this PR is merged, you will only be able to dev/test with `VITE_DGRANTS_CHAIN_ID=31337` which runs the app against Hardhat. This is required because no contracts with this new `MetaPtr` format are deployed to Rinkeby or Polygon
3. To test sending transactions, you'll need to set MetaMask to use one of the default Hardhat accounts (reminder that instructions on how to do this can be found [here](https://github.com/dcgtc/dgrants/tree/main/app#setup))

## Details 

The contracts now use a struct to define the metadata pointer. This is defined in `contracts/contracts/interfaces/IMetadataPointer.sol` as follows:

```solidity
struct MetaPtr {
  uint256 protocol; // mapping from integer to protocol, currently only a value of 1 representing IPFS is supported
  string pointer; // pointer to fetch metadata for the specified protocol
}
```

To account for this on the frontend, we had to make some changes:
- A new method called `metadataId()` was added which takes in the above struct and generates a string from the data. That string can be used in place of the original `metaPtr` string type that was used to index into objects
- The inverse of this is a method called `decodeMetadataId`, which takes in the string and returns the original object

## Questions / TODO before merging

- [x] 1. @phutchins Can I create a `dcgtc/protocol-ids` repo to track the mapping from `protocol` integers to protocol names? The repo will be a standalone repo with a single README.md file that looks very similar to [this file](https://github.com/satoshilabs/slips/blob/master/slip-0044.md), i.e. it will be a simple table and anyone can submit a PR to add a new mapping from integer to protocol
- [x] 2. Given that this PR is a breaking change, do we want to use `bump:minor` instead of the usual `yarn bump:patch` before merging? Normally this would be `yarn bump:major` but I don't think we want to have a v1.0.0 just yet
- [x] 3. We'll want to bump the `version` specified in `app/src/utils/constants.ts`. I haven't done this yet because I know @gdixon is also bumping it in #493 so we'll need to both adjust the version accordingly depending on which gets merged first

### DIFF
--- a/app/src/components/GrantReviewList.vue
+++ b/app/src/components/GrantReviewList.vue
@@ -5,9 +5,9 @@
       <li v-for="grant in sortedGrants" :key="grant.id.toString()">
         <GrantCard
           :id="grant.id"
-          :name="(grantMetadata && grantMetadata[grant.metaPtr]?.name) || '...'"
+          :name="(grantMetadata && grantMetadata[metadataId(grant.metaPtr)]?.name) || '...'"
           :ownerAddress="grant.owner"
-          :imgurl="(grantMetadata && grantMetadata[grant.metaPtr]?.logoURI) || '/placeholder_grant.svg'"
+          :imgurl="(grantMetadata && grantMetadata[metadataId(grant.metaPtr)]?.logoURI) || '/placeholder_grant.svg'"
         />
       </li>
     </ul>
@@ -16,7 +16,7 @@
 
 <script lang="ts">
 import { watch, computed, defineComponent, onMounted, onUnmounted, PropType, ComputedRef, ref } from 'vue';
-// --- App Imports ---
+// --- Components ---
 import GrantCard from 'src/components/GrantCard.vue';
 // --- Store ---
 import useCartStore from 'src/store/cart';
@@ -24,6 +24,7 @@ import useCartStore from 'src/store/cart';
 import { FilterNavItem, Grant, GrantMetadataResolution } from '@dgrants/types';
 // --- Utils ---
 import { DGRANTS_CHAIN_ID } from 'src/utils/chains';
+import { metadataId } from 'src/utils/utils';
 
 type SortingMode = 'newest' | 'oldest' | 'shuffle';
 
@@ -122,9 +123,10 @@ export default defineComponent({
 
     return {
       ...useSortedGrants(grantList),
-      isInCart,
       addToCart,
+      isInCart,
       removeFromCart,
+      metadataId,
     };
   },
 });
--- a/app/src/utils/utils.ts
+++ b/app/src/utils/utils.ts
@@ -7,7 +7,7 @@ import { Event } from 'ethers';
 import { BigNumber, BigNumberish, commify, Contract, ContractTransaction, isAddress, Logger } from 'src/utils/ethers';
 import { BatchFilterQuery, EtherscanGroup } from 'src/types';
 import useWalletStore from 'src/store/wallet';
-import { Grant, GrantRound } from '@dgrants/types';
+import { Grant, GrantRound, MetaPtr } from '@dgrants/types';
 import { formatUnits } from 'src/utils/ethers';
 import { ETH_ADDRESS } from 'src/utils/constants';
 import { ETHERSCAN_BASE_URL, FILTER_BLOCK_LIMIT, SUPPORTED_TOKENS_MAPPING, WETH_ADDRESS } from 'src/utils/chains';
@@ -69,6 +69,12 @@ export function urlFromTwitterHandle(handle: string) {
 // Clean twitter url and return the handle
 export function cleanTwitterUrl(url: string | undefined) {
   return url ? url.replace('https://twitter.com/', '') : undefined;
+}
+
+// Given a metadata pointer object, return a stringified version that can be used as an object key
+export function metadataId(metaPtr: MetaPtr) {
+  const keys = Object.keys(metaPtr).sort(); // ensure deterministic ordering of keys
+  return keys.map((key) => metaPtr[<keyof typeof metaPtr>key].toString()).join('-'); // stringify and join all values
 }
 
 // --- Validation ---
--- a/app/src/utils/utils.ts
+++ b/app/src/utils/utils.ts
@@ -72,9 +72,14 @@ export function cleanTwitterUrl(url: string | undefined) {
 }
 
 // Given a metadata pointer object, return a stringified version that can be used as an object key
-export function metadataId(metaPtr: MetaPtr) {
-  const keys = Object.keys(metaPtr).sort(); // ensure deterministic ordering of keys
-  return keys.map((key) => metaPtr[<keyof typeof metaPtr>key].toString()).join('-'); // stringify and join all values
+export function metadataId(metaPtr: MetaPtr): string {
+  return `${BigNumber.from(metaPtr.protocol).toString()}-${metaPtr.pointer}`;
+}
+
+// Given a metadata ID, decode it back to the original metadata pointer object
+export function decodeMetadataId(id: string): MetaPtr {
+  const parts = id.split('-');
+  return { protocol: parts[0], pointer: parts[1] };
 }
 
 // --- Validation ---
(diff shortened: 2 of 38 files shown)


gdixon (approved):
LGTM!! Nicely done @mds1! 🚀

app/src/utils/utils.ts
```diff
@@ -288,10 +299,13 @@ export const decodeMulticallReturnData = (
...
       ...call.fns.map((fn) => {
-        return call.contract.interface.decodeFunctionResult(
+        const data = call.contract.interface.decodeFunctionResult(
           typeof fn === 'string' ? fn : fn.fn,
           returnData[flatPtr++].returnData
-        )[0];
+        );
+        // If the call returned one element, just get rid of the array wrapper and return the data directly.
+        // If the call returned multiple elements (i.e. metaPtr returns an object), we want to preserve that structure
+        return data.length === 1 ? data[0] : data;
```
thelostone-mc, line 308: Are we sure this a good idea ? while it make sense , now we have a function which returns an array or an object

thelostone-mc (commented):
just a few questions/nitpicks.

app/src/components/GrantReviewList.vue
```diff
@@ -5,9 +5,9 @@
       <li v-for="grant in sortedGrants" :key="grant.id.toString()">
         <GrantCard
           :id="grant.id"
-          :name="(grantMetadata && grantMetadata[grant.metaPtr]?.name) || '...'"
+          :name="(grantMetadata && grantMetadata[metadataId(grant.metaPtr)]?.name) || '...'"
           :ownerAddress="grant.owner"
-          :imgurl="(grantMetadata && grantMetadata[grant.metaPtr]?.logoURI) || '/placeholder_grant.svg'"
+          :imgurl="(grantMetadata && grantMetadata[metadataId(grant.metaPtr)]?.logoURI) || '/placeholder_grant.svg'"
```
thelostone-mc, line 10: would it make sense to wrap grantMetadata[metadataId(grant.metaPtr)] into a variable called `metadata` and return from setup function just so that the templating is a lil easier to read ?

(more for readability)

thelostone-mc (approved):

mds1:
@gdixon I'm good with merging #493 first, but if so let's try to get #493 merged today since we want all contract PRs merged by end of week, and this one is the first of two contract-related PRs (I'll review that PR shortly!)

app/src/utils/utils.ts
```diff
@@ -71,6 +71,17 @@ export function cleanTwitterUrl(url: string | undefined) {
   return url ? url.replace('https://twitter.com/', '') : undefined;
 }
 
+// Given a metadata pointer object, return a stringified version that can be used as an object key
+export function metadataId(metaPtr: MetaPtr): string {
```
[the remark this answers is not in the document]
mds1: I think you mean `getMetadataId` right? I'm indifferent though, so up to everyone else. For context, I tried to keep this method name as short as possible to minimize the diff so this PR is easy to review (if too long, lines >120 characters get wrapped)

app/src/utils/utils.ts
mds1: I have a few justifications for this:
1. If a call returns 1 element, there's no benefit to returning that single element as an array to the caller. They'll need to extract it from the array anyway, so we can do that here to reduce code/work by the caller. This removes a lot of useless code, as shown below
2. If a call returns multiple elements, you have to return all of them to the caller regardless so this makes sense
3. This approach minimizes the code diff / minimizes the risk of breaking things, since the change is contained within this method and the caller does not need to make any changes to how it parses return values

Details on 1:

Current format lets us do this: https://github.com/dcgtc/dgrants/blob/95a3bae36c62e7ac817d3537ab9f54adcb650fb0/app/src/utils/data/grantRounds.ts#L167-L177

If we always return the array, then we'd have to do this, which feels clunky for no benefit

```typescript
[
  _donationTokenAddress,
  _startTime,
  _endTime,
  _metadataAdmin,
  _payoutAdmin,
  _registryAddress,
  metaPtr,
  _hasPaidOut,
  _funds,
] = await callMulticallContract(...)

const donationTokenAddress = donationTokenAddress[0]
const startTime = startTime[0]
const endTime = endTime[0]
const metadataAdmin = metadataAdmin[0]
const payoutAdmin = payoutAdmin[0]
const registryAddress = registryAddress[0]
const hasPaidOut = hasPaidOut[0]
const funds = funds[0]
```

app/src/components/GrantReviewList.vue
mds1: Hm, maybe I'm misunderstanding but IMO it becomes less readable, since instead of a 1 line change, now you need to have a metadata array generated by looping through each grant passed in `sortedGrants`, and keep track of an index in the `v-for` elements to make sure you're accessing the corresponding element in the metadata array. And since grant order changes by sorting order, you need to account for that also and regenerate the `metadata` array each time

contracts/contracts/interfaces/IMetadataPointer.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.7.6;
+
+struct MetaPtr {
+  uint256 protocol; // mapping from integer to protocol
```
apbendi, line 5: nit: word "mapping" in this comment is a little misleading/confusing. Might say something like "integer code which corresponds to a specific storage protocol".

apbendi (approved):
Looks good to me! I think we should get this merged before #493 and rebase the latter to it. Don't feel strongly.

contracts/contracts/interfaces/IMetadataPointer.sol
mds1: Heh, fair enough. I'll standup the repo mentioned in the description first and link to it here + rephrase for clarity

mds1: Done in 1a0b682

PR #497, opened 2021-11-12
data fetching contract changes

### LINKED ISSUE #488: Contract Changes for Easier Payout Script / Data fetching
-  Add parameters to events?
-  timestamps for creations / updates on registry
* Others?

mds1:
Closes #488

- Adds `createdAt` and `lastUpdated` timestamps to grant registry (both in storage and events)
- Adds timestamps to donation logs

### DIFF
--- a/contracts/contracts/GrantRegistry.sol
+++ b/contracts/contracts/GrantRegistry.sol
@@ -19,23 +19,28 @@ contract GrantRegistry {
 
   /// @notice Grant object
   struct Grant {
-    uint96 id; // grant ID, as uint96 to pack into same slot as owner (this implies a max of 2^96-1 = 7.9e28 grants)
+    // Slot 1 (within this struct)
+    // Using a uint96 for the grant ID means a max of 2^96-1 = 7.9e28 grants
+    uint96 id; // grant ID, as
     address owner; // grant owner (has permissions to modify grant information)
+    // Slot 2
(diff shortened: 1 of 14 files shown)


contracts/contracts/GrantRegistry.sol
```diff
@@ -19,23 +19,28 @@ contract GrantRegistry {
 
   /// @notice Grant object
   struct Grant {
-    uint96 id; // grant ID, as uint96 to pack into same slot as owner (this implies a max of 2^96-1 = 7.9e28 grants)
+    // Slot 1 (within this struct)
+    // Using a uint96 for the grant ID means a max of 2^96-1 = 7.9e28 grants
+    uint96 id; // grant ID, as
```
apbendi, line 24: I agree with these integer size changes. I do wonder, though, if this one in particular is worth it. We're not saving a slot, so we might actually be adding gas costs (if this change is considered in isolation).

apbendi (approved):
Simple enough. Looks good to me!

mk1r (approved):
Looking good!

contracts/contracts/GrantRegistry.sol
mds1: Sorry not sure I follow? The change highlighted here is just a comment change, even prior to this PR we packed the ID and owner into the same slot

Issue #501, opened 2021-11-12
Fix MetaPtr for logo pointers stored in metadata


mds1:
This issue is the second half of #481. I forgot to make the change to use the new object format for image pointers in round metadata, even though it's covered in the description of this issue. I'll start working on this once current Fleek API key issues are resolved

PR #506, opened 2021-11-15
`MetaPtr` fixes for `logoURI` (now named `logoPtr`)

mds1:
Closes https://github.com/dcgtc/dgrants/issues/501

Notes
- **_Please QA test thoroughly_**. This PR touches a lot of places and was a big refactor, so almost certainly introduces some bugs, so try anything a user might do and please leave steps to reproduce in this PR
- Known issue so far (I think): When you create a grant (or for all grants?), images are not shown
- `BaseImageUpload.vue` and `GrantRegsitryNewGrant.vue` are heavily coupled with the old `logoURI: string` format, so for simplicity I left those as-is because those pages only let users upload to IPFS anyway
- Please also test the deploy script: @apbendi you should now be able to deploy on this branch so if you can test that out, that would be appreciated!

I plan to review this code again myself before merging because it's a lot of changes and I've been staring at code for too long


app/src/utils/utils.ts
```diff
@@ -82,6 +83,14 @@ export function decodeMetadataId(id: string): MetaPtr {
   return { protocol: parts[0], pointer: parts[1] };
 }
 
+// Given a logoPtr, resolve it to a URI that we can fetch
+export function ptrToURI(logoPtr: MetaPtr | undefined | null) {
+  if (!logoPtr) return null;
+  const { protocol, pointer } = logoPtr;
+  if (protocol === 1) return getMetaPtr({ cid: pointer });
+  throw new Error(`Unsupported protocol ID ${protocol}`);
+}
```
thelostone-mc, line 92: ^ not important but throught on removing ptrToURI and moving this logic to `getMetaPtr`

app/src/utils/data/grants.ts
```diff
@@ -155,7 +155,12 @@ export function grantListener(name: string, refs: Record<string, Ref>) {
         grantId = BigNumber.from(grantId).toNumber();
 
         // update the grants metadata
-        if (_lsGrants[grantId] && _lsGrants[grantId].metaPtr !== metaPtr) {
+        const areMetaPtrsEqual = (x: MetaPtr, y: MetaPtr) => {
+          if (BigNumber.from(x.protocol).toString() !== BigNumber.from(y.protocol).toString()) return false;
+          if (x.protocol !== y.protocol) return false;
```
thelostone-mc, line 160: ^ worth combing the two if statements ?

app/src/utils/data/ipfs.ts
```diff
@@ -21,7 +28,9 @@ export const uploadFile = async (file: File) => {
  * @param obj.description Description of grant
  * @returns CID
  */
-export const uploadGrantMetadata = async ({ name, description, logoURI, properties }: GrantMetadata) => {
+export const uploadGrantMetadata = async ({ name, description, logoPtr, properties }: GrantMetadata) => {
+  assertIPFSPointer(logoPtr);
+  const logoURI = getMetaPtr({ cid: (logoPtr as MetaPtr).pointer });
```
thelostone-mc, line 33: would it make sense having `assertIPFSPointer` within `getMetaPtr` ?

thelostone-mc:
**Testing Status:**

- ✔️  Grant Creation 
- ✖️ Edit Grants [Page doesn't load due to ](https://user-images.githubusercontent.com/5358146/141975406-b5392b21-cda3-4588-9b2d-ed391f784e41.png)
- ✔️ Contribute to Grant
- ✔️ Create Grant Round (via fixtures)
- ✔️ Contribute to Grant Round


**Other**

- Grant Round Image doesn't load 
- Grant Image doesn't load
- Unable to test out edit grant flow

thelostone-mc:
@mds1 so it looks like for image upload like you mentioned, they are direct upload to IPFS so maybe it doesn't make since to make those changes to follow the new structure.
Right now this PR does have code changes which assume that the logo would of format logoPtr which causes the image to not load.

I'd propose that we stick to relying on logoURL for logo. This commit below ensures logo loads on GrantDetail page.
We'd have to do this in a few other places to ensure the logo loads on the UI

https://github.com/dcgtc/dgrants/commit/3d4ad531be80dd6bd7b2d991dcb6c693012669aa

thelostone-mc (approved):
left minor comments  + pushed a commit with fixes for broken pieces.
LGTM. neat work here @mds1

mk1r (approved):
Looking good; however, I do wonder if we can make the change in `BaseImageUpload.vue` too. I can see how the change is more difficult to make in `GrantRegsitryNewGrant.vue` and maybe `GrantRegistryGrantDetail.vue`

mds1:
@mk1r Can you elaborate on what you're thinking? We could make the change, but the benefit is questionable given that BaseImageUpload, grant creation, and grant editing are all tied to IPFS already, and AFAIK we have no plans to generalize them for other file storage protocols

mk1r:
@mds1 on line 77 of `BaseImageUpload.vue` can't we use PropType i.e `type: Object as PropType<MetaPtr>`?

app/src/utils/utils.ts
mds1: `getMetaPtr` is specific to IPFS and lives in `ipfs.ts`, whereas this `ptrToURI` is a wrapper around that to support any supported protocol. For example saw we add Arweave support with a protocol ID of 2, this method would detect that and call `arweave.getMetaPtr` instead of the current `ipfs.getMetaPtr`

app/src/utils/data/grants.ts
mds1: Heh, you could, but I did this because it takes up less lines without compromising readability which I like. But since @mk1r gave you the +1 I'm outvoted and will change this

Actually changing it to just `return BigNumber.from(x.protocol).toString() === BigNumber.from(y.protocol).toString() && x.protocol === y.protocol` seems the simplest

app/src/utils/data/ipfs.ts
mds1: We don't want to put it in `metadataId` cause that's generic to all protocols. We could move it to `getMetaPtr` but then there's a code path in `getMetadata` where it's not called, so it seems ok to leave it as-is IMO, with the guidance being "if we're in `ipfs.ts`, as soon as we take a MetaPtr input type, validate it"

mds1: Oh actually we can't move it to `getMetaPtr` because that only takes a CID, so by the time that method is called we've already split apart the pointer object

app/src/utils/data/grants.ts
mds1: Changed in cddb3fa

mds1:
@mk1r right, you'd also need to rename the prop, then make the corresponding adjustments everywhere `BaseImageUpload` is used. So you also want to change the new grant and edit grant forms to use `MetaPtr` types instead of strings for consistency, etc.

Definitely doable, and feel free to push a commit making the change, but as mentioned above I still think that the  benefit is questionable given that BaseImageUpload, grant creation, and grant editing are all tied to IPFS already, and AFAIK we have no plans to generalize them for other file storage protocols Happy to be convinced otherwise though!

mk1r:
@mds1 sounds good, it was just something I was thinking about for stronger type checking, but if the tradeoffs are questionable no need, we can keep it as is :)

app/src/utils/data/ipfs.ts
thelostone-mc: Ah you're right ! Gotcha Gotcha

app/src/views/Cart.vue
```diff
@@ -66,7 +66,7 @@
                 <figure class="aspect-w-16 aspect-h-9 shadow-light">
                   <img
                     class="w-full h-full object-center object-cover group-hover:opacity-90"
-                    :src="grantMetadata[metadataId(item.metaPtr)]?.logoURI || '/placeholder_grant.svg'"
+                    :src="ptrToURI(grantMetadata[metadataId(item.metaPtr)]?.logoPtr) || '/placeholder_grant.svg'"
                   />
```
gdixon, line 70: Not for now, but all of these images have a very similar structure (and the same list of classes), it might be worth abstracting these into a component.

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -20,9 +20,7 @@
       <!-- grant details row ( image + raised, address, in round, matchin, add to cart button ) -->
       <GrantDetailsRow
         :grant="grant"
-        :logoURI="
-          grantMetadata?.logoURI ? grantMetadata?.logoURI.replace('https://ipfs-dev', 'https://ipfs') : undefined
-        "
+        :logoPtr="grantMetadata?.logoPtr ? grantMetadata.logoPtr : undefined"
```
gdixon, line 23: Isn't this the same as:

```suggestion
        :logoPtr="grantMetadata?.logoPtr"
```

gdixon (approved):
Awesome work Matt!! 🚀

app/src/views/Cart.vue
mds1: Added this to #265, good call

app/src/views/GrantRegistryGrantDetail.vue
mds1: Yep it is, just applied your suggestion. This is an artifact of my first refactor "design" where the code was:

```
:logoURI="grantMetadata?.logoPtr ? ptrToURI(grantMetadata.logoPtr) : undefined"
```

But then I changed the prop to `logoPtr` and forgot to change the conditional. So good catch!

PR #513, opened 2021-11-17
fix: drops subgraph urls so that we don't fetch data from the wrong contracts

gdixon:
This PR drops the subgraph urls for Polygon and Rinkeby which will disable those getters. 

We need to get make sure that our subgraph is updated to reflect recent changes and then we can restore these URLs.


mds1 (approved):

apbendi (approved):
lgtm

PR #515, opened 2021-11-17
update ipfs retrieval endpoints

### LINKED ISSUE #498: Update IPFS Fetching Endpoint
The endpoint used for **fetching** from IPFS should be updated to `https://scopelift.b-cdn.net/ipfs/<cid>/`. The trailing slash after the CID is *required* or the endpoint will 500.

This issue is a spinout from #486

mk1r:
resolves #498

### DIFF
--- a/app/src/utils/data/ipfs.ts
+++ b/app/src/utils/data/ipfs.ts
@@ -6,7 +6,7 @@ import { BigNumber } from 'src/utils/ethers';
 import { LocalForageData } from 'src/types';
 import { Ref } from 'vue';
 
-const retrievalEndpoint = 'https://ipfs.fleek.co/ipfs';
+const retrievalEndpoint = 'https://scopelift.b-cdn.net/ipfs/<cid>/';
 
 function assertIPFSPointer(logoPtr: MetaPtr | undefined) {
   if (!logoPtr) throw new Error('assertIPFSPointer: logoPtr is undefined');
--- a/app/src/utils/data/ipfs.ts
+++ b/app/src/utils/data/ipfs.ts
@@ -6,7 +6,7 @@ import { BigNumber } from 'src/utils/ethers';
 import { LocalForageData } from 'src/types';
 import { Ref } from 'vue';
 
-const retrievalEndpoint = 'https://scopelift.b-cdn.net/ipfs/<cid>/';
+const retrievalEndpoint = 'https://scopelift.b-cdn.net/ipfs';
 
 function assertIPFSPointer(logoPtr: MetaPtr | undefined) {
   if (!logoPtr) throw new Error('assertIPFSPointer: logoPtr is undefined');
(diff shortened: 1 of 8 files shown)


mds1 (approved):

app/src/utils/data/ipfs.ts
```diff
@@ -6,7 +6,7 @@ import { BigNumber } from 'src/utils/ethers';
 import { LocalForageData } from 'src/types';
 import { Ref } from 'vue';
 
-const retrievalEndpoint = 'https://ipfs.fleek.co/ipfs';
+const retrievalEndpoint = 'https://scopelift.b-cdn.net/ipfs/<cid>/';
```
apbendi, line 9: The `/<cid>/` part was meant to be a placeholder for the content's CID.

apbendi (changes requested):

app/src/utils/data/ipfs.ts
gdixon, line 9: I don't think we should be including `<cid>` in the endpoint:

```suggestion
const retrievalEndpoint = 'https://scopelift.b-cdn.net/ipfs';
```

(
[https://scopelift.b-cdn.net/ipfs/<cid\>/QmTQMgoxDRj8gfNn5Cvznt5CoEE6cz9MQeZrQsNb36BMdm ](https://scopelift.b-cdn.net/ipfs/<cid>/QmTQMgoxDRj8gfNn5Cvznt5CoEE6cz9MQeZrQsNb36BMdm 
)
vs 
https://scopelift.b-cdn.net/ipfs/QmTQMgoxDRj8gfNn5Cvznt5CoEE6cz9MQeZrQsNb36BMdm
)

@apbendi - I see in the issue you spoke about the trailing slash but it wasn't there previously - do we need it?

gdixon (changes requested):

app/src/utils/data/ipfs.ts
gdixon: Lol, I just commented on this - I'll delete it now! 🚀

mds1: Clearly I reviewed too quickly, whoops. Thanks all

mk1r:
ah I see makes sense, I was wondering about that lol but then forgot to check

good catches

app/src/utils/data/ipfs.ts
mk1r: so, can it be `https://scopelift.b-cdn.net/ipfs`

gdixon: I think that should work - the URL above resolves so I don't think we need that trailing slash

mk1r:
updated, thanks team

gdixon (approved):

thelostone-mc (approved):

Issue #509, opened 2021-11-16
Update app to always fetch on chain data using 3rd party node


apbendi:
Currently, if the user has not connected a wallet, or is not connected to the correct chain, we configure a provider with a default RPC URL and use that to fetch the data needed to load the app. Submitting transactions is not possible for a user in this state.

Later, when a user connects their wallet, we use the RPC connection from their MetaMask account to update the provider that is used both to fetch data AND to submit transactions to the chain. If the user is using an outdated or non-performant RPC endpoint, they'll see bad site performance trying to load the needed data from the chain.

We want to update the app to *always* use the default RPC URL we configure for the provider that *fetches* data from the chain. The user's connected provider will *only* be used when the user submits a transaction. This will allow us to use a 3rd party node operator, based on the results of #508, to provide the best possible performance for fetching chain data.

apbendi:
Closed by #520

PR #520, opened 2021-11-18
use default provider for fetch

thelostone-mc:
#### Description

`DEFAULT_PROVIDER` from chains.ts  is now used for fetch operations instead of relying on the provider configured in the user's wallet.
Any contract interaction would continue use the rpc url configured in the wallet.

Refs: https://github.com/dcgtc/dgrants/issues/509

### DIFF
--- a/app/src/components/TransactionStatus.vue
+++ b/app/src/components/TransactionStatus.vue
@@ -50,7 +50,7 @@
 <script lang="ts">
 import { computed, defineComponent, Ref, ref, SetupContext, watch } from 'vue';
 import { getEtherscanUrl } from 'src/utils/utils';
-import useWalletStore from 'src/store/wallet';
+import { DEFAULT_PROVIDER } from 'src/utils/chains';
 
 const emittedEventName = 'onReceipt'; // emitted once we receive the transaction receipt
 
@@ -78,12 +78,11 @@ function useTransactionStatus(hash: Ref, context: SetupContext<'onReceipt'[]>) {
   const etherscanUrl = computed(() => getEtherscanUrl(hash.value));
 
   // watch for new tx hashes then fetch receipt and wait for it to be mined, finally emit event with receipt status once mined
-  const { default_provider } = useWalletStore();
   // if the props.hash changes then we need to await the new tx
   watch(
     () => hash.value,
     async () => {
-      const receipt = await default_provider.value.waitForTransaction(hash.value);
+      const receipt = await DEFAULT_PROVIDER.waitForTransaction(hash.value);
       status.value = receipt.status === 1 ? 'success' : 'failed';
       emitTxReceipt(Boolean(receipt.status));
     },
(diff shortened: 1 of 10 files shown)


app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -352,12 +352,12 @@ function useGrantDetail() {
   const txHash = ref<string>();
 
   watch(
-    () => [grant.value, grantId.value, rounds.value, roundsCLRData.value, provider.value],
+    () => [grant.value, grantId.value, rounds.value, roundsCLRData.value, DEFAULT_PROVIDER],
```
gdixon, line 355: `DEFAULT_PROVIDER` should always hold a value, so we shouldn't need to check it here 👍

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -352,12 +352,12 @@ function useGrantDetail() {
...
     () => {
       // enter loading state between loads
       loading.value = true;
       // ensure the computed props are ready before fetching data
-      if (donationToken.value && rounds.value && roundsCLRData.value && provider.value) {
+      if (donationToken.value && rounds.value && roundsCLRData.value && DEFAULT_PROVIDER) {
```
gdixon, line 360: Or here

gdixon (approved):
Tiny nit, but all lgtm!! 🚀

app/src/components/TransactionStatus.vue
```diff
@@ -50,7 +50,7 @@
 <script lang="ts">
 import { computed, defineComponent, Ref, ref, SetupContext, watch } from 'vue';
 import { getEtherscanUrl } from 'src/utils/utils';
-import useWalletStore from 'src/store/wallet';
+import { DEFAULT_PROVIDER } from 'src/utils/chains';
```
mds1, line 53: I think I've found a few other instances we use the user's provider to read contract data, that we need to change to `DEFAULT_PROVIDER`. These link to lines on the last commit in this branch:

- L543 `cart.ts` :https://github.com/dcgtc/dgrants/blob/cc8f5485f8035b3a73e0eec34bea4ed351932734/app/src/store/cart.ts#L543
- L549 `cart.ts` :https://github.com/dcgtc/dgrants/blob/cc8f5485f8035b3a73e0eec34bea4ed351932734/app/src/store/cart.ts#L549
- L188 and L189 of `app/src/utils/utils`: https://github.com/dcgtc/dgrants/blob/cc8f5485f8035b3a73e0eec34bea4ed351932734/app/src/utils/utils.ts#L188-L189
- L464 of `grantRound.ts`: https://github.com/dcgtc/dgrants/blob/cc8f5485f8035b3a73e0eec34bea4ed351932734/app/src/utils/data/grantRounds.ts#L464
- L534 and 536 of `grantRound.ts`: https://github.com/dcgtc/dgrants/blob/cc8f5485f8035b3a73e0eec34bea4ed351932734/app/src/utils/data/grantRounds.ts#L534-L536

app/src/views/GrantRegistryGrantDetail.vue
mds1: Yep, and it's constant/not reactive/will never change, so we can remove it from the watcher entirely

mds1 (approved):

mds1:
Just making sure @thelostone-mc has nothing else left to do on this PR before merging, so will let him handle it

PR #522, opened 2021-11-18
fix logoURI split error 

### LINKED ISSUE #521: logoURI split error and 401 auth required error
The following errors are taking place when creating a grant

- [x] logoURI split error when no logo is uploaded
- [x] 401 authorization required error

Screen Shot 2021-11-17 at 3 35 16 PM

Screen Shot 2021-11-18 at 10 51 09 AM

mk1r:
resolves #521

### DIFF
--- a/app/src/utils/data/ipfs.ts
+++ b/app/src/utils/data/ipfs.ts
@@ -14,8 +14,8 @@ function assertIPFSPointer(logoPtr: MetaPtr | undefined) {
     logoPtr = decodeMetadataId(logoPtr);
   }
   const protocol = BigNumber.from(logoPtr.protocol).toString();
-  if (protocol !== '1' && protocol !== '0' && !logoPtr.pointer)
-    throw new Error(`assertIPFSPointer: Expected protocol ID of 1, found ${protocol}`);
+  if (protocol !== '0' && protocol !== '1')
+    throw new Error(`assertIPFSPointer: Expected protocol ID of 0 or 1, found ${protocol}`);
 }
 
 export const ipfs = createIpfs(import.meta.env.VITE_FLEEK_STORAGE_API_KEY);
--- a/app/src/utils/data/ipfs.ts
+++ b/app/src/utils/data/ipfs.ts
@@ -14,8 +14,7 @@ function assertIPFSPointer(logoPtr: MetaPtr | undefined) {
     logoPtr = decodeMetadataId(logoPtr);
   }
   const protocol = BigNumber.from(logoPtr.protocol).toString();
-  if (!['0', '1'].includes(protocol))
-    throw new Error(`assertIPFSPointer: Expected protocol ID of 0 or 1, found ${protocol}`);
+  if (protocol !== '1') throw new Error(`assertIPFSPointer: Expected protocol ID of 1, found ${protocol}`);
 }
 
 export const ipfs = createIpfs(import.meta.env.VITE_FLEEK_STORAGE_API_KEY);
--- a/app/src/utils/data/ipfs.ts
+++ b/app/src/utils/data/ipfs.ts
@@ -14,8 +14,8 @@ function assertIPFSPointer(logoPtr: MetaPtr | undefined) {
     logoPtr = decodeMetadataId(logoPtr);
   }
   const protocol = BigNumber.from(logoPtr.protocol).toString();
-  if (protocol !== '0' && protocol !== '1')
-    throw new Error(`assertIPFSPointer: Expected protocol ID of 0 or 1, found ${protocol}`);
+  if (!['0', '1'].includes(protocol))
+    throw new Error(`assertIPFSPointer: Expected protocol ID of 1, found ${protocol}`);
 }
 
 export const ipfs = createIpfs(import.meta.env.VITE_FLEEK_STORAGE_API_KEY);
@@ -55,7 +55,7 @@ export const getMetaPtr = ({ cid }: { cid: string | undefined }) => {
  */
 export const formatMetaPtr = (cid: string): MetaPtr => {
   return {
-    protocol: cid ? 1 : 0, // IPFS has a protocol ID of 1
+    protocol: 1, // IPFS has a protocol ID of 1
     pointer: cid,
   };
 };
--- a/app/src/utils/data/ipfs.ts
+++ b/app/src/utils/data/ipfs.ts
@@ -15,7 +15,7 @@ function assertIPFSPointer(logoPtr: MetaPtr | undefined) {
   }
   const protocol = BigNumber.from(logoPtr.protocol).toString();
   if (!['0', '1'].includes(protocol))
-    throw new Error(`assertIPFSPointer: Expected protocol ID of 1, found ${protocol}`);
+    throw new Error(`assertIPFSPointer: Expected protocol ID of 0 or 1, found ${protocol}`);
 }
 
 export const ipfs = createIpfs(import.meta.env.VITE_FLEEK_STORAGE_API_KEY);
--- a/app/src/utils/data/ipfs.ts
+++ b/app/src/utils/data/ipfs.ts
@@ -14,7 +14,8 @@ function assertIPFSPointer(logoPtr: MetaPtr | undefined) {
     logoPtr = decodeMetadataId(logoPtr);
   }
   const protocol = BigNumber.from(logoPtr.protocol).toString();
-  if (protocol !== '1') throw new Error(`assertIPFSPointer: Expected protocol ID of 1, found ${protocol}`);
+  if (protocol !== '1' && protocol !== '0' && !logoPtr.pointer)
+    throw new Error(`assertIPFSPointer: Expected protocol ID of 1, found ${protocol}`);
 }
 
 export const ipfs = createIpfs(import.meta.env.VITE_FLEEK_STORAGE_API_KEY);
@@ -54,7 +55,7 @@ export const getMetaPtr = ({ cid }: { cid: string | undefined }) => {
  */
 export const formatMetaPtr = (cid: string): MetaPtr => {
   return {
-    protocol: 1, // IPFS has a protocol ID of 1
+    protocol: cid ? 1 : 0, // IPFS has a protocol ID of 1
     pointer: cid,
   };
 };
--- a/app/src/views/GrantRegistryGrantDetail.vue
+++ b/app/src/views/GrantRegistryGrantDetail.vue
@@ -603,7 +603,8 @@ function useGrantDetail() {
     if (isMetaPtrUpdated) {
       const twitterURI = twitter === '' ? twitter : urlFromTwitterHandle(twitter);
       const properties = { websiteURI: website, githubURI: github, twitterURI };
-      const cid = (gMetadata?.logoPtr as MetaPtr).pointer;
+      let cid = '';
+      if (logoURI) cid = (gMetadata?.logoPtr as MetaPtr).pointer;
       metaPtr = await ipfs
         .uploadGrantMetadata({ name, description, logoPtr: ipfs.formatMetaPtr(cid), properties })
         .then((cid) => ipfs.formatMetaPtr(cid.toString()));
--- a/app/src/views/GrantRegistryGrantDetail.vue
+++ b/app/src/views/GrantRegistryGrantDetail.vue
@@ -281,7 +281,7 @@ import useCartStore from 'src/store/cart';
 import useDataStore from 'src/store/data';
 import useWalletStore from 'src/store/wallet';
 // --- Methods and Data ---
-import { LOREM_IPSOM_TEXT } from 'src/utils/constants';
+import { LOREM_IPSOM_TEXT, NO_LOGO_OBJECT } from 'src/utils/constants';
 import { ContractTransaction } from 'src/utils/ethers';
 import { DGRANTS_CHAIN_ID } from 'src/utils/chains';
 import {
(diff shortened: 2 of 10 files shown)


app/src/views/GrantRegistryNewGrant.vue
```diff
@@ -235,8 +235,11 @@ function useNewGrant() {
...
-    const splitLogoURI = (logoURI as string).split('/');
-    const cid = splitLogoURI[splitLogoURI.length - 1];
+    let cid = '';
+    if (logoURI) {
+      const splitLogoURI = (logoURI as string).split('/');
+      cid = splitLogoURI[splitLogoURI.length - 1];
+    }
     const metaPtr = await ipfs
       .uploadGrantMetadata({ name, description, logoPtr: ipfs.formatMetaPtr(cid), properties })
       .then((cid) => ipfs.formatMetaPtr(cid.toString()));
```
mds1, line 245: Two questions:
1. Don't we require an image to be uploaded on our new grant form? So how does `logoURI` end up undefined?
2. I haven't tested it, but the code seems like it'd result in metadata with an empty `logoPtr` field, but really it should have no `logoPtr` field. Curious if you've tested/confirmed this?

mk1r: 1) I have tested this out, the grant form does not require a logo image to be uploaded - is this a new behavior we want? 

2) In terms of empty vs no `logoPtr` field, I believe it was always either empty or has a logoURI value

thelostone-mc: Yup it's an optional field which is weird ! Cause ideally we'd want users to upload the image on grant creation.
Worth bringing it up on the call today ?

mds1: > I believe it was always either empty or filled with a logoURI value

I wanted to confirm this so we don't accidentally introduce a breaking changes, so:
1. First I checked out commit 14a84fd (an arbitrary commit before the `logoPtr` change), manually fixed the IPFS endpoint URL, then created a grant with no logo. The metadata for that can be found [here](https://bafybeiefsz5an7n6y6n7sppnr5qnkjdqjz55qywvjerqtlc2fqmb2hjj4a.ipfs.dweb.link/). You can this **does not have** a `logoURI` field
2. Then I checkout the latest commit on this branch and created a new grant. The metadata for that can be found [here](https://bafybeihib727xtcj5qzhaet4lbne34q4f6bcc3jxpo6d2a3xu2xrmqpice.ipfs.dweb.link/). You can see this **does have** a `logoPtr` field, and the `pointer` value is an empty string

So we don't want to merge this as is, because then we'll have portions of the app that think a valid `logoPtr` exists, and it has a valid `protocol` value of `1`, but the app will error when it tries to resolve that logo because the `pointer` is empty.

To result the current bug this PR is intended to fix we can either:
1. Require a logo be uploaded when creating a grant (new behavior), OR
2. Modify the grant metadata upload here so the full `logoPtr` object is not included in the metadata (the prior behavior)

I'm fine with either approach, we can decide at standup today.

**Second decision**: Interestingly, we can also see that previously we had inconsistent behavior of how the metadata is structured when empty fields are present: we had no `logoURI` field when there was no logo, but we did have present-but-empty `properties`, `websiteURI`, `githubURI`, `twitterURI` even when those fields are empty. We should pick an approach for those also and can fix it in this PR

mk1r: Sounds good, we should discuss this in standup - I'm fine with either solution as well

apbendi: Let's investigate this option:

Don't require a logo. If the user doesn't provide one, upload to

```
{
  "protocol": 0,
  "pointer":  ""
}
```

When loading the logos, use placeholder if protocol is not 1

thelostone-mc: That seems like a fair way to go about it

mk1r:
@apbendi @mds1 I think this is good solution

```
{
  protocol: 1, // IPFS has a protocol ID of 1
  pointer: cid ? cid : 'placeholder_grants.svg',
}
```

what this does is that it inserts the placeholder image if there no logo has been uploaded, and it also resolves #532

just one question is about protocol, I believe 1 is fine now since it won't have an empty cid field

mds1:
@mk1r That solution doesn't really make sense with our `MetaPtr` spec—the snippet you have would resolve to trying to load a hash of `placeholder_grants.svg` from IPFS (because `protocol` is 1, and of course that is not a valid IPFS hash. 

The face that we fallback to `placeholder_grants.svg` when there is no image should be independent of that metadata object. Is there a reason we can't go with Ben's suggested approach?

mk1r:
Hmm interesting you mention that we are loading a hash, because I tested it out and it was retrieving the `placeholder_grants.svg` image - could be that if it returns correct data, it falls back (but then not sure why it doesn't fallback if there is an empty cid as stated in #532)

mds1:
Huh interesting, that sounds like a bug then because it should be resolving the pointers using [`ptrToURI`](https://github.com/dcgtc/dgrants/blob/e12d7314ff279fe5b200e70f7d7662b9c6f80655/app/src/utils/utils.ts#L97-L102) which uses [`getMetaPtr`](https://github.com/dcgtc/dgrants/blob/e12d7314ff279fe5b200e70f7d7662b9c6f80655/app/src/utils/data/ipfs.ts#L46-L48).

Either way, even though it does work we don't want to conflate the protocol-level spec (the `MetaPtr` format) with the app behavior (fallback to `placeholder_grants.svg`). Those should be independent of each other and not coupled in the way they are [here](https://github.com/dcgtc/dgrants/pull/522#issuecomment-976944067)

Is there a reason we don't want to use [this approach](https://github.com/dcgtc/dgrants/pull/522#discussion_r753348723)?

mk1r:
@mds1 alright sounds good, and no that approach should work - just thought the approach I mentioned was cleaner and would solve #532 also

apbendi (changes requested):
@mk1r, I created a grant without a logo, but viewing the homepage with this branch, I still get the error "Unsupported Protocol ID 0". The app should not throw an error if there is an unsupported protocol, it should gracefully degrade and show the placeholder image instead.

So to recap:

* When creating a grant, if no image is provided, the logo metaptr should be set to 0/''
* When loading a grant, if the logo metaptr has a protocol value of anything other than `1`, the app should load the placeholder image.

Does this make sense?

mk1r:
@apbendi interesting, I'm not able to reproduce that error - for me, it's doing what you describe in the recap

I'll also double check the code in homepage (which I haven't modified) - which network are you using btw?

apbendi:
@mk1r I'm on Polygon.

> for me, it's doing what you describe in the recap

Maybe I'm missing something, but it looks like we're still throwing here, so I don't see how this would be the case: https://github.com/dcgtc/dgrants/pull/522/files#diff-e87dc6bd5d221441ff0d6a39efdce1b52d884442b20b73c3457e68aca79b29dcR18

mk1r:
@apbendi but it shouldn't throw when protocol is 0 right? https://github.com/dcgtc/dgrants/pull/522/files#diff-e87dc6bd5d221441ff0d6a39efdce1b52d884442b20b73c3457e68aca79b29dcR17

mk1r:
@apbendi think I found the issue, it's in utils - will push out an update, good find

mk1r:
try now @apbendi rebased as well

app/src/utils/data/ipfs.ts
```diff
@@ -54,7 +55,7 @@ export const getMetaPtr = ({ cid }: { cid: string | undefined }) => {
  */
 export const formatMetaPtr = (cid: string): MetaPtr => {
   return {
-    protocol: 1, // IPFS has a protocol ID of 1
+    protocol: cid ? 1 : 0, // IPFS has a protocol ID of 1
```
mds1, line 58: Why is this needed? We shouldn't have to special case protocol 0 within the `ipfs` (protocol 1) file

mk1r: isn't this where we want to make the adjustment to the object?

mds1: I think that would be better handled in the utils method that wraps around this method, similarly to help keep the app vs. protocol separation clean

mk1r: Can you point me to the file / line? This is what's being returned to be uploaded to ipfs, right? I assumed that's where this issue was https://github.com/dcgtc/dgrants/pull/522#discussion_r753219386

mds1: I’m on mobile right now about to take off on a flight, but I think I see what’s happening 

You’re using this in the upload logo step, which is coupled to IPFS, which is why you put it here. I thought this was for parsing the logo

It still feels a little dirty IMO to have it here instead of a createEmptyPtr() method that those files fallback to if no image is uploaded

mk1r: ah I see, I take a look at the method - but is there a different place we want to adjust the object? Isn't this where we want to make the change to the object if no logo is uploaded?

mds1: If we’re setting the logo pointer of metadata JSON to be empty and point to nothing, that is entirely independent of IPFS, which is why I don’t think it should be handled in this file. You can do that in utils or just generate it directly in the file(s?) that needs it

mk1r: I do think there is a way around modifying the object - is that what you're suggesting? i.e your option 2 https://github.com/dcgtc/dgrants/pull/522#discussion_r753219386

app/src/utils/data/ipfs.ts
```diff
@@ -14,7 +14,8 @@ function assertIPFSPointer(logoPtr: MetaPtr | undefined) {
     logoPtr = decodeMetadataId(logoPtr);
   }
   const protocol = BigNumber.from(logoPtr.protocol).toString();
-  if (protocol !== '1') throw new Error(`assertIPFSPointer: Expected protocol ID of 1, found ${protocol}`);
+  if (protocol !== '1' && protocol !== '0' && !logoPtr.pointer)
```
apbendi: Hmm, so as written, if anyone were to upload/update a grant with a metaPtr that had a value other than `1` or `0`, or with an empty pointer, this method would throw. Since we call this method in `getMetaPtr`, and use that method to fetch metadata throughout the app, then if anyone were to do this, the app would start throwing errors.

I don't believe an unsupported `protocol` value, or empty/invalid `metaPtr` value, should not result in the app throwing errors at any point, since users could put anything in there, and we can't control that. I think it should always degrade gracefully in some way, for example by simply excluding the grant if content fails to load. This might require a different PR though. Curious to hear what you think @mds1.

mds1: Generally agreed! Though this file isn't the right place to handle that graceful degradation based on how I was thinking of the architecture.

For context: `getMetaPtr` should primarily only be called by the `ptrToURI` wrapper method, with an exception for new grant / image upload code paths that call it directly because they're currently tightly coupled to IPFS. My intent with this `assertIPFSPointer` was to make sure only IPFS pointers are being passed into the methods in this file, and this file should be limited to handling cases where `protocol` is 1. Therefore if we throw in this method that's a bug and in that case I think it's ok to fail loudly. Given that, this line should only be `if (protocol !== '1') throw new Error(...)` and this entire file should not care about `protocol` being any non-1 value

Given all that, and if you agree with that architecture decision of isolating this file to strictly deal with IPFS, then `ptrToURI` and other methods in `utils.ts` are where we should gracefully handle the failure.

app/src/utils/data/ipfs.ts
```diff
@@ -14,7 +14,8 @@ function assertIPFSPointer(logoPtr: MetaPtr | undefined) {
     logoPtr = decodeMetadataId(logoPtr);
   }
   const protocol = BigNumber.from(logoPtr.protocol).toString();
-  if (protocol !== '1') throw new Error(`assertIPFSPointer: Expected protocol ID of 1, found ${protocol}`);
+  if (protocol !== '0' && protocol !== '1')
```
thelostone-mc, line 17:
```suggestion
  if (!['0', '1'].includes(protocol))
```

app/src/views/GrantRegistryGrantDetail.vue
```diff
@@ -598,7 +598,8 @@ function useGrantDetail() {
     if (isMetaPtrUpdated) {
       const twitterURI = twitter === '' ? twitter : urlFromTwitterHandle(twitter);
       const properties = { websiteURI: website, githubURI: github, twitterURI };
-      const cid = (gMetadata?.logoPtr as MetaPtr).pointer;
+      let cid = '';
+      if (logoURI) cid = (gMetadata?.logoPtr as MetaPtr).pointer;
```
thelostone-mc, line 602:
```suggestion
     const cid = logoURI ? (gMetadata?.logoPtr as MetaPtr).pointer : '';
```

app/src/utils/data/ipfs.ts
```diff
@@ -14,7 +14,8 @@ function assertIPFSPointer(logoPtr: MetaPtr | undefined) {
     logoPtr = decodeMetadataId(logoPtr);
   }
   const protocol = BigNumber.from(logoPtr.protocol).toString();
-  if (protocol !== '1') throw new Error(`assertIPFSPointer: Expected protocol ID of 1, found ${protocol}`);
+  if (!['0', '1'].includes(protocol))
+    throw new Error(`assertIPFSPointer: Expected protocol ID of 0 or 1, found ${protocol}`);
```
mds1: Pushed f1fd499 to revert this because this file shouldn't be aware of protocol ID 0

apbendi (approved):
Seems to be working 👍

PR #523, opened 2021-11-18
fix: optimises graph query pagination

gdixon:
As described in the documentation here (thank you @mds1 for looking through this with me!): https://thegraph.com/docs/developer/graphql-api#example-3

We can up the query limit to `1000` and we will also achieve better performance if we step through the results using `id_gt` (the graph stores data in ascending alphanumeric order based on the `ID` field rather than the creation time).

--

I've tested this out using the old subgraph and everything is pulled in correctly, it can also be tested in isolation here: https://thegraph.com/hosted-service/subgraph/dcgtc/dc-gitcoin-grants-matic?selected=playground

```
{
  grants(
    first: 1000, 
    where: {
      id_gt: "0x01",
      lastUpdatedBlockNumber_gte: "19841897"
    }
  ) {
    id
    owner
    payee
    metaPtr
    lastUpdatedBlockNumber
  }
}
```


app/src/utils/utils.ts
```diff
@@ -415,7 +415,7 @@ export const shuffle = (unshuffled: Grant[] | GrantRound[]) => {
 };
 
 /**
- * @notice Recursively grab every page of results
+ * @notice Recursively grab every 'page' of results - please note that the query MUST return the ID field
```
mds1, line 418: Given that `subgraphUrl`s in `chains.ts` are currently `false`, is there a way to test out this PR in the dgrants app? I don't think so, so perhaps we convert this to draft, and once the new subgraph is deployed we add the subgraph URLs in this same PR and test?

gdixon: Yes, that sounds good to me! As we discussed yesterday, I tested this against the old graph URLs but ran in to some unrelated errors so wasn't fully able to verify that everything is sound. I think setting this as a draft is a good call - I'll get that done now 👍

gdixon: The subgraphs are now deployed so I'm opening this back up for review 🙌

mds1 (approved):
Looks awesome, thanks @gdixon!

app/src/utils/utils.ts
```diff
@@ -426,29 +426,39 @@ export const shuffle = (unshuffled: Grant[] | GrantRound[]) => {
 export const recursiveGraphFetch = async (
```
mds1, line 426: One small comment is that it looks like the function's natspec comments are out of date

gdixon: Good catch - I've refactored so that we don't have to pass in the `lastId` and deleted the aforementioned comment 🙌

apbendi (approved):
Looks good, works great.

PR #524, opened 2021-11-19
Feature/my contributions

jjwoz:
Currently have the My contributions route working from the drop down menu (top right). Filtering data is complete but would potentially like too refactor to ensure the most performant way of aggregating the necessary data. 

Outstanding items:

- [x] transaction time
- enhancement strategy to switch between different data set needs (my contributions, contributions by round, contribution by grant). Ideally, we could allow this to ensure the correct filter function is called to set correct data into the details view.

### DIFF
--- a/app/src/components/ContributionDetail.vue
+++ b/app/src/components/ContributionDetail.vue
@@ -41,9 +41,13 @@
                     <Jazzicon :address="contribution.address" :key="contribution.address" :width="48" />
                   </figure>
                   <div>
-                    <a :href="getEtherscanUrl(contribution.address, 'address')" class="link">{{
-                      formatAddress(contribution.address)
-                    }}</a>
+                    <a
+                      :href="getEtherscanUrl(contribution.address, 'address')"
+                      class="link"
+                      target="_blank"
+                      rel="noopener noreferrer"
+                      >{{ formatAddress(contribution.address) }}</a
+                    >
                   </div>
                 </div>
               </div>
--- a/app/src/components/ContributionDetail.vue
+++ b/app/src/components/ContributionDetail.vue
@@ -1,11 +1,96 @@
 <template>
-  <div class="ContributionDetail">
-    <pre v-text="$attrs" />
-  </div>
+  <template v-if="contributions.length > 0">
+    <section v-for="contribution in contributions" :key="contribution.name" class="px-4 md:px-12">
+      <div class="py-4 md:py-6 lg:py-8 border-b border-grey-100">
+        <div class="grid grid-cols-12 gap-x-8 gap-y-4 items-center">
+          <!-- grand image + grant description + optional grant round -->
+          <article class="col-span-12 md:col-span-6 lg:col-span-5">
+            <!-- subgrid -->
+            <div class="grid grid-cols-4 items-center gap-x-4">
+              <!-- img -->
+              <div class="hidden lg:block col-span-1">
+                <!-- do not forget to link the figure to the grant ^^ -->
+                <figure class="aspect-w-16 aspect-h-9 shadow-light cursor-pointer">
+                  <img
+                    class="w-full h-full object-center object-cover group-hover:opacity-90"
+                    :src="'/placeholder_grant.svg'"
+                  />
+                </figure>
+              </div>
+              <!-- text -->
+              <div class="col-span-4 lg:col-span-3">
+                <a href="#" class="link">{{ contribution.name }}</a>
+                <!-- if this tx was via a matching round show round -->
+                <br /><span class="mr-2">via</span>
+                <a href="#" class="link">Round Dgrants Proof-Of-Concept</a>
+                <!-- end if -->
+              </div>
+            </div>
+          </article>
+
+          <!-- donors + amount -->
+          <article class="col-span-12 md:col-span-6 lg:col-span-4">
+            <!-- subgrid -->
+            <div class="flex justify-between items-center">
+              <!-- donors -->
+              <div>
+                <!-- subgrid -->
+                <div class="flex gap-4 items-center">
+                  <!-- do not forget to link the figure to the donors profile ^^ -->
+                  <figure class="cursor-pointer">
+                    <Jazzicon
+                      address="0x2aB75Ef4DC851CD34d4170B453cA56D2ef14683D"
+                      key="0x2aB75Ef4DC851CD34d4170B453cA56D2ef14683D"
+                      :width="48"
+                    />
+                  </figure>
+                  <div><a href="#" class="link">0x96…f1fe</a></div>
+                </div>
+              </div>
+              <!-- amount -->
+              <div>
+                <div class="text-grey-400">+ 12.1223 ETH</div>
+              </div>
+            </div>
+          </article>
+
+          <!-- transaction time + transaction hash-->
+          <article class="col-span-12 md:col-span-12 lg:col-span-3">
+            <!-- subgrid -->
+            <div class="flex justify-between lg:block lg:text-right">
+              <!-- utc-time -->
+              <div class="text-grey-400">2021-11-01 13:20:41</div>
+              <!-- tx-hash -->
+              <div class=""><a href="#" class="link">0x96…f1fe</a></div>
+            </div>
+          </article>
+        </div>
+      </div>
+    </section>
+  </template>
+  <template v-else>
+    <section class="px-4 md:px-12">
+      <p>No contributions found!</p>
+    </section>
+  </template>
 </template>
 
-<script>
-export default {
+<script lang="ts">
+import { Contribution } from '@dgrants/types';
+import { defineComponent, PropType } from 'vue';
+import Jazzicon from 'src/components/Jazzicon.vue';
+
+// const contributionIdList = ref([]);
+
+export default defineComponent({
   name: 'ContributionDetail',
-};
+  props: {
+    // -- Required props --
+    contributions: { type: Array as PropType<Contribution[]>, required: true },
+  },
+  components: { Jazzicon },
+  setup(props) {
+    console.log('props: ', props);
+  },
+});
 </script>
--- a/app/src/components/ContributionDetail.vue
+++ b/app/src/components/ContributionDetail.vue
@@ -1,0 +1,11 @@
+<template>
+  <div class="ContributionDetail">
+    <pre v-text="$attrs" />
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'ContributionDetail',
+};
+</script>
--- a/app/src/components/ContributionDetail.vue
+++ b/app/src/components/ContributionDetail.vue
@@ -1,6 +1,6 @@
 <template>
   <template v-if="contributions.length > 0">
-    <section v-for="contribution in contributions" :key="contribution.name" class="px-4 md:px-12">
+    <section v-for="contribution in contributions" :key="contribution.grantName" class="px-4 md:px-12">
       <div class="py-4 md:py-6 lg:py-8 border-b border-grey-100">
         <div class="grid grid-cols-12 gap-x-8 gap-y-4 items-center">
           <!-- grand image + grant description + optional grant round -->
@@ -13,16 +13,16 @@
                 <figure class="aspect-w-16 aspect-h-9 shadow-light cursor-pointer">
                   <img
                     class="w-full h-full object-center object-cover group-hover:opacity-90"
-                    :src="'/placeholder_grant.svg'"
+                    :src="contribution.grantLogoURI ? contribution.grantLogoURI : '/placeholder_grant.svg'"
                   />
                 </figure>
               </div>
               <!-- text -->
               <div class="col-span-4 lg:col-span-3">
-                <a href="#" class="link">{{ contribution.name }}</a>
+                <a href="#" class="link">{{ contribution.grantName }}</a>
                 <!-- if this tx was via a matching round show round -->
                 <br /><span class="mr-2">via</span>
-                <a href="#" class="link">Round Dgrants Proof-Of-Concept</a>
+                <a href="#" class="link">{{ contribution.roundName }}</a>
                 <!-- end if -->
               </div>
             </div>
@@ -38,18 +38,21 @@
                 <div class="flex gap-4 items-center">
                   <!-- do not forget to link the figure to the donors profile ^^ -->
                   <figure class="cursor-pointer">
-                    <Jazzicon
-                      address="0x2aB75Ef4DC851CD34d4170B453cA56D2ef14683D"
-                      key="0x2aB75Ef4DC851CD34d4170B453cA56D2ef14683D"
-                      :width="48"
-                    />
+                    <Jazzicon :address="contribution.address" :key="contribution.address" :width="48" />
                   </figure>
-                  <div><a href="#" class="link">0x96…f1fe</a></div>
+                  <div>
+                    <a :href="getEtherscanUrl(contribution.address, 'address')" class="link">{{
+                      formatAddress(contribution.address)
+                    }}</a>
+                  </div>
                 </div>
               </div>
               <!-- amount -->
               <div>
-                <div class="text-grey-400">+ 12.1223 ETH</div>
+                <div class="text-grey-400">
+                  {{ formatNumber(contribution.amount, 4) }}
+                  {{ contribution.donationToken?.symbol }}
+                </div>
               </div>
             </div>
           </article>
@@ -61,7 +64,9 @@
               <!-- utc-time -->
               <div class="text-grey-400">2021-11-01 13:20:41</div>
               <!-- tx-hash -->
-              <div class=""><a href="#" class="link">0x96…f1fe</a></div>
+              <div class="">
+                <a :href="getEtherscanUrl(contribution.txHash || '', 'tx')" class="link">{{ contribution.txHash }} </a>
+              </div>
             </div>
           </article>
         </div>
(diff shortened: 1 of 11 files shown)


apbendi (commented):
Hey @jjwoz, this seems to be looking and working well at first glance, nice work.

One TODO here, now that we've merged the "My Grants" PR, is to rebase, then move the drop down link into the same section of the menu:

Screen Shot 2021-11-24 at 1 11 21 PM

app/src/components/ContributionDetail.vue
```diff
@@ -0,0 +1,104 @@
+<template>
+  <template v-if="contributions.length > 0">
```
thelostone-mc, line 2: Thoughts on removing the template and just push the if condition to section ?

app/src/components/ContributionDetail.vue
```diff
@@ -0,0 +1,104 @@
...
+                      class="link"
+                      target="_blank"
+                      rel="noopener noreferrer"
+                      >{{ formatAddress(contribution.address) }}</a
+                    >
+                  </div>
+                </div>
+              </div>
+              <!-- amount -->
+              <div>
```
thelostone-mc, line 55: do we need the extra div ?

app/src/components/ContributionDetail.vue
```diff
@@ -0,0 +1,104 @@
...
+            <div class="flex justify-between items-center">
+              <!-- donors -->
+              <div>
+                <!-- subgrid -->
+                <div class="flex gap-4 items-center">
+                  <!-- do not forget to link the figure to the donors profile ^^ -->
+                  <figure class="cursor-pointer">
+                    <Jazzicon :address="contribution.address" :key="contribution.address" :width="48" />
+                  </figure>
+                  <div>
```
thelostone-mc, line 43: is this div needed ?

app/src/components/ContributionDetail.vue
```diff
@@ -0,0 +1,104 @@
...
+                    class="w-full h-full object-center object-cover group-hover:opacity-90"
+                    :src="contribution.grantLogoURI ? contribution.grantLogoURI : '/placeholder_grant.svg'"
+                  />
+                </figure>
+              </div>
+              <!-- text -->
+              <div class="col-span-4 lg:col-span-3">
+                <a href="#" class="link">{{ contribution.grantName }}</a>
+                <!-- if this tx was via a matching round show round -->
+                <br /><span class="mr-2">via</span>
```
thelostone-mc, line 24: do we wanna use <p> instead of using br ? or maybe using `block`

app/src/components/ContributionDetail.vue
```diff
@@ -0,0 +1,104 @@
...
+                <a href="#" class="link">{{ contribution.roundName }}</a>
+                <!-- end if -->
```
thelostone-mc, line 26: what is this for ?

app/src/components/ContributionDetail.vue
```diff
@@ -0,0 +1,104 @@
...
+              </div>
+            </div>
+          </article>
+
+          <!-- donors + amount -->
+          <article class="col-span-12 md:col-span-6 lg:col-span-4">
+            <!-- subgrid -->
+            <div class="flex justify-between items-center">
+              <!-- donors -->
+              <div>
```
thelostone-mc, line 36: do we need this extra div ?

app/src/components/ContributionDetail.vue
```diff
@@ -0,0 +1,104 @@
...
+
+          <!-- transaction time + transaction hash-->
+          <article class="col-span-12 md:col-span-12 lg:col-span-3">
+            <!-- subgrid -->
+            <div class="flex justify-between lg:block lg:text-right">
+              <!-- utc-time -->
+              <div class="text-grey-400">{{ contribution.createdAt }}</div>
+              <!-- tx-hash -->
+              <div class="truncate">
+                <a :href="getEtherscanUrl(contribution.txHash || '', 'tx')" class="link">{{ contribution.txHash }} </a>
```
thelostone-mc, line 72: should this open in a new tab / is this fine ?

app/src/components/ContributionDetail.vue
```diff
@@ -0,0 +1,104 @@
...
+              </div>
+            </div>
+          </article>
+        </div>
+      </div>
+    </section>
+  </template>
+  <template v-else>
+    <section class="px-4 md:px-12">
```
thelostone-mc, line 81: thoughts on dropping the template and moving v-else to section ?

app/src/views/Contribution.vue
```diff
@@ -0,0 +1,124 @@
...
+
+    <!-- btw in this component was a "cursor-pointer" for some reason. i removed it as this is
+    not something clickable. just for final PR this component could need that update too-->
+    <SectionHeader :title="`Contributions (${contributionTotal})`" />
+
+    <!-- a single contribution ( right now this would be the "ContributionRow.vue" what we maybe
+    not even wana keep - depends on the architecture you wana do ...
+    i added a v-for loop with some dummy data for {{name}} just to test how this looks
+    you can delete this of cause ...
+    -->
```
thelostone-mc, line 15: ^ do we need to clean up these comments ?

thelostone-mc (commented):
codewise lgtm. dropped a few nitpicks / questions

app/src/components/ContributionDetail.vue
jjwoz: to clarify, which template (ha!), So the only hesitation that I have with moving the conditional to a `section` tag is that I believe it will show up in the DOM, where as, the `template` tag does not appear in the DOM

app/src/components/ContributionDetail.vue
jjwoz: @thelostone-mc I am not sure I follow? Are you asking why this value is here?

app/src/views/Contribution.vue
jjwoz: @thelostone-mc we do not! Will remove

app/src/components/ContributionDetail.vue
mds1: > to clarify, which template

He is referring to the `template` elements that wrap each `section` element

> So the only hesitation that I have with moving the conditional to a `section` tag is that I believe it will show up in the DOM, where as, the `template` tag does not appear in the DOM

This is not entirely correct—if you implement @thelostone-mc's suggestion, each `section` tag would be part of `v-if` clause. When a `v-if` condition evaluates to false, that element does not show in the DOM. If you used a `v-show` instead, the element is in the dom, but hidden, when it evaluates to false

References with more info
- https://vuejs.org/v2/guide/conditional.html#v-if-vs-v-show
- https://stackoverflow.com/questions/51995815/vuejs-template-vs-div-or-related-for-grouping-elements-for-conditional-rend

app/src/components/ContributionDetail.vue
mds1: I don't think we need it, but quickest way to find out is just test removing it and see if it messes up formatting 🙂

app/src/components/ContributionDetail.vue
mds1: I think he is referring to the fact that the comments indicate there should be an `if` condition, but code has no such conditional. So either the comments are outdated or the code is missing something

app/src/components/ContributionDetail.vue
mds1: Agreed, should new tab. Just add `target=_blank` attribute to the `a` tag

app/src/utils/data/contributions.ts
```diff
@@ -368,3 +375,103 @@ export function grantDonationListener(
     off: () => grantRoundManager.value.off(filter, listener),
   };
 }
+
+/**
+ * @notice given a userAddress and the contributions, return only the contributions for that address
+ * @param userAddress
+ * @param contributions
+ */
+export function filterContributionsByUserAddress(userAddress: string, contributions: Contribution[]) {
```
mds1, line 384: This is exported but never used outside this file, so we can remove the export

app/src/utils/data/contributions.ts
```diff
@@ -368,3 +375,103 @@ export function grantDonationListener(
...
+  if (contributions && userAddress) {
```
mds1, line 385: `Boolean([])` resolves to true, so you need to explicitly check array length here if you want to return `undefined` for an empty array

```suggestion
  if (contributions.length > 0 && userAddress) {
```

app/src/utils/data/contributions.ts
```diff
@@ -368,3 +375,103 @@ export function grantDonationListener(
...
+    return contributions.filter((contribution: Contribution) => {
+      return contribution.address === userAddress;
+    });
+  }
+  return undefined;
```
mds1, line 390: Arguably a nit, but semantically we should be returning `null` instead, i.e. the value was set to `null`, whereas `undefined` means the variable hasn't been set yet

app/src/utils/data/contributions.ts
```diff
@@ -368,3 +375,103 @@ export function grantDonationListener(
...
+ * @param grantRoundMeta
+ */
+function filterGrantRoundsForContributions(
+  grantRounds: GrantRound[],
+  grantId: number,
+  grantRoundMeta?: Record<string, GrantRoundMetadata>
+) {
+  let roundName = '...';
+
+  if (grantRoundMeta) {
```
mds1, line 406: Since `Boolean({})` resolves to true, do we need a check that `grantRoundMeta` is not an empty object? Something like`Object.keys(grantRoundMeta).length > 0`

app/src/utils/data/contributions.ts
```diff
@@ -368,3 +375,103 @@ export function grantDonationListener(
...
+ */
+export function filterContributionGrantData(
+  userAddress: string,
+  contributions: Contribution[],
+  grants: Grant[],
+  grantRounds: GrantRound[],
+  grantMetaData?: Record<string, GrantMetadata>,
+  grantRoundsMetaData?: Record<string, GrantRoundMetadata>
+): ContributionDetail[] {
+  if (!userAddress || !contributions) {
```
mds1, line 436: Same comment on checking array length. There might be a few others like this that I missed so maybe just give that a quick review everywhere

app/src/utils/data/contributions.ts
```diff
@@ -368,3 +375,103 @@ export function grantDonationListener(
...
+      grantLogo = ptrToURI(grantMetaData[metaDataVersionId].logoPtr) ?? '';
+      grantName = grantMetaData[metaDataVersionId].name ?? '...';
+    }
+
+    const roundName = filterGrantRoundsForContributions(grantRounds, contribution.grantId, grantRoundsMetaData);
+
+    const date = contribution.createdAt
+      ? new Date(BigNumber.from(contribution.createdAt).toNumber() * 1000)
+      : undefined;
+    const contributionDate = date ? `${date.getDate()}-${date.getMonth() + 1}-${date.getFullYear()}` : undefined;
```
mds1, line 459: Same comments on `null` vs `undefined`

app/src/components/ContributionDetail.vue
thelostone-mc: i can throw in a follow up PR to fix that up if we need to get this across the finish line ! no biggie

app/src/components/ContributionDetail.vue
thelostone-mc: Haha yeah that's pretty which what I had asked :P

app/src/components/ContributionDetail.vue
jjwoz: fixed in push!

app/src/components/ContributionDetail.vue
scco: yes. its have organizational value for better readability

app/src/components/ContributionDetail.vue
scco: i think its fine as it is

app/src/components/ContributionDetail.vue
scco: the idea was to NOT show this, when there is no round data. so there must be a condition

app/src/components/ContributionDetail.vue
jjwoz: @thelostone-mc just to be clear - removed div and it appears to be working as expected

app/src/components/ContributionDetail.vue
scco: yes. i need all my code exactly as i did it.   there might be  1-2 unnessesary divs for sure what are super helpfull on another level of architecture and structure and later readability.

app/src/components/ContributionDetail.vue
jjwoz: @thelostone-mc  it does not appear that we do! removed

app/src/components/ContributionDetail.vue
jjwoz: @thelostone-mc @mds1 in favor of what @scco I am going to leave all html the way it (where comments ask if div is necessary). If that is alright with everyone?

app/src/components/ContributionDetail.vue
jjwoz: done!

app/src/utils/data/contributions.ts
jjwoz: so you are I are correct! Just need to combine  - `if (grantRoundMeta && Object.keys(grantRoundMeta).length > 0) ` because if grantRoundMeta does not exist then the Object.keys will freak out

app/src/components/ContributionDetail.vue
jjwoz: so the linter complains about doing this `3:5  error  This 'v-if' should be moved to the wrapper element  vue/no-use-v-if-with-v-for`  so I will keep it the way it is.

app/src/components/ContributionDetail.vue
mds1: Vue's `template` tag is generally what you should use if you need a div just for structure/readability that doesn't need to be rendered in the DOM

app/src/components/ContributionDetail.vue
jjwoz: @thelostone-mc - the template has to stay there because of the `v-for` otherwise the linter fails and vue complains. But I did try it

app/src/components/ContributionDetail.vue
mds1: Ah yes forgot about that. Ok great, agreed with leaving this as-is then

(More info if interested—in our case, leaving as-is would be best practice. In other cases, you'd want to instead use a computed property: https://eslint.vuejs.org/rules/no-use-v-if-with-v-for.html)

app/src/components/ContributionDetail.vue
```diff
@@ -0,0 +1,106 @@
...
+              <div class="hidden lg:block col-span-1">
+                <!-- do not forget to link the figure to the grant ^^ -->
+                <figure class="aspect-w-16 aspect-h-9 shadow-light cursor-pointer">
+                  <img
+                    class="w-full h-full object-center object-cover group-hover:opacity-90"
+                    :src="contribution.grantLogoURI ? contribution.grantLogoURI : '/placeholder_grant.svg'"
+                  />
+                </figure>
+              </div>
+              <div class="col-span-4 lg:col-span-3" v-if="contribution.grantName && contribution.roundName">
```
mds1, line 20: Nit, but readability it's preferred to put special Vue attributes first, and things like classes last: https://v3.vuejs.org/style-guide/#element-attribute-order-recommended

app/src/components/ContributionDetail.vue
```diff
@@ -0,0 +1,106 @@
...
+            <!-- subgrid -->
+            <div class="flex justify-between items-center">
+              <!-- donors -->
+              <div>
+                <!-- subgrid -->
+                <div class="flex gap-4 items-center">
+                  <!-- do not forget to link the figure to the donors profile ^^ -->
+                  <figure class="cursor-pointer">
+                    <Jazzicon :address="contribution.address" :key="contribution.address" :width="48" />
+                  </figure>
```
mds1, line 42: > `<!-- do not forget to link the figure to the donors profile ^^ -->`

Looks like we are not doing this

app/src/views/Contribution.vue
```diff
@@ -0,0 +1,112 @@
...
+
+function setTitle(route: string) {
+  return route === '/contribution/donations' ? 'My Contributions' : 'Contributions';
+}
+
+function contributionDetails() {
+  const { grantContributions, grants, grantMetadata, grantRounds, grantRoundMetadata } = useDataStore();
+  const { userAddress } = useWalletStore();
+
+  const fullContributions = ref();
```
mds1, line 30: This should have a type specified, otherwise it defaults to `any` which we don't want

app/src/views/Contribution.vue
```diff
@@ -0,0 +1,112 @@
...
+
+  const allGrants = computed(() => {
+    return grants.value ? grants.value : [];
+  });
+
+  const userAddr = computed(() => {
+    return userAddress.value ? userAddress.value : '';
+  });
+
+  watch(
```
mds1, line 58: Some comments on this component:
1. Seems all of these computed properties can be removed as they aren't doing anything significant except returning the same value, in some cases changing `undefined` to e.g. `[]`
2. Then, remove the watcher (there's no `async` calls in the watcher which typically means you don't need it) and can instead use computed properties which are lighter/more performant
3. `fullContributions` and `contributionTotal` should then become the two computed properties

app/src/components/ContributionDetail.vue
jjwoz: love the detail. Updated in push

app/src/components/ContributionDetail.vue
jjwoz: you would be correct. I am not sure this data is available?

app/src/views/Contribution.vue
jjwoz: @mds1 I went ahead and updated this, I agree with what you are saying, should be resolved with latest push

mk1r (approved):
functionality is working well, seems there are a lot of reviewers on the code already

app/src/components/ContributionDetail.vue
mds1: I think we just want to link to the address on Etherscan, since we don't have a profile feature

app/src/components/ContributionDetail.vue
mds1: Replaced divs that only enclose a single child (such as this one) with template tags in 370a9d3

app/src/components/ContributionDetail.vue
mds1: Replaced divs that only enclose a single child (such as this one) with template tags in 370a9d3

app/src/components/ContributionDetail.vue
mds1: Also moved the v-else to the front in 9d03355

app/src/utils/data/contributions.ts
mds1: Changed the `undefined` in the `date` ternary to `null` in 9d03355

app/src/views/Contribution.vue
mds1: Thanks so much for this refactor, appreciate it!

apbendi (approved):

PR #525, opened 2021-11-22
Fixing whitelist

### LINKED ISSUE #459: Change failure to load whitelist behavior
Failure to load the global whitelist should result in the following behavior
- Check for a local cache of the whitelist and use that if it exists
- If no local copy of the whitelist exists, display an error

Additionally we should add a fallback url for the whitelist which could be a different IPFS gateway or http server

Ajand:
Whenever we fetch the whitelist I will cache it and in case of not being able to fetch the whitelist, I will use the cached one.

- [x] Check for a local cache of the whitelist and use that if it exists
- [x] If no local copy of the whitelist exists, display an error

Also, put the fallback strategy into a new function so adding "Additionally we should add a fallback URL for the whitelist which could be a different IPFS gateway or HTTP server" must be pretty straightforward whenever the fallback URL is ready.

**Suggestion: Using the cache for whitelisting while trying to fetch whitelist, will give slightly better UX with faster load time**

fixed #459

### DIFF
--- a/app/src/store/data.ts
+++ b/app/src/store/data.ts
@@ -305,9 +305,13 @@ export default function useDataStore() {
     const whitelistUrl = import.meta.env.VITE_GRANT_WHITELIST_URI;
     if (whitelistUrl) {
       const url = whitelistUrl + uniqueStr;
-      const json = await fetch(url).then((res) => res.json());
-      if (!json) return grants;
-      grants = grants.filter((grant) => json[DGRANTS_CHAIN_ID].includes(grant.id));
+      try {
+        const json = await fetch(url).then((res) => res.json());
+        if (!json) return grants;
+        grants = grants.filter((grant) => json[DGRANTS_CHAIN_ID].includes(grant.id));
+      } catch (err) {
+        return grants;
+      }
     }
 
     return {
--- a/app/src/store/data.ts
+++ b/app/src/store/data.ts
@@ -8,7 +8,7 @@ import { computed, Ref, ref, watch } from 'vue';
 // --- Our imports ---
 import { BigNumber, Contract } from 'src/utils/ethers';
 import useWalletStore from 'src/store/wallet';
-import { getAllGrants, grantListener } from 'src/utils/data/grants';
+import { getAllGrants, grantListener, getApprovedGrants } from 'src/utils/data/grants';
 import { getContributions, getTrustBonusScores, grantDonationListener } from 'src/utils/data/contributions';
 import {
   getAllGrantRounds,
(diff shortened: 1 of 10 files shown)


thelostone-mc (approved):
LGTM! Btw @Ajand would it make sense to move this into utils ? (just curious cause data.ts seems to be growing in size)

Ajand:
> LGTM! Btw @Ajand would it make sense to move this into utils ? (just curious cause data.ts seems to be growing in size)

I think it makes so much sense to make it a pure function and put it into the utils. I also needed to use this util in #526

gdixon:
How are we handling cache invalidation here? What happens when new grants are added to the whitelist?

mk1r (approved):
+1 to moving this to utils, otherwise looking good!

mk1r:
@Ajand feel free to create a new issue to move to utils, or you can also update this PR

Ajand:
> How are we handling cache invalidation here? What happens when new grants are added to the whitelist?

I haven't handled that but what's your opinion on validating whitelist objects on the filterWhiteListed helper function?

app/src/store/data.ts
```diff
@@ -302,8 +302,32 @@ export default function useDataStore() {
     const whitelistUrl = import.meta.env.VITE_GRANT_WHITELIST_URI;
     if (whitelistUrl) {
       const url = whitelistUrl + uniqueStr;
-      const json = await fetch(url).then((res) => res.json());
-      grants = grants.filter((grant) => json[DGRANTS_CHAIN_ID].includes(grant.id));
+
+      const filterWhiteListed = (whiteList: Record<number, [number]>) =>
```
mds1, line 306: Two things here:
1. Nit: here we style variable as `whiteList` but above we use `whitelist`. I say we stick with `whitelist` everywhere for consistency
2. The type is defined as `Record<number, [number]>`, where `[number]` means a 1-element tuple where that element is a number. I think you want `number[]`, indicating an array of numbers

app/src/store/data.ts
```diff
@@ -302,8 +302,32 @@ export default function useDataStore() {
...
+        grants.filter((grant) => whiteList[DGRANTS_CHAIN_ID].includes(grant.id));
+
+      const fallback = async () => {
+        const cachedWhiteList: any = await getStorageKey('whitelist');
```
mds1, line 310: We should cast this to `Record<number, number[]>` it seems, instead of typing it as `any`. Might also want to define a new type of `type Whitelist = Record<number, number[]>` in `types/src/app.d.ts`

mds1:
> How are we handling cache invalidation here? What happens when new grants are added to the whitelist?

@gdixon It seems we always try to fetch from the whitelist URL and upon a successful fetch save the new data to the cache, so I don't think we need any additional cache invalidation behavior do we? Since it's always updated when possible + we only use as a fallback

gdixon:
Ahh, I misread, right you are!! 😅

Ajand:
- [x] define a new type of type Whitelist = Record<number, number[]> in types/src/app.d.ts
- [x] moved this into utils
- [x] fixed #526

mk1r:
rather than merging main into this branch, it's better to rebase against main (since we use the rebase workflow) - try pulling latest main and then switch to this branch and `git rebase main`

PR #529, opened 2021-11-22
error handling: decode metadataId

thelostone-mc:
#### 

Grants don't load due to face that metaPtr is not stored as MetaPtr but instead as a string on the metadata
This is causing the functions to error out.

The right way to do this would be to 
- fix create grant flow
- update grants metadata

The quick fix to unblock the team would be to get this PR in which does an error handling

Bug: 

Screenshot 2021-11-22 at 11 17 22 PM

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dgrants/app",
-  "version": "0.2.9-9061a2e.0",
+  "version": "0.2.10-e65fa59.0",
   "private": true,
   "scripts": {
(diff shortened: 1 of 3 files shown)


phutchins (approved):

gdixon (approved):
LGTM! @thelostone-mc have you located where the string version is coming from? 

Does this only happen when a user creates a new grant? When another user retrieves that grant, it has the correct structure right?

mk1r (approved):

PR #530, opened 2021-11-22
Show pending transaction when contributing to GrantRound

### LINKED ISSUE #426: Donating to a GrantRound should give feedback and show pending
Currently when donating directly to a GrantRound there is no feedback and it may appear that nothing is happening.

- Implement a similar UX flow to what we have in the cart donate process where feedback is immediately given upon actions taken (like clicking donate)

mk1r:
resolves #426 

side note: found more tech-debt, and will add it to the tech-debt issue

### DIFF
--- a/app/src/views/GrantRound.vue
+++ b/app/src/views/GrantRound.vue
@@ -351,7 +351,11 @@ function useGrantRoundDetail() {
     await addMatchingFunds(round, contributionAmount);
 
     // After successful transaction, clear the form data
-    form.value.amount = '';
+    if (txHash.value) {
+      form.value.amount = '';
+    } else {
+      roundStatus.value = '';
+    }
   }
 
   /**
--- a/app/src/views/GrantRound.vue
+++ b/app/src/views/GrantRound.vue
@@ -325,8 +325,6 @@ function useGrantRoundDetail() {
     if (!signer.value) throw new Error('Please connect a wallet');
     if (!isCorrectNetwork.value) throw new Error('Wrong network');
 
-    roundStatus.value = '1 of 1 pending';
-
     // Pull data from form
     const { amount } = form.value;
     const tokenAddress = grantRound.value.matchingToken.address;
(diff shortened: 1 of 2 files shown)


app/src/views/GrantRound.vue
```diff
@@ -44,7 +44,7 @@
         </InputRow>
 
         <!-- Submit and cancel buttons -->
-        <div class="flex justify-end pt-6">
+        <div v-if="roundStatus === ''" class="flex justify-end pt-6">
```
thelostone-mc, line 47: Thoughts on swapping the if and else conditions.

```
<div v-if="roundStatus">...</div>
<div v-else>...</div>
```

app/src/views/GrantRound.vue
```diff
@@ -329,19 +336,28 @@ function useGrantRoundDetail() {
     // contributionAmount must have the right number of decimals and be hexed
     const contributionAmount = parseUnits(amount, grantRound.value.matchingToken.decimals);
 
+    let txIndex = 1;
+    let txsNeeded = 1;
```
thelostone-mc, line 340: is this needed ? it seems like it's always `txsNeeded = txIndex + 1`

app/src/views/GrantRound.vue
mk1r: was matching what was done in `cart.ts` but I like your suggestion

app/src/views/GrantRound.vue
mk1r: if you need to approve (line `343`) txsNeeded will be different

apbendi:
hey @mk1r I'm getting the old BigNumber error, I think because this branch needs to be rebased. Can you do that real quick so we can evaluate this effectively?

mk1r:
@apbendi sure, rebased

apbendi (approved):
Looks good, works well

thelostone-mc (approved):

PR #531, opened 2021-11-22
feat: records timestamps on grants and contributions

gdixon:
This PR stores timestamps in to the models for `Grants` and `Contributions`

--

Tested locally

### DIFF
--- a/app/src/utils/data/grants.ts
+++ b/app/src/utils/data/grants.ts
@@ -60,7 +60,6 @@ export async function getAllGrants(forceRefresh = false) {
               }`,
               fromBlock
             );
-            console.log(grants);
             // update each of the grants
             grants.forEach((grant: GrantSubgraph) => {
               const grantId = BigNumber.from(grant.id).toNumber();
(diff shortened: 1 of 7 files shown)


app/src/utils/data/grants.ts
```diff
@@ -53,11 +53,14 @@ export async function getAllGrants(forceRefresh = false) {
...
                   payee
                   metaPtr
+                  createdAtTimestamp
+                  lastUpdatedTimestamp
                   lastUpdatedBlockNumber
                 }
               }`,
               fromBlock
             );
+            console.log(grants);
```
mk1r, line 63: do we need to log here? if so, I'd maybe make it more descriptive; or remove it otherwise

gdixon: Noo, that was accidentally left behind - I'll get it removed 😅 👍

mk1r (approved):
Looking good! 🤝

thelostone-mc (approved):

PR #549, opened 2021-12-01
fix: can now checkout even if grant is in an inactive round

### LINKED ISSUE #547: BUG: Checkout fails when grant round is not active
I experienced this in our production environment with a grant around that had not been started yet, trying to check out with one grant that was in the round.

Steps to reproduce:
1) Load dGrants with a grant round that has not been started
2) Add a grant from that round (or not in that round) to your cart
3) Click checkout
4) Check the dev console

mds1:
closes #547

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dgrants/app",
-  "version": "0.2.20-4e6ffa7.0",
+  "version": "0.2.21-f75d259.0",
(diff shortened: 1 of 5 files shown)


thelostone-mc:
tested it locally worksssss
image

thelostone-mc (approved):

phutchins (approved):

Issue #551, opened 2021-12-01
error : on creating grant without logo


thelostone-mc:
#### Description 

Looks like we have an error on creating a grant without a logo. 
Ideally we'd want to allow users to be able create a grant with a logo right ?

![image](https://user-images.githubusercontent.com/5358146/144268371-8399e8d1-accd-4ca1-b7bf-709b962d153a.png)

PR #554, opened 2021-12-01
fix: allow grant creation without logo

thelostone-mc:
### Description 

Allows you to create grant without logo

Closes https://github.com/dcgtc/dgrants/issues/551

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dgrants/app",
-  "version": "0.2.23-2999e9e.0",
+  "version": "0.2.24-097b29f.0",
   "private": true,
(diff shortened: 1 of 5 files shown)


mk1r (approved):

mds1:
This is an ok stopgap, but the fact that `ipfs.ts` needs to be aware of protocol ID 0 indicates we have coupling somewhere we shouldn't. @mk1r did you add this to the tech debt issue like we mentioned in discord?

mds1 (approved):

mk1r:
@mds1 updated it now, item 20 https://github.com/dcgtc/dgrants/issues/265

Issue #557, opened 2021-12-02
Improve how logo pointers are loaded


apbendi:
Currently, the [ptrToURI](https://github.com/dcgtc/dgrants/blob/61335b7076ba740cb0a5804380152d860a2314a3/app/src/utils/utils.ts#L103) utility method is coupled to the applications placeholder image, and to our use of lazy loading images.

This method should be thought of as a utility method that might eventually exist in a separate `dgrants` library. As such, it should know nothing about the implementation details of the frontend.

In particular:

1. The app should accept *only* a `MetaPtr`, not null or undefined. A component that uses this function is responsible for ensuring there is a `MetaPtr` object before calling this function, and should itself fallback to a placeholder rather than relying on this function to return one.
2. The app should not return the placeholder if the protocol is not supported. Instead, it should return `null`, and the component consuming this function's return value should render the placeholder if it does not return a URI.

PR #559, opened 2021-12-02
feat: adds component to contain `LogoPtr` image resolution and fallback process

gdixon:
This PR abstracts `ptrToURI()` calls to a component (`LogoPtrImage`) - this component should be responsible for the `logoPtr` handling and implementation details.

--

Fixes: #557

### DIFF
--- a/app/src/utils/data/contributions.ts
+++ b/app/src/utils/data/contributions.ts
@@ -445,7 +445,7 @@ export function filterContributionGrantData(
     if (grantData && grantMetaData && Object.keys(grantMetaData).length) {
       const metaDataVersionId = metadataId(grantData.metaPtr);
       const { logoPtr, name } = grantMetaData[metaDataVersionId];
-      grantLogo = ptrToURI(logoPtr, grantLogo);
+      grantLogo = (logoPtr && logoPtr.protocol === 1 ? ptrToURI(logoPtr) : false) || grantLogo;
       grantName = name || grantName;
     }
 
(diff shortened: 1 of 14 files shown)


jjwoz (approved):
lgtm - ran through the application and everything is working as expected from my point of view

app/src/utils/data/contributions.ts
```diff
@@ -445,7 +445,7 @@ export function filterContributionGrantData(
     if (grantData && grantMetaData && Object.keys(grantMetaData).length) {
       const metaDataVersionId = metadataId(grantData.metaPtr);
       const { logoPtr, name } = grantMetaData[metaDataVersionId];
-      grantLogo = ptrToURI(logoPtr, grantLogo);
+      grantLogo = (logoPtr && logoPtr.protocol === 1 ? ptrToURI(logoPtr) : false) || grantLogo;
```
mds1, line 448: This should also work right? `ptrToUri` will return null if the protocol is not 1
```suggestion
      grantLogo = ptrToURI(logoPtr) || grantLogo;
```

mds1 (approved):
Code LGTM, just one small comment

app/src/utils/data/contributions.ts
gdixon: I think you're right! Before I had this setup as `IpfsImg` and figured the component didn't need to be so specific, this is a remnant! I'll try it out and get it committed, thanks!

PR #560, opened 2021-12-02
add round contributions

### LINKED ISSUE #322: feat: show list of contributors for the matching pool
### Description

- show list of contributors who contribute to matching pool
- order by amount
- show amount

need to sync up with @scco  on design

mk1r:
closes #322


apbendi (changes requested):
hey @mk1r, just checked this out locally, and while it looks and works, it seems there was confusion about the actual requirements.

What it seems you built is "The contributions to grants in this grant round." What we were looking for was "The contributions to this Round's matching pool", i.e. those denoted by this event: https://github.com/dcgtc/dgrants/blob/main/contracts/contracts/GrantRound.sol#L95

mk1r:
@apbendi ah I see, okay let's discuss this in standup - do we still want this functionality?

PR #563, opened 2021-12-03
Disconnect wallet

### LINKED ISSUE #299: disconnect wallet doesn't truly disconnect
1. when your connected, hover over your top right address + identicon
2. click "disconnect wallet"
3. refresh page - you are still connected

https://user-images.githubusercontent.com/569641/135262752-99dca720-6805-4667-a3f5-7e7b16ba8efb.mov

mk1r:
Based on discussions here #299 it seems like the functionality we want is to keep the wallet disconnected when user clicks the disconnect wallet button, rather than auto reconnecting again on refresh. It will still auto reconnect on refresh when a user has their wallet connected.

closes #299

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dgrants/app",
-  "version": "0.2.25-e975958.0",
+  "version": "0.2.26-16695ff.0",
   "private": true,
   "scripts": {
     "clean": "rimraf dist",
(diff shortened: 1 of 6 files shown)


mds1 (approved):
LGTM, thanks!

apbendi (approved):
In testing this PR, I uncovered this bug, and created an issue for it: https://github.com/dcgtc/dgrants/issues/565

However this bug already exists in prod, and is not related to this change per-se. Therefore I think it's fine to merge this PR and address the new issue separately (but soon!).

PR #564, opened 2021-12-03
fix: 'require  is not defined' on prod for walletconnect

### LINKED ISSUE #555: Walletconnect throwing error
When connecting with walletconnect, it throws the following error

Screen Shot 2021-12-01 at 2 24 06 PM

mds1:
Resolves #555, see that issue for details

This is not yet tested since this issue only happens in production build. @phutchins do branches deploy to a staging env we can test this on before merging?

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dgrants/app",
-  "version": "0.2.26-16695ff.0",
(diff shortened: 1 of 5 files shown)


mk1r (approved):
not entirely sure how your change fixes it, but looking good!

PR #566, opened 2021-12-03
fix: ensure matching is calculated once metadata loads

gdixon:
This PR ensures that the `watch` method in data is triggered when the `GrantRounds` metadata is resolved.

--

Tested locally on Rinkeby


apbendi (approved):
Looks good and working well!

mds1 (approved):

PR #567, opened 2021-12-03
Matching pool contributions

### LINKED ISSUE #322: feat: show list of contributors for the matching pool

mk1r:
closes #322 
Screen Shot 2021-12-03 at 3 10 15 PM
(the view is updated)

### DIFF
--- a/app/src/utils/data/contributions.ts
+++ b/app/src/utils/data/contributions.ts
@@ -427,9 +427,7 @@ export function filterMatchingPoolContributions(
   const roundContributions = roundDonations[roundAddress];
   if (!roundContributions || roundContributions?.length === 0) return [];
   return roundContributions?.map((contribution: GrantRoundDonation) => {
-    const logo =
-      (roundLogoPtr && roundLogoPtr.protocol === 1 ? ptrToURI(roundLogoPtr) : false) || '/placeholder_grant.svg';
-
+    const logo = roundLogoPtr && ptrToURI(roundLogoPtr) ? ptrToURI(roundLogoPtr) : '/placeholder_grant.svg';
     const date = contribution.createdAt ? new Date(BigNumber.from(contribution.createdAt).toNumber() * 1000) : null;
     const contributionDate = date ? `${date.getDate()}-${date.getMonth() + 1}-${date.getFullYear()}` : null;
     const contributionAmount = Number(formatUnits(contribution.amount, grantRound.donationToken.decimals));
--- a/app/src/utils/data/contributions.ts
+++ b/app/src/utils/data/contributions.ts
@@ -250,7 +250,7 @@ export function filterContributionsByGrantId(grantId: number, contributions: Con
  * @notice given a grantRound and the contributions return only contributions for the grantRound
  */
 export function filterContributionsByGrantRound(round: GrantRound, contributions: Contribution[]) {
-  // filter contributions that are for this grantId
+  // filter contributions that are for this grantRound
   return contributions.filter((contribution: Contribution) => {
     // check that the contribution is valid
     const forThisGrantRound = contribution.inRounds?.includes(round.address);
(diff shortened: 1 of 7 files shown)


apbendi (commented):
Hey @mk1r, this doesn't seem to be showing Matching Pool contributions. It's just listing the contributions to grants in the round differently. Did something get mixed up reconciling the two PRs?

app/src/utils/data/contributions.ts
```diff
@@ -7,6 +7,8 @@ import {
   GrantMetadata,
   GrantRound,
   GrantRoundMetadata,
+  GrantRoundDonation,
```
mds1, line 10: (random comment location for threading)

Can we have the contributions sorted by timestamp, so the most recent ones are on top? Sort order seems to be based on transaction hash which isn't too intuitive

image

app/src/utils/data/contributions.ts
```diff
@@ -7,6 +7,8 @@ import {
...
+  MetaPtr,
```
mds1, line 11: (random comment location for threading)

Clicking the "Gitcoin Building Gitcoin" link takes me to the home page, but I'd expect it to take me to the round page. You can see in this screenshot that it shows the homepage URL on hover. It seems this doesn't work in other places that leverage the same component, so this slipped by previously it seems

image

app/src/utils/data/contributions.ts
```diff
@@ -414,6 +416,38 @@ function filterGrantRoundsForContributions(
...
+  roundName: string | undefined,
+  roundLogoPtr: MetaPtr | undefined
+): ContributionsDetail[] {
+  if (!grantRound || !roundDonations) return [];
+  const roundAddress = grantRound.address;
+  const roundContributions = roundDonations[roundAddress];
+  if (!roundContributions || roundContributions?.length === 0) return [];
+  return roundContributions?.map((contribution: GrantRoundDonation) => {
+    const logo =
+      (roundLogoPtr && roundLogoPtr.protocol === 1 ? ptrToURI(roundLogoPtr) : false) || '/placeholder_grant.svg';
```
mds1, line 431: `ptrToURI` returns null for unsupported protocols so the below should be easier to read and more scalable (i.e. automatically works when new protocol IDs are supported)

```suggestion
    const logo = roundLogoPtr && ptrToURI(roundLogoPtr) ? ptrToURI(roundLogoPtr) : '/placeholder_grant.svg';
```

mds1 (commented):
Looks good overall, a few comments

app/src/utils/data/contributions.ts
mk1r: this one is more difficult to do than it seems, since we are using the `ContributionDetail.vue` component for both matchingPool contributions as well as grant contributions - we need a different link for each, I guess we could make create a link field inside ContributionsDetail

app/src/utils/data/contributions.ts
mk1r: Just remembered that we want it sorted by amount (based on the issue). Should we sort by amount or timestamp?

mds1: oh sure amount is fine if that's what was spec'd out!

app/src/utils/data/contributions.ts
mds1: If you want to create a separate issue to track this and handle it in a separate PR, I think that's fine, cc @apbendi

mk1r: #597

jjwoz:
Looks good to me

jjwoz (approved):

mds1 (approved):

Issue #569, opened 2021-12-06
tech-debt :  grantContributions should always returns array


thelostone-mc:
### Description 

`grantContributions` ref in `data.ts` should default to empty array instead of undefined? 
If yes, this removes the need for `|| []` clause here

PR #570, opened 2021-12-06
chore: grantContributions should return []

thelostone-mc:
#### Description

Ensure grantContributions always returns array and never undefined

closes https://github.com/dcgtc/dgrants/issues/569

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dgrants/app",
-  "version": "0.2.30-dca5149.0",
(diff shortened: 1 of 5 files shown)


phutchins (approved):

mds1 (approved):
Looks simple enough, but before merging, what is this PR supposed to fix, i.e. why do we need this change?

thelostone-mc:
Picked that up from the tech-debt issue you'd created ! 
Plus it made sense as opposed to do `contributions || []` everywhere. Seemed cleaner
@mds1

PR #573, opened 2021-12-06
Feature/decouple metaptr logic

jjwoz:
Main goals were to separate logic out of each respected domain.


app/src/utils/utils.ts
```diff
@@ -17,9 +17,11 @@ import {
   SUPPORTED_TOKENS_MAPPING,
   WETH_ADDRESS,
 } from 'src/utils/chains';
-import { getMetaPtr } from 'src/utils/data/ipfs';
 import { Ref } from 'vue';
 
+// --- Constants ---
+const retrievalEndpoint = 'https://scopelift.b-cdn.net/ipfs';
```
gdixon, line 23: Could we move this to `constants.ts` and import from there?

app/src/utils/utils.ts
```diff
@@ -478,3 +480,32 @@ Promise<any[]> => {
     return await recursiveGraphFetch(url, key, query, fromBlock, [...before, ...json.data[key]]);
   }
 };
+
+export function assertIPFSPointer(logoPtr: MetaPtr | undefined) {
+  if (!logoPtr) throw new Error('assertIPFSPointer: logoPtr is undefined');
+  const protocol = BigNumber.from(logoPtr.protocol).toString();
+  if (!['0', '1'].includes(protocol))
+    throw new Error(`assertIPFSPointer: Expected protocol ID of 0 or 1, found ${protocol}`);
+}
```
mds1, line 489: `assertIPFSPointer` was intended an IPFS-specific method to ensure that _if_ you are calling a method within the `ipfs.ts` file _then_ the provided `MetaPtr` object must have a protocol ID of 1. Therefore IMO it doesn't make sense to move that out of `ipfs.ts` and into a generic `utils` file. Similarly, this method should not need to be aware of Protocol ID 0 (that was just temporary to get things working) 

Gotta run to a meeting now and will finish reviewing later, but this hints to me that the PR may need some refactoring to fully decouple things

mds1: (you might already know that, which might be why this is still a draft PR)

app/src/utils/utils.ts
jjwoz: We most certainly could

app/src/utils/utils.ts
jjwoz: You would be correct. I moved it to Utils because it doesn't really have anything to do with data specific needs like the rest of the functions in this file.

jjwoz: @mds1 I don't think we should hard code 1 or 0 - are we saying that for now we need to ensure it is equal to one? And if so we could possibly warrant a function that is solely used to check if the protocol number coming in, is acceptable?

mds1: sorry @jjwoz I'm not sure I follow your question here?

mds1: Not sure if this helps, but IMO the approach should be:
- anything specific to IPFS (e.g. asserting something has a protocol ID of 1, or caching something to the `ipfs-` storage key, or taking a pointer that's _known_ to have an ID of 1 and generating the URL) should live in `ipfs.ts`
- anything generic (e.g. taking a pointer with an unknown protocol ID and determining it's URL) should live in the new `metaPtrs.ts` file, e.g. it seems to make sense to have the `ptrToURI` method live here

jjwoz: so the reason I moved it because it is under the `data` folder - which I believe is surrounding async specific information. This is the basis for why I moved things around the way that I did.

jjwoz: @mds1 after standup today, I think the intent now for this PR will to be separate out by domain context (love it!) What I will probably start doing is:

- [ ] Remove the data folder
- [ ] organize by domain
- [ ] start conversation and analyze to remove ambiguous  `utils` files

PR #576, opened 2021-12-07
fix: moves to a post request when calling the trust-bonus endpoint

gdixon:
This PR will fix the `414 Request-URI Too Large` error we are seeing in production

--

Tested the call locally


mk1r (approved):
nice

apbendi (approved):
Working well and looks simple enough!

Issue #577, opened 2021-12-08
bug: Error when reloading the cart `Cannot read properties of undefined (reading 'grants')`


gdixon:
**Describe the bug**

If the user adds Grants which are both in a round and not in a round to their cart and then refreshes the page, they will see the following error:

`Cannot read properties of undefined (reading 'grants')`

**To Reproduce**

- Add all grants to your cart
- Go to the cart page
- Hard reload

**Expected behavior**

The cart should load without throwing any errors

**Screenshots**

<img width="1920" alt="Screenshot 2021-12-08 at 02 12 50" src="https://user-images.githubusercontent.com/5897836/145136689-6a60c3d3-9c03-4401-825d-9510d7e2cd98.png">

PR #578, opened 2021-12-08
fix: ensures metadata is present before checking the content of the grants property

gdixon:
This PR will fix an error on the cart where we were checking the contents of the metadata before it was available/resolved.

--

Fixes: #577

### DIFF
--- a/app/src/store/cart.ts
+++ b/app/src/store/cart.ts
@@ -393,7 +393,8 @@ export default function useCartStore() {
     return cart.value
       .map((item) =>
         (grantRounds.value || []).reduce((inRound: boolean, round) => {
-          return inRound || grantRoundMetadata.value[metadataId(round.metaPtr)].grants?.includes(item.grantId) || false;
+          const metadata = grantRoundMetadata.value[metadataId(round.metaPtr)];
+          return inRound || metadata?.grants?.includes(item.grantId) || false;
         }, false)
       )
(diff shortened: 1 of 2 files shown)


mk1r (approved):
nice fix! 👌

app/src/store/cart.ts
```diff
@@ -393,7 +393,8 @@ export default function useCartStore() {
     return cart.value
       .map((item) =>
         (grantRounds.value || []).reduce((inRound: boolean, round) => {
-          return inRound || grantRoundMetadata.value[metadataId(round.metaPtr)].grants?.includes(item.grantId) || false;
+          const metadata = grantRoundMetadata.value[metadataId(round.metaPtr)];
+          return inRound || metadata?.grants?.includes(item.grantId) || false;
```
mds1, line 397: Sanity check—the fix here is the added `?` before accessing the `grants` key right?

mds1 (approved):

app/src/store/cart.ts
mds1: And this means #577 was not actually related to a user adding "Grants which are both in a round and not in a round to their cart and then refreshes the page"? i.e. it's seems like this just has to do with refreshing on the cart page?

gdixon: Thats correct, I wasn't exactly sure what was going on when I wrote the issue up 😅

Issue #585, opened 2021-12-09
Donate to dGitcoin link goes to non dGitcoin grant


phutchins:
<img width="1379" alt="Screen Shot 2021-12-08 at 9 50 34 PM" src="https://user-images.githubusercontent.com/1074345/145325613-e1d9ac84-1e9b-4c1a-992a-a2a964f29f3d.png">
<img width="1472" alt="Screen Shot 2021-12-08 at 9 52 31 PM" src="https://user-images.githubusercontent.com/1074345/145325858-a83bb35c-54bb-42eb-be71-54abbaa3016a.png">

This should either be configurable or potentially able to be disabled.

PR #586, opened 2021-12-09
feat: allows for donate to dgrants btn to be disabled/configured

gdixon:
This PR adds the `VITE_DGRANTS_GRANT_ID` `.env` value to allow the `Donate to dGrants` button to be disabled/configured in the build.

--

Closes: #585

### DIFF
--- a/app/src/components/About.vue
+++ b/app/src/components/About.vue
@@ -34,7 +34,11 @@
 
         <!-- ToDo. Link to the real dGrant grant -->
         <div class="mt-16 flex flex-wrap gap-4">
-          <router-link to="/dgrants/0" @click="emitEvent('toggleAbout')">
+          <router-link
+            v-if="dgrantsGrantID && dgrantsGrantID !== 'false'"
+            :to="`/dgrants/${dgrantsGrantID}`"
+            @click="emitEvent('toggleAbout')"
+          >
             <button class="btn">
               <span>donate to <span class="text-teal">d</span>grants</span>
             </button>
@@ -79,11 +83,16 @@
 
 <script lang="ts">
 import { defineComponent, onMounted, ref } from 'vue';
+// --- Icons ---
 import { CloseIcon as XIcon } from '@fusion-icons/vue/interface';
 import { TwitterIcon as TwitterIcon } from '@fusion-icons/vue/interface';
 import { GithubIcon as GithubIcon } from '@fusion-icons/vue/interface';
 
+// package.json is used to inform the displayed version
 import packagejson from './../../package.json';
+
+// allow for the dgrants ID to be set in .env
+const dgrantsGrantID = import.meta.env.VITE_DGRANTS_GRANT_ID || false;
 
 function useContributors() {
   type Contributor = {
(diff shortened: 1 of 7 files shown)


app/src/components/About.vue
```diff
@@ -34,7 +34,11 @@
 
         <!-- ToDo. Link to the real dGrant grant -->
         <div class="mt-16 flex flex-wrap gap-4">
-          <router-link to="/dgrants/0" @click="emitEvent('toggleAbout')">
+          <router-link
+            v-if="dgrantsGrantID && dgrantsGrantID !== 'false'"
```
mds1, line 38: Might be cleaner to have `VITE_DGRANTS_GRANT_ID` be 0 or 1 and just check `v-if="dgrantsGrantID` and cast it to a boolean when reading the value

IME 0 or 1 is typically how boolean env vars are handled to avoid bugs where the string `false` is casted to true by JS. Plus `VITE_MAINTENANCE_MODE` is a boolean that uses 0 or 1, so this would keep things consistent

app/src/components/About.vue
```diff
@@ -79,12 +83,17 @@
...
+// --- Icons ---
 import { CloseIcon as XIcon } from '@fusion-icons/vue/interface';
 import { TwitterIcon as TwitterIcon } from '@fusion-icons/vue/interface';
 import { GithubIcon as GithubIcon } from '@fusion-icons/vue/interface';
 
+// package.json is used to inform the displayed version
 import packagejson from './../../package.json';
 
+// allow for the dgrants ID to be set in .env
+const dgrantsGrantID = import.meta.env.VITE_DGRANTS_GRANT_ID || false;
```
mds1, line 95: If above comment is implemented, this becomes
```suggestion
const dgrantsGrantID = Boolean(import.meta.env.VITE_DGRANTS_GRANT_ID);
```

app/src/components/About.vue
gdixon: The only issue with this is that part of the requirements is to be able to set which `grantID` actually relates to the dGrants grant. I think having the string version of `false` here should be fine, `0` is a valid `grantID` and we have no reason to cast this var to a number/boolean (in this use case at least).

mds1: Oh of course, saw the boolean check and apparently forget we needed an actual number here 🤦

So just to confirm then, if the env var is set to zero, it will get read in as a string so the `v-if="dgrantsGrantID && dgrantsGrantID !== 'false'"` check can still resolve to true? Because if it gets read as a number, `dgrantsGrantID` would resolve to false (because `Boolean(0)` is false, `Boolean('0')` is true)

app/src/components/About.vue
mds1: this comment can be ignored, see above

app/src/components/About.vue
gdixon: It will be read as a string, and that check will pass!

mds1: Thanks!

mds1 (approved):

apbendi (approved):
Looks and works well. Thanks 👍

Issue #587, opened 2021-12-09
BUG: Cache does not persist between page reloads


gdixon:
**Describe the bug**

The cache is not persisting between page reloads

**To Reproduce**

1. Visit https://grants.gtcdao.net/ and open the `Application` panel
2. Reload the page
3. Reload the stale data in the `Application` panel
4. Watch as the application data is cleared and then populated again

**Expected behavior**

The indexedDB cache should persist between page reloads

**Screenshots**

https://www.loom.com/share/9f177f61528646909d424a5069cf8109

PR #588, opened 2021-12-09
fix: ensures cache persists a page reload

gdixon:
On page-load the `network.value` was `undefined` which caused the `immediate` watcher to clear the storage. 

This PR corrects that behaviour and will ensure that the cache persists between page reloads. 

--

Fixes: #587 

-- 

Tested locally

### DIFF
--- a/app/src/store/data.ts
+++ b/app/src/store/data.ts
@@ -376,15 +376,17 @@ export default function useDataStore() {
 
     // record the network value to detect changes
     const networkValue = (await getStorageKey('network'))?.data || network.value;
+
     // watch the network
     watch(
       () => [network.value],
       async () => {
         // clear storage and poll again for all network changes after first load
-        if (networkValue && networkValue.chainId !== network.value?.chainId) {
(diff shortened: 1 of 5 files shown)


mds1 (approved):
Do you mind giving a short explanation about how this fix works? I'm not too familiar with the logic in this file so it's not immediately clear to me, and just curious to understand

app/src/store/data.ts
```diff
@@ -353,24 +353,24 @@ export default function useDataStore() {
...
     // record the network value to detect changes
     const networkValue = (await getStorageKey('network'))?.data || network.value;
+
     // watch the network
     watch(
       () => [network.value],
       async () => {
         // clear storage and poll again for all network changes after first load
-        if (networkValue && networkValue.chainId !== network.value?.chainId) {
+        if (networkValue && network.value && networkValue.chainId !== network.value?.chainId) {
```
apbendi, line 362:
```js
const hasChainIdChanged = networkValue && network.value && networkValue.chainId !== network.value?.chainId;
```

mk1r (approved):
nice

PR #590, opened 2021-12-10
fix: only allow donations if grant is whitelisted

### LINKED ISSUE #589: BUG: UI allows for donations to un-whitelisted grants
When a user attempts to add a grant that is not in the whitelist to their cart, it should fail. Also we could not display the add to cart button on non-whitelisted grants.

mk1r:
resolves #589

### DIFF
--- a/app/src/components/GrantCard.vue
+++ b/app/src/components/GrantCard.vue
@@ -24,7 +24,7 @@
 
       <!--cart button-->
       <div class="absolute bottom-0 right-0">
-        <template v-if="idList.includes(id)">
+        <template v-if="idList && idList.includes(id)">
           <button
             class="btn opacity-100 md:opacity-0 group-hover:opacity-100"
             :class="{ 'in-cart': isInCart(id) }"
@@ -96,8 +96,7 @@ export default defineComponent({
     const grantId = ref<number>(props.id);
     const raised = computed(() => getTotalRaised(grantId.value));
     const { addToCart, isInCart, removeFromCart } = useCartStore();
-    const { approvedGrantsPk } = useDataStore();
-    const idList = computed(() => approvedGrantsPk.value || []);
+    const { approvedGrantsPk: idList } = useDataStore();
 
     return {
       addToCart,
--- a/app/src/components/GrantCard.vue
+++ b/app/src/components/GrantCard.vue
@@ -24,13 +24,15 @@
 
       <!--cart button-->
       <div class="absolute bottom-0 right-0">
-        <button
-          class="btn opacity-100 md:opacity-0 group-hover:opacity-100"
-          :class="{ 'in-cart': isInCart(id) }"
-          @click.stop="isInCart(id) ? removeFromCart(id) : addToCart(id)"
-        >
-          <CartIcon />
-        </button>
+        <template v-if="idList.includes(id)">
+          <button
+            class="btn opacity-100 md:opacity-0 group-hover:opacity-100"
+            :class="{ 'in-cart': isInCart(id) }"
+            @click.stop="isInCart(id) ? removeFromCart(id) : addToCart(id)"
+          >
+            <CartIcon />
+          </button>
+        </template>
       </div>
     </div>
 
@@ -94,6 +96,8 @@ export default defineComponent({
     const grantId = ref<number>(props.id);
     const raised = computed(() => getTotalRaised(grantId.value));
     const { addToCart, isInCart, removeFromCart } = useCartStore();
+    const { approvedGrantsPk } = useDataStore();
+    const idList = computed(() => approvedGrantsPk.value || []);
 
     return {
       addToCart,
@@ -103,6 +107,7 @@ export default defineComponent({
       getEtherscanUrl,
       pushRoute,
       raised,
+      idList,
     };
   },
 });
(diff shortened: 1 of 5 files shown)


app/src/components/GrantDetailsRow.vue
```diff
@@ -86,6 +88,7 @@ export default defineComponent({
   components: { CartIcon, LogoPtrImage },
   props: {
     grant: { type: Object as PropType<Grant>, required: true },
+    idList: { type: Array as number, required: true },
```
mds1: Moving the fetching and storage of the whitelist to the `data` store feels like a much better architecture here.  As is, it seems we re-fetch the whitelist every time a different grant's detail page is loaded, which is wasteful and bad UX. We should fix that here, instead of coupling these components further

app/src/components/GrantDetailsRow.vue
```diff
@@ -86,6 +88,7 @@ export default defineComponent({
   components: { CartIcon, LogoPtrImage },
   props: {
     grant: { type: Object as PropType<Grant>, required: true },
+    idList: { type: Array as PropType<number[]>, required: true },
```
mds1, line 91: Moving the fetching and storage of the whitelist to the `data` store feels like a much better architecture here.  As is, it seems we re-fetch the whitelist every time a different grant's detail page is loaded, which is wasteful and bad UX. We should fix that here, instead of coupling these components further

mk1r: Yeah I think this is `tech-debt` which I thought got resolved in a different PR, right now this is a hotfix considering the issue is high priority. I'm now figuring out how to quickly make proper changes here.

mds1: Just add a `whitelistUrl` ref to the data store, return it as a computed property at the bottom of that file, and fetch the whitelist during `init` or `rawInit` (not sure where exactly is best offhand, cc @gdixon), then import it into the components

mk1r: sounds good - I was mulling between what you're suggesting but also if it makes sense to get it from `grants.ts` since  we are fetching it here already https://github.com/dcgtc/dgrants/blob/main/app/src/utils/data/grants.ts#L241

mk1r: seems like we have that in data store too, so I should be good - it probably never got changed in the `GrantRegistryGrantDetail.vue` file

mds1: Ah so then you just need to use the `approvedGrantsPk` that's returned from `data.ts`, which is where you're linked function is used. Even easier then 🙂

apbendi (approved):
Looking and working well to me 👍

app/src/components/GrantCard.vue
```diff
@@ -94,6 +96,8 @@ export default defineComponent({
     const grantId = ref<number>(props.id);
     const raised = computed(() => getTotalRaised(grantId.value));
     const { addToCart, isInCart, removeFromCart } = useCartStore();
+    const { approvedGrantsPk } = useDataStore();
+    const idList = computed(() => approvedGrantsPk.value || []);
```
mds1, line 100: `approvedGrantsPk` is already computed when returned from the store, so we don't need to re-wrap it in `computed` here. Let's just return `approvedGrantsPk` a few lines down

app/src/components/GrantDetailsRow.vue
```diff
@@ -93,7 +96,9 @@ export default defineComponent({
   },
   setup() {
     const { addToCart, isInCart, removeFromCart } = useCartStore();
-    return { addToCart, isInCart, removeFromCart, formatAddress, getEtherscanUrl };
+    const { approvedGrantsPk } = useDataStore();
+    const idList = computed(() => approvedGrantsPk.value || []);
```
mds1, line 100: Same comment about not needed to re-wrap this, since it's already computed

app/src/components/GrantCard.vue
```diff
@@ -24,13 +24,15 @@
...
       <!--cart button-->
       <div class="absolute bottom-0 right-0">
-        <button
-          class="btn opacity-100 md:opacity-0 group-hover:opacity-100"
-          :class="{ 'in-cart': isInCart(id) }"
-          @click.stop="isInCart(id) ? removeFromCart(id) : addToCart(id)"
-        >
-          <CartIcon />
-        </button>
+        <template v-if="idList.includes(id)">
```
mds1, line 27: Since `idList` can be undefined after removing the unneeded computed, just make sure to account for that. I think either of these should work, but worth testing
```suggestion
        <template v-if="idList?.includes(id)">
```

```suggestion
        <template v-if="idList && idList.includes(id)">
```

app/src/store/cart.ts
```diff
@@ -178,6 +178,8 @@ export default function useCartStore() {
    */
   function addToCart(grantId: number | undefined) {
     if (grantId === undefined) return;
+    const idList = approvedGrantsPk.value as Array<number>;
+    if (!idList.includes(grantId)) return;
```
mds1, line 182: If we're hiding the add to cart button for non-whitelisted grants, why do we need this? Seems like if executed this would result in silent failures that confuse the user (i.e. nothing happens when they try to add a grant to the cart)

app/src/components/GrantCard.vue
mk1r: I see, it was throwing linter errors so might still need the variable `idList` think it's bc approvedGrantsPk could be null

mk1r: just saw your comments below and will do that instead

app/src/store/cart.ts
mk1r: don't really need it but thought it could be used as an extra safety check idk

mds1 (approved):

PR #595, opened 2021-12-10
added card button in the nav when wallet is not connected

### LINKED ISSUE #593: Show The Cart Icon In Header Even When Wallet Is Not Connected
Should appear even when not connected, UI should match the connected case:

Screen Shot 2021-12-10 at 11 51 09 AM

Screen Shot 2021-12-10 at 11 51 15 AM

Ajand:
resolves #593

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dgrants/app",
-  "version": "0.2.37-1b39ccc.0",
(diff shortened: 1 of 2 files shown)


mk1r (approved):
Looks good and functions well

a nice to have would be to enable the Checkout button, and display a message saying 'Please connect wallet to checkout' (when not connected)

Issue #594, opened 2021-12-10
Cart Matching Estimates Are Off


apbendi:
A number of things seem to be broken when calculating cart estimates, possibly due to the same underlying cause:

* Matching estimates for USDT is calculated wrong (very high amounts)
* Matching estimates for fractions of MATIC/DAI/USDC is *higher* than for whole amounts (i.e. 10x more matching for 0.1 MATIC than 1 MATIC)
* Matching estimates for fractions of MATIC are *lower* than for equivalent amounts of DAI/USDC (despite the former being worth *more* than the latter), e.g. 0.1 MATIC shows lower match than 0.1 DAI

![dgrants-cart-matching-estimate-issues](https://user-images.githubusercontent.com/1479802/145612743-fb6a032a-efba-4d89-9da6-df943f867b52.gif)

PR #596, opened 2021-12-13
fix: clr prediction estimates showing incorrect value

thelostone-mc:
## Description

- fixes the issue with regard to USDT
- ensures prediction diff is always used 

fixes https://github.com/dcgtc/dgrants/issues/594

Untitled

Untitled

### DIFF
--- a/dcurve/src/internal/utils.ts
+++ b/dcurve/src/internal/utils.ts
@@ -75,8 +75,8 @@ export function getPredictedMatchingForAmount(clrPredictions: GrantPrediction, a
   // predictions should be provided with the same axis
   const contributionsAxis = [0, 1, 10, 100, 1000, 10000];
   // get all predictions for this grant in this round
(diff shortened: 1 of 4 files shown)


dcurve/src/internal/utils.ts
```diff
@@ -75,8 +75,8 @@ export function getPredictedMatchingForAmount(clrPredictions: GrantPrediction, a
   // predictions should be provided with the same axis
   const contributionsAxis = [0, 1, 10, 100, 1000, 10000];
   // get all predictions for this grant in this round
-  const clrPredictionCurve = clrPredictions?.predictions?.map((prediction, key) => {
-    return key == 0 ? prediction.predictedGrantMatch : prediction.predictionDiff;
```
thelostone-mc, line 79: @gdixon do you recollect why this change was made ?

mk1r (approved):
nice!

jjwoz (approved):
It looks good and functions as expected but should we use === 0?

PR #621, opened 2022-01-19
Adding handling for errors parsing grant round funds

### LINKED ISSUE #616: Site not working
Hey just wanted to inform that the site is broken its just loading for me and others too.. 
image
wondering what caused this.... the payout claim update?

phutchins:
Closes #616

### DIFF
--- a/app/src/utils/data/grantRounds.ts
+++ b/app/src/utils/data/grantRounds.ts
@@ -203,7 +203,15 @@ export async function getGrantRound(blockNumber: number, grantRoundAddress: stri
         matchingToken = SUPPORTED_TOKENS_MAPPING[matchingTokenAddress];
         donationToken = SUPPORTED_TOKENS_MAPPING[donationTokenAddress];
         // record the funds as a human readable number
-        funds = parseFloat(formatUnits(BigNumber.from(funds), SUPPORTED_TOKENS_MAPPING[matchingTokenAddress].decimals));
+        try {
+          funds = parseFloat(
+            formatUnits(BigNumber.from(funds), SUPPORTED_TOKENS_MAPPING[matchingTokenAddress].decimals)
+          );
+        } catch {
+          // If parsing funds for the round fails, ignore the round
+          console.log('Failed to parse funds for round');
+          return undefined;
+        }
       } else if (LocalForageData && _lsBlockNumber < blockNumber) {
         // get updated metadata
         const [newMetaPtr, balance] = await Promise.all([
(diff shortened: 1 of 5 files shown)


phutchins:
Replaced undefined return with empty object instead which appears to be passing linting.

BucksterBuckster (approved):

app/src/utils/data/grantRounds.ts
```diff
@@ -203,7 +203,15 @@ export async function getGrantRound(blockNumber: number, grantRoundAddress: stri
...
         // record the funds as a human readable number
-        funds = parseFloat(formatUnits(BigNumber.from(funds), SUPPORTED_TOKENS_MAPPING[matchingTokenAddress].decimals));
+        try {
+          funds = parseFloat(
+            formatUnits(BigNumber.from(funds), SUPPORTED_TOKENS_MAPPING[matchingTokenAddress].decimals)
+          );
+        } catch {
+          // If parsing funds for the round fails, ignore the round
+          console.log('Failed to parse funds for round');
+          return {};
```
mds1, line 213: We normally return a grant round which has fields that are expected to be defined based on the type definitions, which aren't necessarily reflected by the LocalForage abstractions—are you sure returning an empty `grantRound` object like this is safe and won't cause other issues?

metafraction (approved):
Looks good and fixes the hanging loading spinner. I was curious why you needed to return. I'm assuming it's to exit out of the function if error is caught. Returning an empty object is probably okay for that. 

nit: maybe use console.error instead of console.log?

app/src/utils/data/grantRounds.ts
metafraction: @mds1 what issues do you anticipate? I tested it out and seems to be fixing the error (hanging load) we're seeing, but there could be some unforeseen issues

metafraction: There could also be a more gracious way of fixing the error

phutchins: @mds1 there are a couple of places we use this method like where we update grant rounds. 

We should set a standard way of doing this (if we have not already) throughout our codebase. I.E. if you request an object from a method and it does not exist, fails to retrieve, or has errors, we should consistently return the same thing across the board. That way anywhere we are implementing that method, we can handle the results consistently.

phutchins:
Going to move forward with merging this and we can add an issue for ensuring that we handle errors processing grant rounds consistently.

